PR #8, opened 2015-10-29
Add contains to interface Node

zhengbli:
Fix https://github.com/Microsoft/TypeScript/issues/1253


mhegazy:
:+1:

sccolbert:
`contains()` should also be removed from the HTMLElement interface.

zhengbli:
@sccolbert That makes sense. I'll update soon

PR #9, opened 2015-10-29
Change HTMLSelectElement.options type to HTMLSelectElement | HTMLCollection

zhengbli:
Fix https://github.com/Microsoft/TypeScript/issues/1558


mhegazy:
:+1:

PR #19, opened 2015-11-06
Add HTMLTemplateDocument

dashaus:
Fix https://github.com/Microsoft/TypeScript/issues/5199


dashaus:
But I need to add the following, right?:

```typescript
declare var HTMLTemplateElement: {
  prototype: HTMLTemplateElement;
  new(): HTMLTemplateElement;
}
```

But I don't know how do it by addedTypes.json... :(

zhengbli:
@dashaus I'm adding support for those. Will keep you updated.

dashaus:
:+1:

zhengbli:
@dashaus The new bit is merged in. So to add constructors, you need to add a `constructorSignatures` string array property to your json object. Also, as this is a DOM type that should not be added to webworker definition, you may want to add a `flavor` property that has the value `Web`.

Your added json content should be something like:
```
    {
        "kind": "interface",
        "name": "HTMLTemplateElement",
        "flavor":  "Web",
        "extends": "HTMLElement",
        "constructorSignatures": [
            "new(): HTMLTemplateElement"
        ],
        "properties": [
            {
                "name": "content",
                "type": "DocumentFragment"
            }
        ]
    }
```

dashaus:
Something I did wrong ... I'm a newbie with git ... :(

I created a new pull request

PR #20, opened 2015-11-10
Add support for constructors in added interface json file

zhengbli:
For added interfaces in the json files, if there is a `constructorSignature` string array property, they will be incorporated in the emitted file.


sandersn:
Looks good, although it would be nice to have an actual usage in the interfaces json. I guess that's coming soon?

zhengbli:
Yes, it should be used by https://github.com/Microsoft/TSJS-lib-generator/pull/19

sandersn:
Cool :+1:

PR #22, opened 2015-11-11
Add HTMLTemplateDocument

dashaus:
Fix https://github.com/Microsoft/TypeScript/issues/5199

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -12610,6 +12610,15 @@ interface ProgressEventInit extends EventInit {
     total?: number;
 }
 
+interface HTMLTemplateElement extends HTMLElement {
+    content: DocumentFragment;
+}
+
+declare var HTMLTemplateElement: {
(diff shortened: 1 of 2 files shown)


mhegazy:
:+1:

mhegazy:
@zhengbli can we get this in?

zhengbli:
@mhegazy We can for now, but I found that many of the similar interfaces (HTMLxxxElement) doesn't actually have a legal constructor, while my script generates constructors for them because at the time I thought it would be consistent with the old version (before the script) and reduce breaking changes. Although I think we probably should remove them all together at some point.

mhegazy:
if we remove the constructors we break `instanceof` checks. so we need another solution here.

zhengbli:
In that case we can take this change now which solves this particular issue, and then figure out how to deal with the constructors systematically later.

zhengbli:
@dashaus Thansk for your contribution :)

PR #24, opened 2015-11-11
Overwrite return types for HTMLTableElement, HTMLTableSectionElement, and HTMLTableRowElement methods

zhengbli:
Fix https://github.com/Microsoft/TypeScript/issues/3583


mhegazy:
:+1:

PR #26, opened 2015-11-27
Fixes Microsoft/TypeScript#5813

xabufr:
Fixes https://github.com/Microsoft/TypeScript/issues/5813


DanielRosenwasser:
:+1: 

As a heads up @xabufr, GitHub doesn't link issues unless they're in the body of the PR, so I've added that for you.

mhegazy:
:+1:

PR #27, opened 2015-12-03
Added correct definition for createIndex method of IDBObjectStore

aksharpatel47:
Fixes Microsoft/Typescript#5832

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -5,6 +5,12 @@
          "name": "createIndex",
          "signatures": ["createIndex(name: string, keyPath: string | string[], optionalParameters?: any): IDBIndex"]
      },
+	 {
+		"kind": "property",
+		"interface": "IDBIndex",
+		"name": "keyPath",
+		"type": "string | string[]"
+	 },
     {
         "kind": "method",
         "interface": "WebGLRenderingContext",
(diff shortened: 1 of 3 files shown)


mhegazy:
:+1:

inputfiles/overridingTypes.json
```diff
@@ -1,4 +1,16 @@
  [
+     {
+         "kind": "method",
+         "interface": "IDBObjectStore",
+         "name": "createIndex",
+         "signatures": ["createIndex(name: string, keyPath: string | string[], optionalParameters?: any): IDBIndex"]
+     },
+	 {
```
zhengbli: niche: the indentation seems inconsistent. Use spaces instead of tabs.

zhengbli:
Other than the comment, looks good to me :+1:. Thanks for contributing @aksharpatel47 !

aksharpatel47:
@zhengbli There are a lot of other IndexedDB definitions missing such as multiEntry property of an Index, interfaces for options when creating an object store and an index. I can add them, but should I create a new issue and a pull request or add them here?

zhengbli:
@aksharpatel47 An issue should be created, the missing types can all belong to a single issue to avoid too much efforts. Some of them may be corrected next time we take an updated XML spec files from the Edge team, but it would be great if we can get a quick fix before that (which might be a painful process with many breaking changes). Thanks!

PR #29, opened 2015-12-09
Mark Element.setAttribute's parameters as required

LPGhatguy:
This fixes Microsoft/TypeScript#4407, which was reported as being an error on `HTMLElement`. The actual error was with `Element`.


zhengbli:
:+1: LGTM. Thanks @LPGhatguy !

PR #31, opened 2015-12-14
ReadyState type is a string

MattiasBuelens:
`MediaSource.readyState` is a [`ReadyState` string enum](https://w3c.github.io/media-source/#idl-def-ReadyState), but TypeScript [reports it](https://github.com/Microsoft/TSJS-lib-generator/blob/e9be0cb8e95e2a4b998cc0a658675127aac35e17/baselines/dom.generated.d.ts#L7641) to be a `number`.


zhengbli:
LGTM :+1: 
Thanks for contributing @MattiasBuelens !

PR #33, opened 2015-12-14
Override type of HTMLMediaElement.readyState

MattiasBuelens:
[The WebIDL file](https://github.com/Microsoft/TSJS-lib-generator/blob/472b4b1921d2b58ed8ae6505119d336d706ad972/inputfiles/browser.webidl.xml#L3747), reports `HTMLMediaElement.readyState` to be of type `any`. According to [the specification](https://html.spec.whatwg.org/multipage/embedded-content.html#htmlmediaelement) however, it should be of type `unsigned short` (which translates to `number` in TypeScript).

This PR adds an override for `HTMLMediaElement.readyState`.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5513,7 +5513,7 @@ interface HTMLMediaElement extends HTMLElement {
       * Gets or sets the current playback position, in seconds.
       */
     preload: string;
-    readyState: any;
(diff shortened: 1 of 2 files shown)


DanielRosenwasser:
:+1:

zhengbli:
:+1: LGTM

PR #38, opened 2016-01-06
Detect changes in json files regarding dictionary types

zhengbli:
Related to https://github.com/Microsoft/TypeScript/issues/5773

This fix enables the script to detect added/removed/overridden property for dictionary types in the spec.

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -538,6 +538,7 @@ let EmitDictionaries flavor =
             getRemovedItems ItemKind.Property flavor 
             |> Array.filter (matchInterface dict.Name)
             |> Array.map (fun rp -> rp.Name.Value)
+            |> Set.ofArray
         let addedProps = 
             getAddedItems ItemKind.Property flavor 
             |> Array.filter (matchInterface dict.Name)
@@ -545,7 +546,7 @@ let EmitDictionaries flavor =
         Pt.increaseIndent()
         Array.iter emitJsonProperty addedProps
         dict.Members
-        |> Array.filter (fun m -> not (Array.contains m.Name removedPropNames))
+        |> Array.filter (fun m -> not (Set.contains m.Name removedPropNames))
         |> Array.iter (fun m -> 
             match (findOverriddenItem m.Name ItemKind.Property dict.Name) with
             | Some om -> emitJsonProperty om
--- a/TS.fsx
+++ b/TS.fsx
@@ -531,8 +531,25 @@ let EmitDictionaries flavor =
         | "Object" -> Pt.printl "interface %s {" dict.Name
         | _ -> Pt.printl "interface %s extends %s {" dict.Name dict.Extends
 
+        let emitJsonProperty (p: ItemsType.Root) =
+            Pt.printl "%s: %s;" p.Name.Value p.Type.Value
+
+        let removedPropNames = 
+            getRemovedItems ItemKind.Property flavor 
+            |> Array.filter (matchInterface dict.Name)
+            |> Array.map (fun rp -> rp.Name.Value)
+        let addedProps = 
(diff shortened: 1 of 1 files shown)


TS.fsx
```diff
@@ -531,8 +531,25 @@ let EmitDictionaries flavor =
         | "Object" -> Pt.printl "interface %s {" dict.Name
         | _ -> Pt.printl "interface %s extends %s {" dict.Name dict.Extends
 
+        let emitJsonProperty (p: ItemsType.Root) =
+            Pt.printl "%s: %s;" p.Name.Value p.Type.Value
+
+        let removedPropNames = 
+            getRemovedItems ItemKind.Property flavor 
+            |> Array.filter (matchInterface dict.Name)
+            |> Array.map (fun rp -> rp.Name.Value)
```
sandersn: I would tack on 

```
to make the contains check below faster

File: TS.fsx [hunk 2]
```diff
@@ -531,8 +531,25 @@ let EmitDictionaries flavor =
...
+
         Pt.increaseIndent()
-        dict.Members |> Array.iter (fun m -> Pt.printl "%s?: %s;" m.Name (DomTypeToTsType m.Type))
+        Array.iter emitJsonProperty addedProps
+        dict.Members
+        |> Array.filter (fun m -> not (Array.contains m.Name removedPropNames))
+        |> Array.iter (fun m -> 
+            match (findOverriddenItem m.Name ItemKind.Property dict.Name) with
+            | Some om -> emitJsonProperty om
+            | None -> Pt.printl "%s?: %s;" m.Name (DomTypeToTsType m.Type))
```
sandersn: there's always multiple way to do things in F#. My personal style preference would be to do
```fs
let isRemovedPropName =
    let removedPropNames = ...
    not << fun m -> Set.contains m.Name removedPropNames
(* ... find and emit addedProps ... *)
for m in Array.filter isRemovedPropName dict.Members do
  (* ... same match body as before ... *)
```

This makes the side-effecting syntactically obvious and reduces the overhead from F#'s lambda syntax.

TS.fsx
zhengbli: Which contains check?

sandersn:
```fs
(fun m -> not (Array.contains m.Name removedPropNames)
```

I would change `Array.contains` to `Set.contains`.

TS.fsx
zhengbli: It is more or less subjective when it comes to the syntax preferences though. I used the continuation style because it makes the data flow clearer, and avoids the burden of naming things. Though I agree both styles have merits.

TS.fsx
zhengbli: How is `Set.contains` faster than `Array.contains`? My impression is the opposite.

sandersn: `Set.contains` doesn't do any implicit conversions. You'd pay the construction cost once when building removedPropNames by adding `|> Set.ofArray`. 

*Generally*, Set.contains is O(lg n) or O(1) and Array.contains is O(n), so if removedPropNames is ever long-ish (10-100 or more items probably?) it will be worth it.

However, I can't find any documentation on `Array.contains`! Where does `Array.contains` come from? I see that Shared.fs adds `Seq.contains x = Seq.exists ((=)x)`, which is O(n), but I can't find a definition of `Array.contains` in our own code.

zhengbli: Good point. For some reason I was thinking about `Seq` instead of `Set`, therefore had the impression that `Array.contains` is faster. Will update soon.

I believe `Array.contains` is added in F# 4.0 which was not very well documented.

zhengbli: Done.

sandersn:
:1+:

PR #39, opened 2016-01-09
Fix `XMLHttpRequest` interface

falsandtru:
issue: https://github.com/Microsoft/TypeScript/issues/5767
spec: https://xhr.spec.whatwg.org/#interface-xmlhttprequest


falsandtru:
@zhengbli excuse me, it seems impossible that I pass the tests with changes of json files in this case. How do I change the json files?

zhengbli:
@falsandtru I commented on the issue in TypeScript repo.

falsandtru:
thanks!

PR #40, opened 2016-01-09
Add a 'README.md' to the 'inputfiles' directory.

DanielRosenwasser:
This will help avoid PRs where people spend time and we end up telling them that we can't accept the changes.


zhengbli:
:+1:

DanielRosenwasser:
We should probably do both.

PR #48, opened 2016-01-15
Update `StorageEvent` interface

falsandtru:
fixes https://github.com/Microsoft/TypeScript/issues/6295
spec https://www.w3.org/TR/webstorage/


zhengbli:
:+1: Thanks!

PR #49, opened 2016-01-15
Update IndexedDB interfaces

falsandtru:
fixes https://github.com/Microsoft/TypeScript/issues/6501
spec https://www.w3.org/TR/IndexedDB/


falsandtru:
I recreate this pr for latest branch merging.

PR #44, opened 2016-01-12
To fix #2583

plantain-00:
https://github.com/Microsoft/TypeScript/issues/2583

It seems the constructor of interface URL is not generated, is this a bug?

PR #52, opened 2016-01-19
Detect added json items for static interfaces

zhengbli:
The issue:
For interfaces marked as `static` in the spec, we used to not detect changes made in the external json files.

Related to PR #44

### DIFF
--- a/Shared.fsx
+++ b/Shared.fsx
@@ -88,9 +88,19 @@ module JsonItems =
             t.Kind.ToLower() = kind.ToString() && 
             (t.Flavor.IsNone || t.Flavor.Value = flavor.ToString() || flavor = All))
 
-    let getOverriddenItems kind flavor = getItems overriddenItems kind flavor
-    let getAddedItems kind flavor = getItems addedItems kind flavor
-    let getRemovedItems kind flavor = getItems removedItems kind flavor
+    let getOverriddenItems kind flavor =
+        getItems overriddenItems kind flavor
+    let getAddedItems kind flavor =
+        getItems addedItems kind flavor
+    let getRemovedItems kind flavor =
+        getItems removedItems kind flavor
+
+    let getAddedItemsByInterfaceName kind flavor iName =
+        getItems addedItems kind flavor |> Array.filter (matchInterface iName)
+    let getOverriddenItemsByInterfaceName kind flavor iName =
+        getItems overriddenItems kind flavor |> Array.filter (matchInterface iName)
+    let getRemovedItemsByInterfaceName kind flavor iName =
+        getItems removedItems kind flavor |> Array.filter (matchInterface iName)
 
 module Comments =
     type CommentType = JsonProvider<"inputfiles/comments.json">
--- a/Shared.fsx
+++ b/Shared.fsx
@@ -11,8 +11,8 @@ open Microsoft.FSharp.Reflection
 /// ===========================================
 /// Global variables
 /// ===========================================
-module GlobalVars = 
-    if not (Directory.Exists(__SOURCE_DIRECTORY__ + @"/generated")) then 
+module GlobalVars =
+    if not (Directory.Exists(__SOURCE_DIRECTORY__ + @"/generated")) then
         Directory.CreateDirectory(__SOURCE_DIRECTORY__ + @"/generated") |> ignore
 
     let inputFolder = __SOURCE_DIRECTORY__ + @"/inputfiles"
@@ -29,7 +29,7 @@ module GlobalVars =
 /// ===========================================
 
 /// Quick checker for option type values
-let OptionCheckValue value = function 
+let OptionCheckValue value = function
     | Some v when v = value -> true
     | _ -> false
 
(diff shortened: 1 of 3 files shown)


Shared.fsx
```diff
@@ -88,9 +88,19 @@ module JsonItems =
...
 
-    let getOverriddenItems kind flavor = getItems overriddenItems kind flavor
-    let getAddedItems kind flavor = getItems addedItems kind flavor
-    let getRemovedItems kind flavor = getItems removedItems kind flavor
+    let getOverriddenItems kind flavor =
+        getItems overriddenItems kind flavor
+    let getAddedItems kind flavor =
+        getItems addedItems kind flavor
+    let getRemovedItems kind flavor =
+        getItems removedItems kind flavor
```
sandersn: why not let these be the point-free versions, eg

```fs
let getAddedItems = getItems addedItems
```

Shared.fsx
```diff
@@ -88,9 +88,19 @@ module JsonItems =
...
+
+    let getAddedItemsByInterfaceName kind flavor iName =
+        getItems addedItems kind flavor |> Array.filter (matchInterface iName)
+    let getOverriddenItemsByInterfaceName kind flavor iName =
+        getItems overriddenItems kind flavor |> Array.filter (matchInterface iName)
+    let getRemovedItemsByInterfaceName kind flavor iName =
+        getItems removedItems kind flavor |> Array.filter (matchInterface iName)
```
sandersn: why not use the pre-composed versions you just added? eg instead of `getItems removedItems` say `getRemovedItems`

TS.fsx
```diff
@@ -465,17 +465,33 @@ let EmitStaticInterface flavor (i:Browser.Interface) =
...
+            else
+                JsonItems.getAddedItemsByInterfaceName ItemKind.Property flavor i.Name |> Array.isEmpty |> not
+        hasNonStaticMethod || hasProperty
+
+    let emitAddedConstructor () =
+        match JsonItems.getAddedItemsByInterfaceName ItemKind.Constructor flavor i.Name with
+        | [||] -> ()
+        | ctors ->
+            Pt.printl "prototype: %s;" i.Name
+            ctors |> Array.iter (fun ctor -> ctor.Signatures |> Array.iter (Pt.printl "%s;"))
```
sandersn: You can save a couple lines and some visual noise by getting rid of the match:

```fs
let ctors = JsonItems....
if not <| Array.isEmpty ctors then Pt.printl ...
ctors |> Array.iter ...
```

TS.fsx
```diff
@@ -506,7 +523,7 @@ let EmitStaticInterface flavor (i:Browser.Interface) =
         EmitConstants i
         EmitEventHandlers prefix i
         EmitIndexers EmitScope.StaticOnly i
-
+        emitAddedConstructor ()
```
sandersn: when does a pure static interface have a constructor?

TS.fsx
```diff
@@ -465,17 +465,33 @@ let EmitStaticInterface flavor (i:Browser.Interface) =
...
+        let hasNonStaticMethod =
+            match JsonItems.getAddedItemsByInterfaceName ItemKind.Method flavor i.Name with
+            | [||] ->
+                match i.Methods with
+                | Some ms -> ms.Methods |> Array.exists (fun m -> m.Static.IsNone)
+                | _ -> false
+            | addedMs -> addedMs |> Array.exists (fun m -> m.Static.IsNone || m.Static.Value = false)
+        let hasProperty =
+            if i.Properties.IsSome then true
+            else
```
sandersn: `if x then true else y` is equivalent to `x || y`

TS.fsx
```diff
@@ -465,17 +465,33 @@ let EmitStaticInterface flavor (i:Browser.Interface) =
...
-    let hasNonStaticMember = 
-        match i.Methods with
-        | Some ms -> ms.Methods |> Array.exists (fun m -> m.Static.IsNone)
-        | _ -> false
+    let hasNonStaticMember =
+        let hasNonStaticMethod =
+            match JsonItems.getAddedItemsByInterfaceName ItemKind.Method flavor i.Name with
+            | [||] ->
+                match i.Methods with
+                | Some ms -> ms.Methods |> Array.exists (fun m -> m.Static.IsNone)
```
sandersn: why don't you have to also check `m.Static.Value` like in the added methods case?

TS.fsx
```diff
@@ -465,17 +465,33 @@ let EmitStaticInterface flavor (i:Browser.Interface) =
...
+                | _ -> false
+            | addedMs -> addedMs |> Array.exists (fun m -> m.Static.IsNone || m.Static.Value = false)
```
sandersn: In the added methods case, don't you also need to check the non-overridden methods in `i.Methods`?

Shared.fsx
zhengbli: Good point!

TS.fsx
zhengbli: This happens when the spec wrongly marked an interface as pure static, and a constructor is added in the json files. For example Microsoft/TypeScript#2583

TS.fsx
zhengbli: The original idea was the there is no static property, the only thing similar may be a static constant, which should be added as `constant` in the json file because it has a value. Though you may be right that it is better checking here to be conservative.

sandersn:
Besides a minor comment, :+1:

PR #53, opened 2016-01-19
Add HTMLPictureElement Interface

YuichiNukiyama:
fix [HTMLPictureElement interface definition is missing
#5323](https://github.com/Microsoft/TypeScript/issues/5323) issue. And add
 methods of Document & Element interface which related HTMLPictureElement.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2569,6 +2569,8 @@ interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEven
       * @param content The text and HTML tags to write.
       */
     writeln(...content: string[]): void;
+    createElement(tagName: "picture"): HTMLPictureElement;
+    getElementsByTagName(tagname: "picture"): NodeListOf<HTMLPictureElement>;
     addEventListener(type: "MSContentZoom", listener: (ev: UIEvent) => any, useCapture?: boolean): void;
     addEventListener(type: "MSGestureChange", listener: (ev: MSGestureEvent) => any, useCapture?: boolean): void;
     addEventListener(type: "MSGestureDoubleTap", listener: (ev: MSGestureEvent) => any, useCapture?: boolean): void;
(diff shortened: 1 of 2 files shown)


baselines/dom.generated.d.ts
```diff
@@ -2569,6 +2569,8 @@ interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEven
       * @param content The text and HTML tags to write.
       */
     writeln(...content: string[]): void;
+    createElement(tagName: "picture"): HTMLPictureElement;
```
mhegazy: these will be ignored if they are defined here after the less specific signature: `createElement(tagName: string)`. 
@zhengbli why is the order messed up here? and how can we fix it?

baselines/dom.generated.d.ts
```diff
@@ -2569,6 +2569,8 @@ interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEven
...
+    getElementsByTagName(tagname: "picture"): NodeListOf<HTMLPictureElement>;
```
mhegazy: same as above.

baselines/dom.generated.d.ts
```diff
@@ -2981,6 +2983,7 @@ interface Element extends Node, GlobalEventHandlers, ElementTraversal, NodeSelec
     webkitRequestFullscreen(): void;
     getElementsByClassName(classNames: string): NodeListOf<Element>;
     matches(selector: string): boolean;
+    getElementsByTagName(tagname: "picture"): NodeListOf<HTMLPictureElement>;
```
mhegazy: and here too.

baselines/dom.generated.d.ts
zhengbli: Actually the overloads wouldn't be ignored, even if they are defined after the more generic signature. The reason why the order is messed up is that the information in the json files are picked up after the emitting from the spec was done, so they are always placed at the end. I can fix it later, though it shouldn't affect the functionality now.

zhengbli:
LGTM.

PR #60, opened 2016-01-30
flesh out webcrypto

rapropos:
This should, among other things, sort of address [TS issue #3984](https://github.com/Microsoft/TypeScript/issues/3984). Fully making the return type a `Promise` instead of a `PromiseLike` is dependent on some sort of resolution of [the ability to target the ES6 library separately](https://github.com/Microsoft/TSJS-lib-generator/issues/47). Reference document used to create these typings is [the 2014.12 W3C WebCrypto API CR](https://www.w3.org/TR/WebCryptoAPI/).

Generally speaking, this should make code work that used to not compile (such as importing JSON web keys). The only *more* restrictive change, I believe, is making the `name` property of `Algorithm` no longer optional. It is explicitly documented as required in the W3C document.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4,7 +4,7 @@
 /////////////////////////////
 
 interface Algorithm {
-    name?: string;
+    name: string;
 }
 
 interface AriaRequestEventInit extends EventInit {
@@ -11478,18 +11478,24 @@ declare var StyleSheetPageList: {
 }
 
 interface SubtleCrypto {
-    decrypt(algorithm: string | Algorithm, key: CryptoKey, data: ArrayBufferView): PromiseLike<any>;
-    deriveBits(algorithm: string | Algorithm, baseKey: CryptoKey, length: number): PromiseLike<any>;
-    deriveKey(algorithm: string | Algorithm, baseKey: CryptoKey, derivedKeyType: string | Algorithm, extractable: boolean, keyUsages: string[]): PromiseLike<any>;
-    digest(algorithm: string | Algorithm, data: ArrayBufferView): PromiseLike<any>;
-    encrypt(algorithm: string | Algorithm, key: CryptoKey, data: ArrayBufferView): PromiseLike<any>;
-    exportKey(format: string, key: CryptoKey): PromiseLike<any>;
-    generateKey(algorithm: string | Algorithm, extractable: boolean, keyUsages: string[]): PromiseLike<any>;
-    importKey(format: string, keyData: ArrayBufferView, algorithm: string | Algorithm | null, extractable: boolean, keyUsages: string[]): PromiseLike<any>;
-    sign(algorithm: string | Algorithm, key: CryptoKey, data: ArrayBufferView): PromiseLike<any>;
-    unwrapKey(format: string, wrappedKey: ArrayBufferView, unwrappingKey: CryptoKey, unwrapAlgorithm: string | Algorithm, unwrappedKeyAlgorithm: string | Algorithm | null, extractable: boolean, keyUsages: string[]): PromiseLike<any>;
-    verify(algorithm: string | Algorithm, key: CryptoKey, signature: ArrayBufferView, data: ArrayBufferView): PromiseLike<any>;
-    wrapKey(format: string, key: CryptoKey, wrappingKey: CryptoKey, wrapAlgorithm: string | Algorithm): PromiseLike<any>;
+    decrypt(algorithm: string | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams, key: CryptoKey, data: BufferSource): PromiseLike<ArrayBuffer>;
+    deriveBits(algorithm: string | EcdhKeyDeriveParams | DhKeyDeriveParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, baseKey: CryptoKey, length: number): PromiseLike<ArrayBuffer>;
+    deriveKey(algorithm: string | EcdhKeyDeriveParams | DhKeyDeriveParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, baseKey: CryptoKey, derivedKeyType: string | AesDerivedKeyParams | HmacImportParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    digest(algorithm: AlgorithmIdentifier, data: BufferSource): PromiseLike<ArrayBuffer>;
+    encrypt(algorithm: string | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams, key: CryptoKey, data: BufferSource): PromiseLike<ArrayBuffer>;
+    exportKey(format: "jwk", key: CryptoKey): PromiseLike<JsonWebKey>;
+    exportKey(format: "raw" | "pkcs8" | "spki", key: CryptoKey): PromiseLike<ArrayBuffer>;
+    exportKey(format: string, key: CryptoKey): PromiseLike<JsonWebKey | ArrayBuffer>;
+    generateKey(algorithm: string, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKeyPair | CryptoKey>;
+    generateKey(algorithm: RsaHashedKeyGenParams | EcKeyGenParams | DhKeyGenParams, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKeyPair>;
+    generateKey(algorithm: AesKeyGenParams | HmacKeyGenParams | Pbkdf2Params, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    importKey(format: "jwk", keyData: JsonWebKey, algorithm: string | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | DhImportKeyParams, extractable:boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    importKey(format: "raw" | "pkcs8" | "spki", keyData: BufferSource, algorithm: string | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | DhImportKeyParams, extractable:boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    importKey(format: string, keyData: JsonWebKey | BufferSource, algorithm: string | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | DhImportKeyParams, extractable:boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    sign(algorithm: string | RsaPssParams | EcdsaParams | AesCmacParams, key: CryptoKey, data: BufferSource): PromiseLike<ArrayBuffer>;
+    unwrapKey(format: string, wrappedKey: BufferSource, unwrappingKey: CryptoKey, unwrapAlgorithm: AlgorithmIdentifier, unwrappedKeyAlgorithm: AlgorithmIdentifier, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    verify(algorithm: string | RsaPssParams | EcdsaParams | AesCmacParams, key: CryptoKey, signature: BufferSource, data: BufferSource): PromiseLike<boolean>;
+    wrapKey(format: string, key: CryptoKey, wrappingKey: CryptoKey, wrapAlgorithm: AlgorithmIdentifier): PromiseLike<ArrayBuffer>;
 }
 
 declare var SubtleCrypto: {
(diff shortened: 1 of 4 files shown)


baselines/dom.generated.d.ts
```diff
@@ -10439,18 +10439,20 @@ declare var StyleSheetPageList: {
...
-    digest(algorithm: string | Algorithm, data: ArrayBufferView): any;
-    encrypt(algorithm: string | Algorithm, key: CryptoKey, data: ArrayBufferView): any;
-    exportKey(format: string, key: CryptoKey): any;
-    generateKey(algorithm: string | Algorithm, extractable: boolean, keyUsages: string[]): any;
-    importKey(format: string, keyData: ArrayBufferView, algorithm: string | Algorithm, extractable: boolean, keyUsages: string[]): any;
-    sign(algorithm: string | Algorithm, key: CryptoKey, data: ArrayBufferView): any;
-    unwrapKey(format: string, wrappedKey: ArrayBufferView, unwrappingKey: CryptoKey, unwrapAlgorithm: string | Algorithm, unwrappedKeyAlgorithm: string | Algorithm, extractable: boolean, keyUsages: string[]): any;
-    verify(algorithm: string | Algorithm, key: CryptoKey, signature: ArrayBufferView, data: ArrayBufferView): any;
-    wrapKey(format: string, key: CryptoKey, wrappingKey: CryptoKey, wrapAlgorithm: string | Algorithm): any;
+    decrypt(algorithm: string | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams, key: CryptoKey, data: ArrayBuffer | ArrayBufferView): PromiseLike<ArrayBuffer>;
```
mhegazy: the algorithm parameter is of type `AlgorithmIdentifier`,  which can be defined as `string|Algorithm`. not sure i see the point of enumerating all possible Algorithm types here.

baselines/dom.generated.d.ts
```diff
@@ -10439,18 +10439,20 @@ declare var StyleSheetPageList: {
...
+    digest(algorithm: string | Algorithm, data: ArrayBuffer | ArrayBufferView): PromiseLike<ArrayBuffer>;
+    encrypt(algorithm: string | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams, key: CryptoKey, data: ArrayBuffer | ArrayBufferView): PromiseLike<ArrayBuffer>;
+    exportKey(format: string, key: CryptoKey): PromiseLike<ArrayBuffer | JsonWebKey>;
+    generateKey(algorithm: string, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKeyPair | CryptoKey>;
+    generateKey(algorithm: RsaHashedKeyGenParams | EcKeyGenParams | DhKeyGenParams, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKeyPair>;
+    generateKey(algorithm: AesKeyGenParams | HmacKeyGenParams | Pbkdf2Params, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    importKey(format: string, keyData: ArrayBuffer | ArrayBufferView | JsonWebKey, algorithm: string | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | DhImportKeyParams, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
```
mhegazy: looks like we should have a few overloads here:

```ts
importKey(format: "jwk", keyData: JsonWebKey , algorithm: AlgorithmIdentifier, extractable:boolean, keyUsages: string[]): ): PromiseLike<ArrayBuffer>;
importKey(format: "raw" | "pkcs8" | "spki", keyData: BufferSource , algorithm: AlgorithmIdentifier, extractable:boolean, keyUsages: string[]): ): PromiseLike<ArrayBuffer>;
importKey(format: string, keyData: JsonWebKey | BufferSource , algorithm: AlgorithmIdentifier, extractable:boolean, keyUsages: string[]): ): PromiseLike<ArrayBuffer>;

```

baselines/dom.generated.d.ts
mhegazy: that also applies to all the other ones. i would suggest adding a new type alias:
```ts
type AlgorithmIdentifier = string | Algorithm;
```

rapropos:
Thank you for taking the time to comment. The documentation lists available types for adding as "property, method, interface, constructor, or indexer". None of those looked like they would be appropriate for defining the AlgorithmIdentifier and BufferSource aliases that you suggest, so I am unclear as to the proper procedure for defining them.

mhegazy:
@zhengbli, how can we define a type alias?

baselines/dom.generated.d.ts
zhengbli: As a side note, the type `AlgorithmIdentifier` is already in the dom.generated.d.ts in master. So you can use that directly.

baselines/dom.generated.d.ts
```diff
@@ -12658,6 +12660,177 @@ declare var HTMLPictureElement: {
...
+}
+
+interface RsaOtherPrimesInfo {
+    r: string;
+    d: string;
+    t: string;
+}
+
+interface JsonWebKey {
+    kty: string;
```
zhengbli: Why `kty` is required while others are not? I didn't see any differences among these properties from the spec [here](https://dvcs.w3.org/hg/webcrypto-api/raw-file/tip/spec/Overview.html#JsonWebKey-dictionary)

zhengbli:
Sorry for the delay. With the latest updates in master, you can define a type alias like the follows:
```json
    {
        "kind": "typedef",
        "name": "IDBValidKey",
        "type": "number | string | Date | IDBArrayKey"
    }
```

baselines/dom.generated.d.ts
rapropos: @zhengbli: `kty` is required because of the comment in [RFC 7517 section 4.1](https://tools.ietf.org/html/rfc7517#section-4.1) that says:

> This member MUST be present in a JWK.

baselines/dom.generated.d.ts
rapropos: > not sure i see the point of enumerating all possible Algorithm types here

I'm not sure I fully understood the point of this comment until looking at it again now that we have the ability to write type aliases. Assuming you are asking "why are you listing all those xxxParams types instead of just using `AlgorithmIdentifier`?", the reason is that different algorithms have different allowed purposes, and I thought having them all enumerated would help users see at a glance which algorithms are usable for (e.g.) encrypting/decrypting and which are usable for signing/verifying.

rapropos:
Having the algorithm parameter to importKey be strictly an `AlgorithmIdentifier` instead of enumerating the possible types causes problems in my client code:
```
this.subtle.importKey("jwk", pubjwk,
                                {name: "RSASSA-PKCS1-v1_5", hash: {name: "SHA-256"}},
                                true, ["verify"])
                                .then((pubkey:CryptoKey) => {
                                    this.accessKeys.publicKey = pubkey;
                                })
```
```
Argument of type '{ name: string; hash: { name: string; }; }' is not assignable to parameter of type 'string | Algorithm'.
  Object literal may only specify known properties, and 'hash' does not exist in type 'string | Algorithm'.
```

rapropos:
CI seems to have failed due to some sort of network error. Is there a way I can try to rerun it?

zhengbli:
Seems like npm can't install tsd for now for some reason, tested on my local machine with the same result. I'll rerun the test later when it recovers.

rapropos:
Squashed and rebased.

baselines/dom.generated.d.ts
```diff
@@ -11473,18 +11473,23 @@ declare var StyleSheetPageList: {
...
+    decrypt(algorithm: string | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams, key: CryptoKey, data: BufferSource): PromiseLike<ArrayBuffer>;
+    deriveBits(algorithm: string | EcdhKeyDeriveParams | DhKeyDeriveParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, baseKey: CryptoKey, length: number): PromiseLike<ArrayBuffer>;
+    deriveKey(algorithm: string | EcdhKeyDeriveParams | DhKeyDeriveParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, baseKey: CryptoKey, derivedKeyType: string | AesDerivedKeyParams | HmacImportParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    digest(algorithm: AlgorithmIdentifier, data: BufferSource): PromiseLike<ArrayBuffer>;
+    encrypt(algorithm: string | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams, key: CryptoKey, data: BufferSource): PromiseLike<ArrayBuffer>;
+    exportKey(format: "jwk", key: CryptoKey): PromiseLike<JsonWebKey>;
+    exportKey(format: "raw" | "pkcs8" | "spki", key: CryptoKey): PromiseLike<ArrayBuffer>;
```
mhegazy: this needs a string overload as well. otherwise somehing like:

```ts
var format = "jwk";
c.exportKey(format); /// Error

```

baselines/dom.generated.d.ts
```diff
@@ -11473,18 +11473,23 @@ declare var StyleSheetPageList: {
...
-    verify(algorithm: string | Algorithm, key: CryptoKey, signature: ArrayBufferView, data: ArrayBufferView): PromiseLike<any>;
-    wrapKey(format: string, key: CryptoKey, wrappingKey: CryptoKey, wrapAlgorithm: string | Algorithm): PromiseLike<any>;
+    decrypt(algorithm: string | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams, key: CryptoKey, data: BufferSource): PromiseLike<ArrayBuffer>;
+    deriveBits(algorithm: string | EcdhKeyDeriveParams | DhKeyDeriveParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, baseKey: CryptoKey, length: number): PromiseLike<ArrayBuffer>;
+    deriveKey(algorithm: string | EcdhKeyDeriveParams | DhKeyDeriveParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, baseKey: CryptoKey, derivedKeyType: string | AesDerivedKeyParams | HmacImportParams | ConcatParams | HkdfCtrParams | Pbkdf2Params, extractable: boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    digest(algorithm: AlgorithmIdentifier, data: BufferSource): PromiseLike<ArrayBuffer>;
```
mhegazy: where is AlgorithmIdentifier defined?

rapropos: As @zhengbli mentioned in an earlier line note, AlgorithmIdentifier is defined in the stock library now.

baselines/dom.generated.d.ts
rapropos: Thank you. Added.

samuelhorwitz:
Thank you for this! I've been trying it locally and it fixes all my issues with WebCrypto breaking due to bad type definitions. However, I am now wrestling with `Promise` versus `PromiseLike` sadly. Anyway, hopefully this gets merged into TS soon.

rapropos:
@samuelhorwitz: Depending on your framework, the `PromiseLike`/`Promise` business may turn out to be a blessing in disguise. I ran into some subtle problems with Angular2, whereby it inserts some zone magic into its Promises that isn't sprinkled on the things coming back from WebCrypto. Having the return value be a `PromiseLike` reminds me that I need to wrap everything coming back from WebCrypto in `Promise.resolve()` in order for the change detection to fire correctly. In any event, whatever your Promise library, that `resolve` wrap should be sufficient to convert a `PromiseLike` to a `Promise`.

samuelhorwitz:
Yeah, I suppose so. I was just thrown through a loop because I guess there is no `catch` on promise likes and I just walked away from the code for a bit, frustrated about how just trying to upgrade Typescript led to so much issues. But anyway I found something else that might be wrong with the PR but I don't know enough about the specs to say: 

```
interface EcKeyImportParams {
    typedCurve: string;
}
```

This should probably be `namedCurve`. That's what Chrome expects, at least. Not sure if this is some discrepancy between spec and implementation (haven't looked at the spec).

rapropos:
Incorporated @samuelhorwitz's catch on `EcKeyImportParams` and resquashed.

baselines/dom.generated.d.ts
```diff
@@ -11473,18 +11473,24 @@ declare var StyleSheetPageList: {
...
+    importKey(format: "jwk", keyData: JsonWebKey, algorithm: string | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | DhImportKeyParams, extractable:boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    importKey(format: "raw" | "pkcs8" | "spki", keyData: BufferSource, algorithm: string | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | DhImportKeyParams, extractable:boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    importKey(format: string, keyData: JsonWebKey | BufferSource, algorithm: string | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | DhImportKeyParams, extractable:boolean, keyUsages: string[]): PromiseLike<CryptoKey>;
+    sign(algorithm: string | RsaPssParams | EcdsaParams | AesCmacParams, key: CryptoKey, data: BufferSource): PromiseLike<ArrayBuffer>;
+    unwrapKey(format: string, wrappedKey: BufferSource, unwrappingKey: CryptoKey, unwrapAlgorithm: AlgorithmIdentifier, unwrappedKeyAlgorithm: AlgorithmIdentifier, extractable: boolean, keyUsages: string[]): PromiseLike<ArrayBuffer>;
```
zhengbli: The [spec](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/unwrapKey) says the `unwrapKey` method returns `PromiseLike<CryptoKey>`

zhengbli:
:+1:

rapropos:
Rebased again. It sure would be convenient if the parser could process a bunch of separate files instead of having two monolithic ones. Would cut down on the merge conflicts a ton.

mhegazy:
:+1:

zhengbli:
@rapropos Thank you for your contribution!

PR #62, opened 2016-02-03
add 'sequence<T>' uniform variants to WebGLRenderingContext

### LINKED ISSUE #61: WebGLRenderingContext is missing sequence<T> variants
the [WebGL spec](https://www.khronos.org/registry/webgl/specs/latest/1.0/) lists variants of the `uniform*[if]v()` methods that take `sequence<GLfloat or long>`, for example:
````
void uniform1fv(WebGLUniformLocation? location, sequence<GLfloat> v);
````
these variants are missing from lib.d.ts, causing correct code to fail with

> Argument of type 'number[]' is not assignable to parameter of type 'Float32Array'.

piersh:
fixes #61

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -32,6 +32,7 @@ let rec DomTypeToTsType (objDomType: string) =
     | "EventListener" -> "EventListenerOrEventListenerObject"
     | "double" | "float" -> "number"
     | "Function" -> "Function"
+    | "GLfloat" | "GLint" -> "number"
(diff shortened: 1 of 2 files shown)


mhegazy:
Thanks for the PR. the input files should not be updated. please update one https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/addedTypes.json instead. also see https://github.com/Microsoft/TSJS-lib-generator/blob/master/README.md#input-files for more details.

PR #64, opened 2016-02-11
Change TouchEvent on() handlers  type

YuichiNukiyama:
Change TouchEvent on() handlers  type.
[TouchEvent on() handlers in "lib.d.ts" #5708](https://github.com/Microsoft/TypeScript/issues/5708)


mhegazy:
thanks!

PR #65, opened 2016-02-11
Add 'import' to HTMLLinkElements

pimterry:
With HTML Imports, <link> elements now have an 'import' property, providing access to the referenced document.

Spec: https://www.w3.org/TR/html-imports/#interface-import

I've left this optional, since it won't exist at all for browsers that don't support HTML Imports yet, and it's always null for browsers that do if the <link> doesn't have `rel='import'`.


mhegazy:
thanks.

mhegazy:
For future references, please file an issue on https://github.com/Microsoft/TypeScript for tracking purposes.

Issue #50, opened 2016-01-15
How to override extending targets of the interface


falsandtru:
How can I override extending targets of the interface?
https://github.com/Microsoft/TSJS-lib-generator/pull/51/files#diff-02015aece7bbbd0234e4c3c785287c1cR884

zhengbli:
It is not supported yet, but I'll try to make it available soon. Thank you for pointing out.

falsandtru:
Thank you. I want to apply my fixes to the next version 1.8.

PR #66, opened 2016-02-15
Add nodeParent property to HTML, Document and DocumentFragment

YuichiNukiyama:
Add parentNode property to Element, Document and DocumentFragment.
spec: [ParentNode](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode)
original issue: [ ParentNode interface isn't implemented in lib.d.ts #4400 ](https://github.com/Microsoft/TypeScript/issues/4400)


YuichiNukiyama:
I add childElementCoount : number, latter.

YuichiNukiyama:
In the Living standard, parentNode property of the Node interface return a Node. However, it not implemented because of abstract interface. In order to eliminate the error, should I change the parentNode of the return type?
[4.4. Interface Node](https://dom.spec.whatwg.org/#interface-node)

baselines/dom.generated.d.ts
```diff
@@ -12659,6 +12662,13 @@ declare var HTMLPictureElement: {
     new(): HTMLPictureElement;
 }
 
+interface parentNode {
```
mhegazy: use Camel Case. so `ParentNode`

baselines/dom.generated.d.ts
```diff
@@ -2677,6 +2678,7 @@ declare var Document: {
 }
 
 interface DocumentFragment extends Node, NodeSelector {
+    parentNode: parentNode;
```
mhegazy: this should be added to the extends clause and not a property. the spec says `DocumentFragment` implements `ParentNode`, and not that it has a `parentNode` property, all `Node`s have a parentNode alaredy. unless i am missing something

baselines/dom.generated.d.ts
```diff
@@ -12659,6 +12662,13 @@ declare var HTMLPictureElement: {
     new(): HTMLPictureElement;
 }
 
+interface parentNode {
+    children: HTMLCollection;
+    firstElementChild : Element;
+    lastElementChild : Element;
+    childElementCount: number;
+}
```
mhegazy: what about querySelector, and querySelctorAll?

inputfiles/addedTypes.json
```diff
@@ -400,5 +400,46 @@
...
         "signatures": ["getElementsByTagName(tagname: \"picture\"): NodeListOf<HTMLPictureElement>"]
+    },
+    {
+        "kind": "interface",
+        "name": "parentNode",
+        "flavor": "DOM",
+        "properties": [
+            {
+                "name": "children",
+                "type": "HTMLCollection"
```
mhegazy: i would add a "read-only" property here and make it "1". thought we do not support this in the script now, we should soon. this way we get the correct declaration later on. that also applies to the other properties.

mhegazy:
`parentNode: Node` seems fine to me,

baselines/dom.generated.d.ts
YuichiNukiyama: did it.

baselines/dom.generated.d.ts
YuichiNukiyama: I see. But, I can't override extending target.
[#50](https://github.com/Microsoft/TSJS-lib-generator/issues/50)

baselines/dom.generated.d.ts
YuichiNukiyama: HTML, Document and DocumentFragment interface already had been implimented NodeSelector interface. So, I think that no selector methods are required for ParentNode interface.

inputfiles/addedTypes.json
YuichiNukiyama: Should I send New Issue?

mhegazy: sure.

inputfiles/addedTypes.json
```diff
@@ -400,5 +400,40 @@
...
+                "type": "Element"
+            },
+            {
+                "name": "childElementCount",
+                "type": "number"
+            }
+        ]
+    },
+    {
+        "kind": "property",
```
mhegazy: i do not think this should be a property.

inputfiles/addedTypes.json
YuichiNukiyama: did it. #67

inputfiles/addedTypes.json
YuichiNukiyama: Is this interface, too?

mhegazy: it should be added to the extends list. i do not know how to do that though :) @zhengbli what is the right way to do this?

mhegazy:
@zhengbli any ideas?

zhengbli:
The support for adding `extends` is being added at #79 , once that merge in you can update the PR. Thanks!

YuichiNukiyama:
@zhengbli @mhegazy I update addType.json L404 to L438

YuichiNukiyama:
I missed operation. I close this PR and redo.

PR #69, opened 2016-02-24
Add minLength property to input & textarea

michaelbromley:
Fixes https://github.com/Microsoft/TypeScript/issues/7197
HTML5 spec:
https://www.w3.org/TR/html5/forms.html#the-maxlength-and-minlength-attributes

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5126,6 +5126,7 @@ interface HTMLInputElement extends HTMLElement {
       * Returns whether an element will successfully validate based on forms validation rules and constraints.
       */
     willValidate: boolean;
+    minLength: number;
     /**
       * Returns whether a form will validate when it is submitted, without having to submit it.
       */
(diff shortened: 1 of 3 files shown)


mhegazy:
why was inputfiles/comments.json edited. github seems to think of it as a binary file.. does it have a BOM at the top?

mhegazy:
looks good except the comments file that i am not sure what is going on with.

michaelbromley:
I thought it was necessary to edit comments.json in order to provide comments for the new properties I added. Is it not the case?

No idea why it shows as a binary. I was like that even before I made changes. Should I just revert that file to the original then?

mhegazy:
i think it is a recent change to the github UI that makes files show as binaries.

mhegazy:
can you merge in master again. i have removed the BOM from the file.

zhengbli:
:+1: Thanks @michaelbromley !

michaelbromley:
You're welcome - TS is making my life a lot easier, happy to contribute back :)

PR #70, opened 2016-02-25
Fix IDBIndex interface for openCursor/openKeyCursor

falsandtru:
issue: https://github.com/Microsoft/TypeScript/issues/7218


DanielRosenwasser:
See my comment on https://github.com/Microsoft/TypeScript/issues/7218#issuecomment-188668347.

falsandtru:
@mhegazy @DanielRosenwasser @zhengbli can you merge this fix?

zhengbli:
@falsandtru I just noticed this change is covered by #72 , which is a bigger change that updates the XML spec file.

falsandtru:
Oh nice update. I follow your fixes.

PR #72, opened 2016-02-27
Update the XML spec

zhengbli:
This PR updates the XML spec with the Windows 10 TH2 release version (10586).

[Review on Reviewable](https://reviewable.io/reviews/microsoft/tsjs-lib-generator/72)


TS.fsx
```diff
@@ -49,22 +50,26 @@ let rec DomTypeToTsType (objDomType: string) =
...
-            elif (objDomType.Contains(" or ")) then
+            elif objDomType.Contains(" or ") then
                 let allTypes = objDomType.Trim('(', ')').Split([|" or "|], StringSplitOptions.None)
-                                |> Array.map DomTypeToTsType
+                                |> Array.map (fun t -> DomTypeToTsType (t.Trim('?', ' ')))
                 if Seq.contains "any" allTypes then "any" else String.concat " | " allTypes
             else
                 // Check if is array type, which looks like "sequence<DOMString>"
-                let genericMatch = Regex.Match(objDomType, @"^(\w+)<(\w+)>$")
+                let unescaped = objDomType.Replace("&lt;", "<").Replace("&gt;", ">")
```
DanielRosenwasser: [HttpUtility.HtmlDecode](https://msdn.microsoft.com/en-us/library/7c5fyk1k(v=vs.110).aspx)

TS.fsx
```diff
@@ -49,22 +50,26 @@ let rec DomTypeToTsType (objDomType: string) =
...
+                let genericMatch = Regex.Match(unescaped, @"^(\w+)<(\w+)>$")
```
DanielRosenwasser: What about multiple type arguments? What about nested type arguments?

baselines/webworker.generated.d.ts
```diff
@@ -344,8 +349,8 @@ interface IDBIndex {
     count(key?: any): IDBRequest;
     get(key: any): IDBRequest;
     getKey(key: any): IDBRequest;
-    openCursor(range?: IDBKeyRange, direction?: string): IDBRequest;
-    openKeyCursor(range?: IDBKeyRange, direction?: string): IDBRequest;
+    openCursor(range?: any, direction?: string): IDBRequest;
+    openKeyCursor(range?: any, direction?: string): IDBRequest;
```
DanielRosenwasser: This is too lax, as I mentioned in https://github.com/Microsoft/TypeScript/issues/7218#issuecomment-188668347

mhegazy:
LGTM

TS.fsx
zhengbli: The nested case should be taken care of, as the implementation is recursive. In line 69 the `paramName` is obtained via another run of `DomTypeToTsType` using the matched part, so even if it has a type argument, it will be recognized.

As for the case with multiple type arguments, the logic would become much more complex, because it is non-trivial to tell whether ',' delimits two types or is just inside a type argument. Up to now I haven't seen the case appeared yet in the spec, it might be worth another PR to address the issue.

falsandtru:
:+1:

baselines/webworker.generated.d.ts
```diff
@@ -383,7 +388,7 @@ interface IDBObjectStore {
     deleteIndex(indexName: string): void;
     get(key: any): IDBRequest;
     index(name: string): IDBIndex;
-    openCursor(range?: any, direction?: string): IDBRequest;
+    openCursor(range?: IDBKeyRange | IDBValidKey, direction?: string): IDBRequest;
     put(value: any, key?: any): IDBRequest;
```
falsandtru: also should be `put(value: any, key?: IDBValidKey): IDBRequest;`. Other methods are same.

zhengbli: Fixed now

PR #73, opened 2016-03-04
Update build.cmd to avoid usage of curl, and to instead shell out to …

aholmes:
…PowerShell, which uses Invoke-WebRequest.

This fixes issues on systems that either don't have Curl installed, or Curl is not in PATH.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -5,7 +5,7 @@ if not exist .paket mkdir .paket
 
 if not exist .paket/paket.bootstrapper.exe (
   @echo "Installing Paket"
-  Powershell.exe -File download.ps1
+  Powershell.exe -Command "Invoke-WebRequest -Uri https://github.com/fsprojects/Paket/releases/download/2.12.5/paket.bootstrapper.exe -OutFile .paket\paket.bootstrapper.exe"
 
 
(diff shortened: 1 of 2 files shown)


build.cmd
```diff
@@ -1,10 +1,12 @@
...
 cls
 
-if not exist .paket (
+if not exist .paket mkdir .paket
+
+if not exist .paket/paket.bootstrapper.exe (
   @echo "Installing Paket"
-  mkdir .paket
-  curl https://github.com/fsprojects/Paket/releases/download/2.12.5/paket.bootstrapper.exe -L --insecure -o .paket\paket.bootstrapper.exe
+  Powershell.exe -File download.ps1
```
zhengbli: No need to create a new file. 
```cmd
Powershell.exe -Command "Invoke-WebRequest -Uri https://github.com/fsprojects/Paket/releases/download/2.12.5/paket.bootstrapper.exe -OutFile .paket\paket.bootstrapper.exe"
```
should work

aholmes:
Thanks for the tip!

zhengbli:
LGTM. @aholmes Thanks for contributing!

PR #74, opened 2016-03-07
Add HTMLFileInputElement and update HTMLInputElement.

aholmes:
This change introduces a new interface for file input elements and updates the "files" object of all other input elements.

The existing "files" object is `undefined` when the input type is not a "file" type. The new interface correctly reflects when the "files" object is a FileList.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5636,9 +5636,9 @@ interface HTMLInputElement extends HTMLElement {
     defaultValue: string;
     disabled: boolean;
     /**
-      * Returns a FileList object on a file type input object.
+      * Returns a FileList object on a file type input object. This is null for all other input objects.
       */
-    files: FileList;
+    files: void;
     /**
       * Retrieves a reference to the form that the object is embedded in. 
       */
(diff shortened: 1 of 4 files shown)


mhegazy:
I do not see any reference to the new HTMLFileInputElement in the spec. can you provide some context here?
Also in the future logging an issue on https://github.com/Microsoft/TypeScript/issues first is the correct way to proceed.

baselines/dom.generated.d.ts
```diff
@@ -13668,6 +13668,15 @@ interface ClipboardEventInit extends EventInit {
 interface IDBArrayKey extends Array<IDBValidKey> {
 }
 
+interface HTMLFileInputElement extends HTMLInputElement {
+    files: FileList;
+}
+
+declare var HTMLFileInputElement: {
```
mhegazy: this introduces a new value `HTMLFileInputElement` that does not exisit any where.

aholmes:
My mistake, on both counts. I'll make an issue first next time.

Regarding the type, I wasn't aware the types in lib.d.ts matched the types from the spec and created this type willy nilly. I'm okay with this pull request being rejected for that reason, however, I would like to ask if there is a way to alter HTMLInputElement's signature when it's type is "file".

mhegazy:
From the sepc: https://html.spec.whatwg.org/multipage/forms.html#dom-input-files

> The files IDL attribute allows scripts to access the element's selected files. On getting, if the IDL attribute applies, it must return a FileList object that represents the current selected files. The same object must be returned until the list of selected files changes. If the IDL attribute does not apply, then it must instead return **null**. [FILEAPI]

so it seems that the right thing to do here is to define `files: FileList | null`. that needs the nullable types proposal to be in, see https://github.com/Microsoft/TypeScript/pull/7140.

I would punt on this PR. but please file an issue to track chainging the definition to include `null` when  https://github.com/Microsoft/TypeScript/pull/7140 is merged.

PR #79, opened 2016-03-21
Add support for extends

zhengbli:
Help with #66 

Now one can add a base interface to an interface as follows:
```json
    {
        "kind": "extends",
        "baseInterface": "ParentNode",
        "interface": "Document"
    }
```

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -381,8 +381,15 @@ let EmitNamedConstructors () =
 
 let EmitInterfaceDeclaration (i:Browser.Interface) =
     Pt.printl "interface %s" i.Name
-    match i.Extends::(List.ofArray i.Implements) with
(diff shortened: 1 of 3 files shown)


TS.fsx
```diff
@@ -381,8 +381,15 @@ let EmitNamedConstructors () =
 
 let EmitInterfaceDeclaration (i:Browser.Interface) =
```
sandersn: Looks like this only works with interfaces. Is there a need for classes too? (I'm not familiar with lib.d.ts so could be there's not.)

zhengbli: There are no classes in `dom.generated.d.ts`, all types in the spec translate to an interface (instance side) and a declared var constructor function (static side).

sandersn:
:+1:

PR #80, opened 2016-03-21
Override WebGLRenderingContext methods

### LINKED ISSUE #61: WebGLRenderingContext is missing sequence<T> variants

zhengbli:
Fix #61


mhegazy:
:+1:

PR #86, opened 2016-04-01
Add "readonly" modifier for property and constants

### LINKED ISSUE #55: Add readonly modifier to DOM declarations
This is gated on: https://github.com/Microsoft/TypeScript/pull/6532

The webidl file already provide this information, we need to translate that in the script:

for instance `read-only`: 
https://github.com/Microsoft/TSJS-lib-generator/blob/ea717dddb01dbeb6386b844f667194f71c7d7165/inputfiles/browser.webidl.xml#L408

and `constant` : https://github.com/Microsoft/TSJS-lib-generator/blob/ea717dddb01dbeb6386b844f667194f71c7d7165/inputfiles/browser.webidl.xml#L315

zhengbli:
Fix #55

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -212,7 +212,10 @@ let EmitEnums () =
 
 let EmitProperties flavor prefix (emitScope: EmitScope) (i: Browser.Interface)=
     let emitPropertyFromJson (p: ItemsType.Root) =
-        let readOnlyModifier = if p.Readonly.IsSome && p.Readonly.Value = true && prefix = "" then "readonly " else ""
+        let readOnlyModifier =
+            match p.Readonly with
+            | Some(true) -> "readonly "
+            | _ -> ""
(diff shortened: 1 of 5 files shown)


TS.fsx
```diff
@@ -212,7 +212,8 @@ let EmitEnums () =
 
 let EmitProperties flavor prefix (emitScope: EmitScope) (i: Browser.Interface)=
     let emitPropertyFromJson (p: ItemsType.Root) =
-        Pt.printl "%s%s: %s;" prefix p.Name.Value p.Type.Value
+        let readOnlyModifier = if p.Readonly.IsSome && p.Readonly.Value = true && prefix = "" then "readonly " else ""
```
DanielRosenwasser: What about

```fsharp
match p.Readonly with
    Some(true) when prefix = "" -> "readonly "
    _ -> ""

DanielRosenwasser:
:+1:

mhegazy:
yaaay. :+1:

PR #91, opened 2016-04-13
Add CanvasRenderingContext2D.msImageSmoothingEnabled with vender-prefix

YuichiNukiyama:
fix: [#7846](https://github.com/Microsoft/TypeScript/issues/7846)
spec: [CanvasRenderingContext2D.imageSmoothingEnabled](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled)


mhegazy:
thanks!

PR #98, opened 2016-04-29
 CSSStyleDeclaration.properttyName should return only value

### LINKED ISSUE #97: CSSStyleDeclaration.*** should return only `string` value.
By the spec, `CSSStyleDeclaration.properttyName` should only return a `string` value, not be `null`.
- https://www.w3.org/TR/cssom-1/#dom-cssstyledeclaration-getpropertyvalue
- https://www.w3.org/TR/cssom-1/#dom-cssstyledeclaration-camel-cased-attribute

But the current code in this repository defines them as `properttyName: string | null`. This is incorrect.

saneyuki:
Fix #97


YuichiNukiyama:
`browser.webidl.xml` is generated by Edge. You must not edit this file (see [Input Files](https://github.com/Microsoft/TSJS-lib-generator/blob/master/README.md#input-files)).
If you want to override doms, you should add changes to `overridingTypes.json ` (see [Contributing  lib.d.ts  fixes](https://github.com/Microsoft/TypeScript/blob/master/CONTRIBUTING.md#contributing-libdts-fixes)).

mhegazy:
as noted by @YuichiNukiyama this file should not be changed. please see https://github.com/Microsoft/TSJS-lib-generator/blob/master/README.md and https://github.com/Microsoft/TypeScript/blob/master/CONTRIBUTING.md#contributing-libdts-fixes
for more information.

saneyuki:
Ah, Sorry. This was my miss.

PR #95, opened 2016-04-27
Node.insertBefore requires two arguments

ntrrgc:
Fixes https://github.com/Microsoft/TypeScript/issues/8320

PR #99, opened 2016-04-29
Second parameter of insertBefore should accept null

ivogabe:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore), the second parameter of insertBefore is not optional (which is was fixed in #95), but it can accept `null`. Under `strictNullChecks` this is currently not allowed, this PR fixes that.


mhegazy:
:+1:

PR #102, opened 2016-05-03
specify return type for querySelector

YuichiNukiyama:
fix [querySelector return type could be more specific for single-element selectors #8114](https://github.com/Microsoft/TypeScript/issues/8114)


mhegazy:
:+1:

zhengbli:
Thanks @YuichiNukiyama !

yuttie:
@YuichiNukiyama Thanks for your great work.

I found I cannot select an `SVGTitleElement` (https://developer.mozilla.org/en-US/docs/Web/API/SVGTitleElement) by querying "title" with SVGSVGElement's `querySelector()`.  I think the problem is related to this PR.

I think it would be nice if `querySelector("title")` returns a value of type `HTMLTitleElement | SVGTitleElement`.

How do you think?

YuichiNukiyama:
@yuttie 
A good point. However, cast is necessary, so it is hard to use.

PR #103, opened 2016-05-05
Change the type of HTMLSelectElement.options

### LINKED ISSUE #68: Typing of HTMLSelectElement
There are several properties of `HTMLSelectElement` that could be typed more specifically.  For example [`options`](https://github.com/Microsoft/TypeScript/blob/0f67f4b6f1589756906782f1ac02e6931e1cff13/src/lib/dom.generated.d.ts#L6147) is typed as `HTMLCollection` when it could be typed as `HTMLOptionsCollection` (which isn't currently typed) which it's index would be of type `HTMLOptionElement`. (see: [MDN HTMLSelectElement](https://developer.mozilla.org/en/docs/Web/API/HTMLSelectElement))

I would also argue that `HTMLCollection` should actually return `HTMLElement` instead of `Element`.

Note, this "broke" some things with TS 1.8.2 because previously the index returned `any` thereby allowing a lot of code to "just compile".

zhengbli:
Fix #68

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -404,14 +404,22 @@ let EmitNamedConstructors () =
 
 let EmitInterfaceDeclaration (i:Browser.Interface) =
     Pt.printl "interface %s" i.Name
-    let extendsFromSpec =
-        match i.Extends::(List.ofArray i.Implements) with
-        | [""] | [] | ["Object"] -> []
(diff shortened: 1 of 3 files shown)


mhegazy:
:+1:

PR #106, opened 2016-05-14
try to fix https://github.com/Microsoft/TypeScript/issues/8202

plantain-00:
Fixes https://github.com/Microsoft/TypeScript/issues/8202

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -919,7 +919,7 @@ interface ApplicationCache extends EventTarget {
     oncached: (ev: Event) => any;
     onchecking: (ev: Event) => any;
     ondownloading: (ev: Event) => any;
-    onerror: (ev: Event) => any;
+    onerror: (ev: ErrorEvent) => any;
     onnoupdate: (ev: Event) => any;
     onobsolete: (ev: Event) => any;
     onprogress: (ev: ProgressEvent) => any;
@@ -2512,7 +2512,7 @@ interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEven
       * Fires when an error occurs during object loading.
       * @param ev The event.
       */
-    onerror: (ev: Event) => any;
+    onerror: (ev: ErrorEvent) => any;
     /**
       * Fires when the object receives focus. 
       * @param ev The event.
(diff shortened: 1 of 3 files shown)


baselines/dom.generated.d.ts
```diff
@@ -9438,7 +9438,7 @@ declare var RTCRtpSender: {
 }
 
 interface RTCSrtpSdesTransport extends EventTarget {
-    onerror: ((ev: Event) => any) | null;
```
mhegazy: we should not remove the ` | null` here.

baselines/dom.generated.d.ts
```diff
@@ -9283,7 +9283,7 @@ declare var RTCDTMFToneChangeEvent: {
 
 interface RTCDtlsTransport extends RTCStatsProvider {
     ondtlsstatechange: ((ev: RTCDtlsTransportStateChangedEvent) => any) | null;
-    onerror: ((ev: Event) => any) | null;
```
mhegazy: we should not remove the ` | null` here.

baselines/dom.generated.d.ts
```diff
@@ -9338,7 +9338,7 @@ declare var RTCIceCandidatePairChangedEvent: {
 
 interface RTCIceGatherer extends RTCStatsProvider {
     readonly component: string;
-    onerror: ((ev: Event) => any) | null;
```
mhegazy: and here

baselines/dom.generated.d.ts
```diff
@@ -9397,7 +9397,7 @@ declare var RTCIceTransportStateChangedEvent: {
 }
 
 interface RTCRtpReceiver extends RTCStatsProvider {
-    onerror: ((ev: Event) => any) | null;
```
mhegazy: and here.

baselines/dom.generated.d.ts
```diff
@@ -9417,7 +9417,7 @@ declare var RTCRtpReceiver: {
 }
 
 interface RTCRtpSender extends RTCStatsProvider {
-    onerror: ((ev: Event) => any) | null;
```
mhegazy: and here.

mhegazy:
@zhengbli is there a better way to do this, may be on the generation script side?

zhengbli:
This is easier to change from the script. I'll send up a PR shortly.

mhegazy:
closing in favor of https://github.com/Microsoft/TSJS-lib-generator/pull/108

PR #107, opened 2016-05-14
try to fix https://github.com/Microsoft/TypeScript/issues/8101

plantain-00:
Fixes https://github.com/Microsoft/TypeScript/issues/8101

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -13597,6 +13597,7 @@ interface WindowTimers extends Object, WindowTimersExtension {
     clearTimeout(handle: number): void;
     setInterval(handler: any, timeout?: any, ...args: any[]): number;
     setTimeout(handler: any, timeout?: any, ...args: any[]): number;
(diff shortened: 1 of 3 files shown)


mhegazy:
I left a comment about the return type in https://github.com/Microsoft/TypeScript/issues/8101#issuecomment-219497942. it will address the original issue, but will make clearer (i think :)).

mhegazy:
let's change the return type from `any` to `void`, and we should be good to go.

plantain-00:
Sorry for delay, a little busy.

mhegazy:
thanks! might want to add a similar change to setImmediate as well.

sandersn:
Shouldn't the new overload be first since it's more specific?  Otherwise I think the compiler will still always choose the original overload.

plantain-00:
Since `setImmediate` will need a similar overload as `setTimeout`, what about `setInterval`? a similar overload too?

mhegazy:
sounds about right.

PR #108, opened 2016-05-16
Fix OnError handlers

zhengbli:
Correct the event type for event handler properties.

Related to https://github.com/Microsoft/TSJS-lib-generator/pull/106


mhegazy:
:+1:

mhegazy:
Fixes Microsoft/TypeScript#8202

PR #109, opened 2016-05-19
files attribute of HTMLInputElement is nullable 

### LINKED ISSUE #75: Change HTMLInputElement's files attribute to match spec
lib.d.ts currently shows HTMLInputElement as having a `files` attribute which is always a `FileList`.

According to the spec, `files` is only a `FileList` when the IDL attribute applies, and null when it does not. https://html.spec.whatwg.org/multipage/forms.html#dom-input-files

The `files` attribute only applies when the input's type is `File Upload (file=type)`. https://html.spec.whatwg.org/multipage/forms.html#concept-input-apply

See rejected PR #74 for more information. Additionally, non-nullable types in TS may assist with making this change. Microsoft/TypeScript#7140

YuichiNukiyama:
fix #75


mhegazy:
:+1:

zhengbli:
Thanks!

PR #110, opened 2016-05-19
Fix event types of XMLHttpRequest

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/8691

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -13134,7 +13134,7 @@ declare var XMLDocument: {
 
 interface XMLHttpRequest extends EventTarget, XMLHttpRequestEventTarget {
     msCaching: string;
-    onreadystatechange: (ev: ProgressEvent) => any;
+    onreadystatechange: (ev: Event) => any;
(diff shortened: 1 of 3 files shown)


falsandtru:
How can I override `addEventListener` signatures?

falsandtru:
@zhengbli Are you supported signature overriding like this?

mhegazy:
ping @zhengbli

zhengbli:
Overriding the signatures of `addEventListener` is tricky because the way string overloads works in TypeScript. The overload signature may come from the spec, or inherited from base classes. However if it has its own overload signatures, all inherited ones will be wiped. So the normal way of overriding properties may not work properly here. I'll come up with a fix shortly.

falsandtru:
Any updates?

zhengbli:
Actually I found it is easier to change the emitting F# code. I sent another PR at https://github.com/Microsoft/TSJS-lib-generator/pull/144, sorry for the delay.

falsandtru:
LGTM 👍

PR #111, opened 2016-05-19
Add closest method of Element

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/8699

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -209,6 +209,11 @@
         "signatures": ["matches(selector: string): boolean"]
     },
     {
+        "kind": "method",
+        "interface": "Element",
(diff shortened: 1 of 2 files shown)


inputfiles/addedTypes.json
```diff
@@ -209,6 +209,11 @@
         "signatures": ["matches(selector: string): boolean"]
     },
     {
+        "kind": "method",
+        "interface": "Element",
+        "signatures": ["closest(selector: string): Element"]
```
mhegazy: `Element | null`

falsandtru: :+1:

mhegazy:
thanks!

PR #112, opened 2016-05-19
add override function for setImmediate and setInterval

plantain-00:
add override function for setImmediate and setInterval;
change the order of functions of setTimeout.


mhegazy:
:+1:

PR #118, opened 2016-06-07
add fetch api

plantain-00:
Fixes https://github.com/Microsoft/TypeScript/issues/8966.
The definitions are from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/whatwg-fetch/whatwg-fetch.d.ts

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -13008,6 +13008,7 @@ interface Window extends EventTarget, WindowTimers, WindowSessionStorage, Window
     webkitConvertPointFromNodeToPage(node: Node, pt: WebKitPoint): WebKitPoint;
     webkitConvertPointFromPageToNode(node: Node, pt: WebKitPoint): WebKitPoint;
     webkitRequestAnimationFrame(callback: FrameRequestCallback): number;
+    fetch(url: string|Request, init?: RequestInit): PromiseLike<Response>;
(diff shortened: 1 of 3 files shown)


mhegazy:
👍

mhegazy:
@zhengbli would this be autogenerated if we took a drop of the Edge definition files today?

zhengbli:
@mhegazy Yes. The latest definite in the redstone build has a `GlobalFetch` interface, and `window` implements it.

baselines/dom.generated.d.ts
```diff
@@ -13855,6 +13856,85 @@ interface JsonWebKey {
...
+    new(input: string|Request, init?:RequestInit): Request;
+}
+
+interface Response extends Body {
+    type: ResponseType;
+    status: number;
+    ok: boolean;
+    statusText: string;
+    headers: Headers;
+    clone(): Response;
```
zhengbli: `Request` should also have a `clone` method that returns `Request`

mhegazy:
then we should just take a new drop, instead of hand authoring these definitions..

zhengbli:
As mentioned above, the new spec files that we haven't used yet contains the definition of `fetch` APIs, so it might be more consistent if we wait to take the drop. Although we should speed up the process to something like a monthly update. I'll talk to the Edge team to see if there are anything blocking that from happening. 

Closing this issue for now.

PR #119, opened 2016-06-10
Check type nullability in json files

### LINKED ISSUE #96: Check overridden items for nullable types
Currently we don't check the nullability of types for items in the `overridingTypes.json`. We should check the nullability of the corresponding types in the spec and complement the content.

zhengbli:
Fixes #96


mhegazy:
👍

mhegazy:
thanks!

PR #121, opened 2016-06-11
add MouseWheelEvent

YuichiNukiyama:
fix #[9071](https://github.com/Microsoft/TypeScript/issues/9071).


mhegazy:
:+1:

mhegazy:
had a chat with @zhengbli offline, this should be added to addedTypes instead.

something like:

```ts
{
        "kind": "typedef",
        "flavor": "Worker",
        "name": "MouseWheelEvent",
        "type": "WheelEvent"
}, 
```

YuichiNukiyama:
@mhegazy Ah, I don't know we can generate types! Thanks.

zhengbli:
:+1: Thanks!

PR #124, opened 2016-06-18
Fix missing types in worker

zhengbli:
Fixes https://github.com/Microsoft/TypeScript/issues/9248


DanielRosenwasser:
Seems reasonable 👍  :shipit:

PR #125, opened 2016-06-23
Add Math property to Window

zuzusik:
Fixes https://github.com/Microsoft/TypeScript/issues/9325

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -12973,6 +12973,7 @@ interface Window extends EventTarget, WindowTimers, WindowSessionStorage, Window
     readonly window: Window;
     URL: typeof URL;
     Blob: typeof Blob;
+    Math: typeof Math;
     alert(message?: any): void;
     blur(): void;
     cancelAnimationFrame(handle: number): void;
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -84,6 +84,13 @@
     },
     {
         "kind": "property",
+        "interface": "Window",
+        "exposeGlobally": false,
+        "name": "Math",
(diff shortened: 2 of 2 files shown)


mhegazy:
As noted in https://github.com/Microsoft/TypeScript/issues/9325#issuecomment-228126856, i do not see why we need to define this but not other ones, e.g. `NaN`, `Array`, `Date`, `Number`, `String`, `HTMLElement`, etc..

SaschaNaz:
From https://github.com/Microsoft/TypeScript/issues/9325#issuecomment-228126856:

>The standard library does not support using Window as the global polluter. i.e. adding members on Window are not reflected on the global and adding members on the global are not reflected on Window.

Not sure what this means, can you elaborate more? I thought JSON, Math, RegExp, etc are all accessible via `window` and `delete window.RegExp` makes the global member inaccessible.

zuzusik:
@mhegazy then I suppose you should also delete Blob and URL from window?

In general this sounds strange to me. If these props exist on window, why shouldn't it be described in TypeScript definitions?

SaschaNaz:
@zuzusik Web specifications explicitly say `Exposed=Window` in [their WebIDL](http://www.w3.org/TR/2015/WD-FileAPI-20150421#file) but JS standard libraries do not. However, still not really understand TypeScript policy yet...

mhegazy:
This file is autogrenerated mostly, so we shown ones marked as such. i am fine adding more to `window` but do not see why add `Math`, but nothing else.

mhegazy:
As i have noted [earlier](https://github.com/Microsoft/TSJS-lib-generator/pull/125#issuecomment-231832031), do not see why `Math` is any different from other declarations. either we do them all or nothing.

PR #126, opened 2016-06-28
Add this parameters

sandersn:
Add this-parameters to event handler so that intellisense improves when assigning literals, and assignability checking improves when assigning functions that have this parameters themselves.

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -243,6 +243,11 @@ let EmitEnums () =
     let emitEnum (e: Browser.Enum) = Pt.printl "declare var %s: string;" e.Name
     browser.Enums |> Array.iter emitEnum
 
+let EmitEventHandlerThis = function
+| (_, "") -> "this: this"
+| (Worker, _) -> "this: WorkerGlobalScope"
+| _ -> "this: Window"
+
 let EmitProperties flavor prefix (emitScope: EmitScope) (i: Browser.Interface)=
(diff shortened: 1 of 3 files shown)


sandersn:
Looks like the linter failed on the branch build. It's in a file that this PR doesn't change -- any idea why?

sandersn:
@yuit points out that the break is from our peer dependency on tslint

zhengbli:
Re-running CI to make sure the tests pass.

TS.fsx
```diff
@@ -243,6 +243,11 @@ let EmitEnums () =
     let emitEnum (e: Browser.Enum) = Pt.printl "declare var %s: string;" e.Name
     browser.Enums |> Array.iter emitEnum
 
+let EmitEventHandlerThis = function
```
zhengbli: We have a function `GetGlobalPollutor` which will get the global pollutor type according to the spec. So it will be
```fs
let EmitEventHandlerThis = function
| _, "" -> "this: this"
| flavor, _ -> match GetGlobalPollutor flavor with
               | Some pollutor -> "this: " + pollutor
               | _ -> ""
```

sandersn: Fixed. I got rid of the outer `match` since it wasn't doing much work any more.

zhengbli:
👍

PR #135, opened 2016-07-22
changed EcKeyGenParams from typedCurve to namedCurve to fix issue

jo-wil:
If you leave the parameters named typedCurve it compiles clean but when run in the browser (Chrome 51) it fails with the following error Uncaught (in promise) TypeError: EcKeyGenParams: namedCurve: Missing or not a string(…).

If you change it to namedCurve tsc will have an error but the generated code runs without issue. 

This pull request should fix this so that you have clean compilation and error free generated code.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -14098,7 +14098,7 @@ interface EcdsaParams extends Algorithm {
 }
 
 interface EcKeyGenParams extends Algorithm {
-    typedCurve: string;
+    namedCurve: string;
 }
 
 interface EcKeyAlgorithm extends KeyAlgorithm {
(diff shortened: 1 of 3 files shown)


mhegazy:
These files should not be changed manually. instead change the source files at: https://github.com/Microsoft/TSJS-lib-generator/tree/master/inputfiles

Please see the readme: https://github.com/Microsoft/TSJS-lib-generator/blob/master/README.md#contribution-guidelines for more information.

mhegazy:
also pinging  @rapropos

jo-wil:
@mhegazy I'm sorry but I'm confused by what you mean by these files should not be changed manually. I also apologize for the questions as this is my first pull request to this repo.

Per step 2 I changed the addedTypes.json file to reflect the needed type change.

Per step 3 I ran build.sh locally to obtain the new generated files.

Per step 4 I copied them to the baselines directory.

I then committed these changes and pushed to my branch, then I submitted the pull request.

If you could advise which step I did incorrectly that would be very helpful.

mhegazy:
OK.. Looks like I was asleeep when I made that comment. Sorry about that.

PR #139, opened 2016-08-10
Make the global var `name` of type `never`

zhengbli:
Fix https://github.com/Microsoft/TypeScript/issues/9850


DanielRosenwasser:
:sparkles: 

Also something's going on with compiling TSServer apparently?

zhengbli:
It seems to be caused by an error in the unit test code. Sent https://github.com/Microsoft/TypeScript/pull/10260 to fix.

PR #142, opened 2016-08-26
Fix the History api

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/10059

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -7536,11 +7536,12 @@ declare var HashChangeEvent: {
 interface History {
     readonly length: number;
     readonly state: any;
-    back(distance?: any): void;
-    forward(distance?: any): void;
-    go(delta?: any): void;
-    pushState(statedata: any, title?: string, url?: string): void;
-    replaceState(statedata: any, title?: string, url?: string): void;
+    scrollRestoration: ScrollRestoration;
+    back(): void;
+    forward(): void;
+    go(delta?: number): void;
+    pushState(data: any, title: string, url?: string | null): void;
+    replaceState(data: any, title: string, url?: string | null): void;
 }
 
 declare var History: {
@@ -14623,3 +14624,4 @@ type ScrollLogicalPosition = "start" | "center" | "end" | "nearest";
 type IDBValidKey = number | string | Date | IDBArrayKey;
 type BufferSource = ArrayBuffer | ArrayBufferView;
 type MouseWheelEvent = WheelEvent;
+type ScrollRestoration = "auto" | "manual";
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -1175,5 +1175,17 @@
         "kind": "extends",
         "baseInterface": "ParentNode",
         "interface": "DocumentFragment"
+    },
+    {
+        "kind": "typedef",
+        "name": "ScrollRestoration",
+        "flavor": "DOM",
+        "type": "\"auto\" | \"manual\""
+    },
+    {
+        "kind": "property",
+        "interface": "History",
+        "name": "scrollRestoration",
+        "type": "ScrollRestoration"
     }
 ]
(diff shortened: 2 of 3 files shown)


falsandtru:
@zhengbli How can I remove `declare var History`?

zhengbli:
Why do you need to remove the `declare var History`? We need a constructor for the `typeof` operator to work, also according to https://github.com/Microsoft/TypeScript/issues/10059#issuecomment-236814714, `console.log(History)` does print something in Chrome.

falsandtru:
Yes, I fixed it. Function signatures are needed for the `instanceof` function as follows.

```ts
history instanceof History
```

falsandtru:
Ah, no, removed.

falsandtru:
`history` is defined in the following spec, but `History` is not.

https://html.spec.whatwg.org/multipage/browsers.html#the-window-object

zhengbli:
The constructor is indeed defined in major browsers, it is just forbidden to call. For comparability I would suggest just leave it there. @mhegazy thoughts?
capture
capture1

falsandtru:
I think `History` variable should be defined by users because it doesn't standardized by spec.

baselines/dom.generated.d.ts
```diff
@@ -7536,16 +7536,12 @@ declare var HashChangeEvent: {
...
     readonly state: any;
-    back(distance?: any): void;
-    forward(distance?: any): void;
-    go(delta?: any): void;
-    pushState(statedata: any, title?: string, url?: string): void;
-    replaceState(statedata: any, title?: string, url?: string): void;
-}
-
-declare var History: {
-    prototype: History;
```
mhegazy: there is a var `History` i find in most browsers. why was this removed?

falsandtru: See https://github.com/Microsoft/TSJS-lib-generator/pull/142#issuecomment-242893870

mhegazy: The `History` is the prototype for the type. the constructor can not be called directly, but that is the same as the rest of the HTML*Element classes as well. so we should keep both a type and a value for `History`. we already have a value `history: History` on `Window` and the global object.

falsandtru: Okay, fixed.

baselines/dom.generated.d.ts
```diff
@@ -7536,11 +7536,12 @@ declare var HashChangeEvent: {
...
-    back(distance?: any): void;
-    forward(distance?: any): void;
-    go(delta?: any): void;
-    pushState(statedata: any, title?: string, url?: string): void;
-    replaceState(statedata: any, title?: string, url?: string): void;
+    scrollRestoration: ScrollRestoration;
+    back(): void;
+    forward(): void;
+    go(delta?: number): void;
+    pushState(data: any, title: string, url?: string | null): void;
```
zhengbli: optional parameter doesn't need the `| null` annotation, it gets that automatically.
Also, if missing the value of the optional parameter should be `undefined` I think

baselines/dom.generated.d.ts
```diff
@@ -7536,11 +7536,12 @@ declare var HashChangeEvent: {
 interface History {
     readonly length: number;
     readonly state: any;
-    back(distance?: any): void;
-    forward(distance?: any): void;
-    go(delta?: any): void;
-    pushState(statedata: any, title?: string, url?: string): void;
-    replaceState(statedata: any, title?: string, url?: string): void;
+    scrollRestoration: ScrollRestoration;
```
zhengbli: is it a read only property? If so add the modifier

falsandtru: See https://html.spec.whatwg.org/multipage/browsers.html#the-history-interface

baselines/dom.generated.d.ts
falsandtru: Its signature in spec is `void pushState(any data, DOMString title, optional USVString? url = null);`. So it needs to accept a `null` value and should be optional.

https://html.spec.whatwg.org/multipage/browsers.html#the-history-interface

zhengbli: Ah, makes sense. Thanks

zhengbli:
👍 Thanks!

PR #143, opened 2016-08-26
Avoid using reflection in hot spots

zhengbli:
This changes brings down the running time from > 2 mins to 1.4 sec on my laptop ...
Lessons learned, reflection is expensive.

### DIFF
--- a/Shared.fsx
+++ b/Shared.fsx
@@ -70,7 +70,18 @@ module JsonItems =
         | SignatureOverload
         | TypeDef
         | Extends
-        override x.ToString() = (unionToString x).ToLower()
+        override x.ToString() = 
+            match x with
+            | Property _ -> "property"
+            | Method _ -> "method"
+            | Constant _ -> "constant"
+            | Constructor _ -> "constructor"
+            | Interface _ -> "interface"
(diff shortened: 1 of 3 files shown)


zhengbli:
It doesn't seem to be very easy to get the caching right, tried static member and didn't work well, and F# honors declaration order so it is hard to cache outside the type. I think it might be good enough to merge for now, may be improved later.

DanielRosenwasser:
> This changes brings down the running time from > 2 mins to 1.4 sec on my laptop 

Nice!

vladima:
Generic caching can be implemented as something like this:
```f#
type UnionCaseNameCache<'T> private ()= 
    static let reader = 
        FSharp.Reflection.FSharpValue.PreComputeUnionTagReader typeof<'T>
    static let cases =
        FSharp.Reflection.FSharpType.GetUnionCases typeof<'T>
        |> Seq.map(fun c -> c.Tag, c.Name)
        |> dict
    static member Get (c: 'T) = cases.[reader c]

let unionToString c = UnionCaseNameCache.Get c
```

but I don't think it will beat the hand-rolled specialization

PR #146, opened 2016-09-09
Remove number indexer declaration on window object.

ChogyDan:
This allows string indexing to be preferred which is the typical expectation for the window object.  Fixes TypeScript issue #10457


mhegazy:
Thanks!

mhegazy:
Fixes https://github.com/Microsoft/TypeScript/issues/10457

PR #148, opened 2016-09-23
change DataTransfer.types property from DOMStringList to string[]

YuichiNukiyama:
fix [#10979](https://github.com/Microsoft/TypeScript/issues/10979)


YuichiNukiyama:
I see. I change all DOMStringList to string[].

Artazor:
@mhegazy please check! DataTransfer.types is still `string[]` - something nasty happened on merge. Commit 09dfad2 has disappeared both from release-2.1 and master (thus in npm we have a TS with bug)

PR #151, opened 2016-10-04
Add Shadow DOM v1 APIs

rictic:
Based on https://www.w3.org/TR/2016/WD-shadow-dom-20160830/

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -76,6 +76,7 @@ interface DoubleRange {
 }
 
 interface EventInit {
+    scoped?: boolean;
     bubbles?: boolean;
     cancelable?: boolean;
 }
@@ -2291,7 +2292,7 @@ declare var DeviceRotationRate: {
     new(): DeviceRotationRate;
 }
 
-interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEvent, ParentNode {
+interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEvent, ParentNode, DocumentOrShadowRoot {
     /**
       * Sets or gets the URL for the current document. 
       */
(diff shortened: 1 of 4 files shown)


baselines/dom.generated.d.ts
```diff
@@ -3465,6 +3466,9 @@ interface Element extends Node, GlobalEventHandlers, ElementTraversal, NodeSelec
     readonly scrollWidth: number;
     readonly tagName: string;
     innerHTML: string;
+    readonly assignedSlot: HTMLSlotElement | null;
```
zhengbli: The property is optional on the spec, we should make it optional here so that `undefined` can be accepted as well.

zhengbli: And the same for other places too.

rictic: I initially wrote it like that, but I when I was doing some double-checking it looked like WebIDL's `?` suffix is different from typescript's `?`. Take a look at this and let me know what you think:
  * For turning an IDL specified type `T?` into an ES value, this sections seems like it's saying that only values from `T` and `null` are allowed: https://heycam.github.io/webidl/#idl-nullable-type
  * For turning an ES value into the IDL type `T?` then `undefined` and `null` are treated the same: https://heycam.github.io/webidl/#es-nullable-type

So for places where the DOM accepts an IDL type `T?` from ES code then `T | null | undefined` looks like the right TS type, but for readonly properties it seems like `T | null` is more correct.

Apologies if this debate has already been litigated and the decision to treat nullable values as also allowing undefined has been made across the board.

I have made the non-required members of dictionaries, like `EventInit#scoped` optional.

rictic:
I noticed a copy-paste mistake in my earlier patch. I accidentally spelled `HTMLSlotElement` as `ShadowRootInit`. Pushed up a fixup commit

zhengbli:
Thank you @rictic for contributing!

baselines/dom.generated.d.ts
mhegazy: I do not see why a readonly propoerty can not be undefined. this is useful for users who want to feature dection. this should be:

```ts
readonly assignedSlot?: HTMLSlotElement | null;
```

mhegazy:
the new properties should be marked optional.

baselines/dom.generated.d.ts
rictic: As I read the spec, in a browser that implements this API, `assignedSlot` will never be `undefined`.

It's true that in an environment without the API it could be `undefined`, but this is true of most browser APIs.

PR #165, opened 2016-11-08
evaluate should allow null for resolver and result

goenning:
This should fix https://github.com/Microsoft/TypeScript/issues/11981 

I've also changed the return type of `XPathExpression.evaluate` to `XPathResult` based on https://developer.mozilla.org/en-US/docs/Web/API/XPathExpression


zhengbli:
👍 Thanks @goenning !

PR #166, opened 2016-11-18
Reduce symbol instantiations in lib.d.ts

mhegazy:
A few changes:
- avoid using `this` types and instead add the overloads to every derived interface
- use Mapped types instead of listing event names/tag names explicitly in overloads

//cc: @zhengbli 

Build diagnostics building `dom.genenrated.d.ts` before the change:
```sh
Files:            2
Lines:        18814
Nodes:       102909
Identifiers:  35546
Symbols:      95889
Types:        12369
Memory used: 94151K
I/O read:     0.00s
I/O write:    0.00s
Parse time:   0.36s
Bind time:    0.18s
Check time:   1.13s
Emit time:    0.02s
Total time:   1.68s
```

After the change:

```sh
Files:            2
Lines:        17559
Nodes:        69571
Identifiers:  23524
Symbols:      18183
Types:         4095
Memory used: 52984K
I/O read:     0.01s
I/O write:    0.00s
Parse time:   0.32s
Bind time:    0.15s
Check time:   0.50s
Emit time:    0.02s
Total time:   0.98s
```


rictic:
This is so cool! It looks like this will be great news for custom elements. e.g. experimenting with the nightly build and some of the example code in https://github.com/Microsoft/TypeScript/issues/12311  it seems like this typechecks:

```typescript
class MySlider extends HTMLElement {
  slide() {}
}
customElements.define('my-slider', MySlider);
interface HTMLKind {
  'my-slider': MySlider;
}

document.createElement('my-slider').slide()
```

So awesome!

Shared.fsx
```diff
@@ -519,25 +519,31 @@ let iNameToEhList =
...
-            | None -> []
+            | Some i when hasHandler i -> [i]
+            | _ -> []
 
         let implementedEventHandler =
             let implementis = i.Implements |> Array.map GetInterfaceByName
             [ for i' in implementis do
                   yield! match i' with
-                         | Some i -> GetEventHandler i
+                         | Some i ->  if hasHandler i then [i] else []
```
zhengbli: So this returns the interface itself instead of the handlers it has?

TS.fsx
```diff
@@ -117,32 +117,55 @@ let matchSingleParamMethodSignature (m: Browser.Method) expectedMName expectedMT
...
+    for e in tagNameToEleName do
+        if iNameToIDependList.ContainsKey e.Value && Seq.contains "HTMLElement" iNameToIDependList.[e.Value] then
+            Pt.printl "\"%s\": %s;" (e.Key.ToLower()) e.Value
+    Pt.decreaseIndent()
+    Pt.printl "}"
+    Pt.printl ""
+
+let EmitElementTagNameMap () =
+    Pt.printl "interface ElementTagNameMap {"
+    Pt.increaseIndent()
```
zhengbli: It looks like `ElementTagNameMap` interface is always the same as `HTMLElementTagNameMap`? Then make is an alias instead

TS.fsx
```diff
@@ -117,32 +117,55 @@ let matchSingleParamMethodSignature (m: Browser.Method) expectedMName expectedMT
...
+        Pt.printl "\"%s\": %s;" (e.Key.ToLower()) e.Value
+    Pt.decreaseIndent()
+    Pt.printl "}"
+    Pt.printl ""
+
+let EmitElementListTagNameMap () =
+    Pt.printl "interface ElementListTagNameMap {"
+    Pt.increaseIndent()
+    for e in tagNameToEleName do
+        Pt.printl "\"%s\": NodeListOf<%s>;" (e.Key.ToLower()) e.Value
```
zhengbli: We can do 
```ts
type ElementListTagNameMap = { [K in ElementTagNameMap]: ElementTagNameMap[K] }
```
and avoid the loop

mhegazy: That is what i had earlier, then changed it. the issue is that NodeListOf has a constraint that `T extends Node`. we can not verify this constraint on `ElementTagNameMap[K]` unless i add a string indexer. adding the indexer makes the language service experience of the signature help much worse. so this is a work around for that.

TS.fsx
```diff
@@ -516,7 +532,28 @@ let EmitIndexers emitScope (i: Browser.Interface) =
...
+    let ownEventHandles = if iNameToEhList.ContainsKey i.Name && not iNameToEhList.[i.Name].IsEmpty then iNameToEhList.[i.Name] else []
+    if ownEventHandles.Length > 0 then
+        Pt.printl "interface %sEventMap" i.Name
+        if iNameToEhParents.ContainsKey i.Name && not iNameToEhParents.[i.Name].IsEmpty then
+            let extends = iNameToEhParents.[i.Name] |> List.map (fun i -> i.Name + "EventMap")
+            Pt.print " extends %s" (String.Join(", ", extends))
+        Pt.print " {"
+        Pt.increaseIndent()
+        ownEventHandles |> List.iter EmitInterfaceEventMapEntry
+        // Pt.printl "[x: string]: Event;"
```
zhengbli: remove this

Shared.fsx
mhegazy: yes. i could make it the interface name. but seemed the same

TS.fsx
mhegazy: no they are different. not sure why though. i wanted to ask you about that.

TS.fsx
mhegazy: done.

Shared.fsx
zhengbli: Then the name of the variable should be changed to `implementedInterface` and other places, otherwise it can be confusing later

TS.fsx
zhengbli: Aha now I remembder, the `EmitHTMLElementTagNameMap` only prints tag elements that either extends or implements the `HTMLElement` type.

zhengbli:
The `addEventListener` in the `addedTypes.json` file need to be changed as well, and other than that looks good to me. Thanks!

PR #167, opened 2016-11-29
Add URLSearchParams

### LINKED ISSUE #153: Support UrlSearchParams
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9378

To quote @ryan-codingintrigue:

> ...global definitions like [UrlSearchParams] should be in TypeScript's lib.d.ts definition
> (See https://github.com/Microsoft/TypeScript/blob/master/src/lib/dom.generated.d.ts#L11759)
> 
> But they only go into that once Microsoft Edge supports the feature and at present, URLSearchParams is not present in Edge.

nicolashenry:
Fixes #153

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -10814,6 +10814,7 @@ interface URL {
     protocol: string;
     search: string;
     username: string;
+    readonly searchparams: URLSearchParams;
     toString(): string;
 }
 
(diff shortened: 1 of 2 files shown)


mhegazy:
Can you include the comments from https://github.com/Microsoft/TypeScript/issues/12517#issuecomment-263445670.

mhegazy:
also we will need to add the iterable definition, like `entries`, `keys` and `[Symbol.Iterator]` in https://github.com/Microsoft/TypeScript/blob/master/src/lib/dom.iterable.d.ts, but this is a separate PR.

nicolashenry:
I tried to add them in comments.json but it doesn't work. I added them in signatures instead but it's maybe not the right way to do it ?

mhegazy:
> I tried to add them in comments.json but it doesn't work.

@zhengbli why dose not this work?

>I added them in signatures instead but it's maybe not the right way to do it ?

Well.. my first reaction was that not right.. but seems to work :)

zhengbli:
There is a bug with comment.json that it doesn't search for the newly added types. But I don't think we should put comments in the json files. Can you split this PR into two, one for the code change (should be ready to merge) and the other for the comments? I'll try to fix it today. Sorry for the hassle.

mhegazy:
sorry for my original comment about adding comments. i thought it would work, and figured it would be better for the user. sorry for the hassel.

zhengbli:
👍 Thanks @nicolashenry !

PR #168, opened 2016-11-30
Add comments to URLSearchParams

nicolashenry:
See https://github.com/Microsoft/TSJS-lib-generator/pull/167#issuecomment-264004981


zhengbli:
This PR is no longer needed in light of #182

PR #170, opened 2016-12-04
fix querySelector method

YuichiNukiyama:
fixes [#12568](https://github.com/Microsoft/TypeScript/issues/12568)


YuichiNukiyama:
done.

baselines/dom.generated.d.ts
```diff
@@ -13736,8 +13736,10 @@ interface NavigatorUserMedia {
 
 interface NodeSelector {
     querySelector<K extends keyof ElementTagNameMap>(selectors: K): ElementTagNameMap[K] | null;
+    querySelector<E extends Element>(selectors: string): E | null;
```
mhegazy: why not combine both signatures as:
```ts
querySelector<E extends Element = Element>(selectors: string): E | null;
```

YuichiNukiyama: I did not know generic default... I fix it 😃

AustinMatherne:
@mhegazy Is there a holdup on getting this merged?

baselines/dom.generated.d.ts
```diff
@@ -13736,8 +13736,9 @@ interface NavigatorUserMedia {
 
 interface NodeSelector {
     querySelector<K extends keyof ElementTagNameMap>(selectors: K): ElementTagNameMap[K] | null;
-    querySelector(selectors: string): Element | null;
+    querySelector<E extends Element = Element>(selectors: string): E | null;
     querySelectorAll<K extends keyof ElementListTagNameMap>(selectors: K): ElementListTagNameMap[K];
+    querySelectorAll<E extends Element>(selectors: string): NodeListOf<E>;
```
AustinMatherne: Can we combine these, too?
```
querySelectorAll<E extends Element = Element>(selectors: string): NodeListOf<E>;
```

YuichiNukiyama: Good point!

AustinMatherne:
@mhegazy anything holding this PR up? This makes testing the DOM much nicer in Angular 2.

PR #173, opened 2016-12-04
Add "| null" to ElementTraversal properties

perryjiang:
From issue (https://github.com/Microsoft/TypeScript/issues/12526).


perryjiang:
I'm confused as to what's causing one of the checks to fail. All the tests passed locally and I only changed files mentioned in the contribution guidelines.

mhegazy:
The error is from running the tests in the ts repo. you do not have to worry about it, we will update it when we port this into the main repro.

PR #176, opened 2016-12-08
Add Custom Elements v1 typings

vvakame:
related https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13045

https://w3c.github.io/webcomponents/spec/custom/
https://dom.spec.whatwg.org/#dom-document-createelement

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3094,8 +3094,8 @@ interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEven
      * Creates an instance of the element for the specified tag.
      * @param tagName The name of an element.
      */
-    createElement<K extends keyof HTMLElementTagNameMap>(tagName: K): HTMLElementTagNameMap[K];
-    createElement(tagName: string): HTMLElement;
+    createElement<K extends keyof HTMLElementTagNameMap>(tagName: K, options?: ElementCreationOptions): HTMLElementTagNameMap[K];
+    createElement(tagName: string, options?: ElementCreationOptions): HTMLElement;
     createElementNS(namespaceURI: "http://www.w3.org/1999/xhtml", qualifiedName: string): HTMLElement;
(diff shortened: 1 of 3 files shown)


baselines/dom.generated.d.ts
```diff
@@ -13009,10 +13009,14 @@ interface ElementDefinitionOptions {
     extends: string;
 }
 
+interface ElementCreationOptions {
+    is: string;
```
rictic: It looks like `is` is optional as I read the spec

vvakame: https://dom.spec.whatwg.org/#dictdef-elementcreationoptions
I think `is` is not optional. 
I want to check document that you seen.

> `[NewObject] Element createElement(DOMString localName, optional ElementCreationOptions options);`

options is optional.

vvakame:
oh, I read README.md again. new guideline added.
https://github.com/Microsoft/TSJS-lib-generator#when-should-a-dom-api-be-included-here

This PR contains in living standard already.
https://dom.spec.whatwg.org/#dom-document-createelement

baselines/dom.generated.d.ts
rictic: Fair point. I was going off of the listed algorithm, which handles the case where `is` isn't provided. Looking at chrome's implementation confirms your reading, it doesn't crash, but it does `setAttribute('is', '[object Object]')`.

In short, I withdraw the suggestion, `is` should be required.

vvakame:
👀

baselines/dom.generated.d.ts
SaschaNaz: Months later, but `is` is optional. Web IDL spec says a dictionary member is by default optional unless it has `required` keyword.

vvakame: Ah, you are right. I'll fix this PR.
https://heycam.github.io/webidl/#prod-DictionaryMember

vvakame:
I did rebase to master branch & address comment.
@mhegazy could you review this PR?

saschanaz:
This looks simple and short enough to review... Hmm.

PR #179, opened 2016-12-29
Add guidelines for when a DOM API belongs here vs on DefinitelyTyped

andy-ms:
Not sure what exact standards should be used. Hopefully someone more familiar with web standards can comment.
This will affect whether we choose #176 or DefinitelyTyped/DefinitelyTyped#13045.


andy-ms:
@zhengbli Do you know why tests are failing?

zhengbli:
The appveyor tests basically clones the TypeScript code, replace the dom lib files and run the TypeScript test suite. Looks like the "verifyDefaultLib_dom.errors.txt" baseline has changed so the tests failed. You would need to run the TS tests locally to make sure all baseline changes are expected, then check in here and update the baselines seperately in the TS repo.

zhengbli:
Fixing in #180

PR #180, opened 2016-12-29
Fix build error

zhengbli:
Build was broken because `Element` inherits from both `ElementTraversal` and `ParentNode`, while a recent PR made two properties from `ElementTraversal` nullable, so TS complains these properties have different types in different base types. 

Related spec:
https://dom.spec.whatwg.org/#dom-parentnode-firstelementchild
The `firstElementChild` and `lastElementChild` are indeed nuallble as well in `ParentNode` interface.


mhegazy:
thanks!

PR #181, opened 2017-01-03
Fix https://github.com/Microsoft/TypeScript/issues/13269: Make `DataTransfer.types` a `DOMString[]`

mhegazy:
According to the spec: https://html.spec.whatwg.org/multipage/interaction.html#datatransfer
`types` is `DOMString[]` and not `string[]`.

Fixes https://github.com/Microsoft/TypeScript/issues/13269
//cc: @falsandtru


mhegazy:
I got confused, `DOMString` is string, `DOMStringList` is not `DOMString[]`. the current behavior is correct.

PR #182, opened 2017-01-03
Remove stale JS-related code + Refactor + Improve comment.json support for added types

zhengbli:
The JS-related things used to be useful for the JS language service in VS, but now it is not used anymore. This PR cleans up the code:
 - removed JS-related stuff
 - put all code in a single file `TS.fsx`
 - break things into smaller modules, making it more readable
 - code refactor in many places to improve speed

and also incorporated https://github.com/Microsoft/TSJS-lib-generator/pull/168 for the comment.json support for added types.


zhengbli:
@mhegazy as this is a quite big change, it can go stale pretty fast. Please take a look if you have time, thanks!

PR #183, opened 2017-01-05
Replace ElementListTagNameMap 

FranklinWhale:
For Microsoft/TypeScript#13297


TS.fsx
```diff
@@ -784,6 +784,7 @@ module Emit =
         for e in tagNameToEleName do
             if iNameToIDependList.ContainsKey e.Value && Seq.contains "HTMLElement" iNameToIDependList.[e.Value] then
                 Pt.Printl "\"%s\": %s;" (e.Key.ToLower()) e.Value
+        Pt.Printl "[index: string]: HTMLElement;"
```
mhegazy: i did not add these originally to make the signature help more informative. with this change you will see something like:
`getElementsByTagName<K>(tagname: K)...` instead of `getElementsByTagName<"a" | "div" | ...>(tagname: "a" | "div" | ..)...`

FranklinWhale: I see. To fix that, instead of placing the index signature directly inside `ElementTagNameMap`, I have created another interface (known as `IndexedElementTagNameMap`) that extends `ElementTagNameMap` and moved the index signature there.

The change is now in commit f8c3f19e3143d045ba077ced23cda14b5ba37924.

FranklinWhale:
@mhegazy I would like to also add the following `getElementsByTagNameNS` overload:

```ts
getElementsByTagNameNS<K extends keyof HTMLElementTagNameMap>(namespaceURI: "http://www.w3.org/1999/xhtml", localName: K): HTMLCollectionOf<IndexedHTMLElementTagNameMap[K]>;
```

However, I am not sure whether the overload above should be added to TS.fsx or overridingTypes.json. `IndexedHTMLElementTagNameMap` is generated in TS.fsx along with the overloads of `getElementsByTagName`, but the overloads of `getElementsByTagNameNS` are located in overridingTypes.json.

mhegazy:
looks like the file has errors now, tests/cases/compiler/modularizeLibrary_Dom.iterable.ts is not passing.

mhegazy:
> However, I am not sure whether the overload above should be added to TS.fsx or overridingTypes.json. IndexedHTMLElementTagNameMap is generated in TS.fsx along with the overloads of getElementsByTagName, but the overloads of getElementsByTagNameNS are located in overridingTypes.json.

we handle these all in the fsx file. maybe now we do not need to do this, so i would keep it there for consistency.

FranklinWhale:
> maybe now we do not need to do this, so i would keep it there for consistency.

I am not sure which file "there" refers to. I have added method signature to overridingTypes.json. Is that correct?

> looks like the file has errors now, tests/cases/compiler/modularizeLibrary_Dom.iterable.ts is not passing.

Ya, I noticed that earlier (https://github.com/Microsoft/TypeScript/issues/13297#issuecomment-270755323). Seem that tests/baselines/reference/modularizeLibrary_Dom.iterable.types should be updated to change `NodeListOf` to `HTMLCollectionOf`. Should I also submit another pull request to update the test case?

FranklinWhale:
@mhegazy I have created a PR (Microsoft/TypeScript#13516) to update the test case.

FranklinWhale:
@mhegazy Ping. Any update on this PR?

DaSchTour:
@FranklinWhale I guess the conflicts needs to be resolved to get this merged.
Or is there anything else holding up this?

mhegazy:
sorry for the delay. @FranklinWhale can you refresh the PR

PR #186, opened 2017-01-10
Update libs using new drop of Spec file

zhengbli:
Other related changes:

1. Added some non-trivial logic to deal with the "extend conflict":
    `interface B extends A`, however `B.member` is not assignable to `A.member`. This happened twice in this update for `AudioContext.suspend` and `OfflineAudioContext.suspend`; and `HTMLCollection.namedItem` and `HTMLFormControlsCollection.namedItem`

2. Updated the `known worker interface` list. Now the list contains every type used by worker types so that the worker type file can be used independently.

3. The event types in event handler now prefer the event type defined in the current interface's own event collection; if none is found then fall back to the global event map


zhengbli:
The test fails because a new "HTMLOutputElement" is added and the `getElementByTagName` 's parameter's union type changed accordingly.

PR #193, opened 2017-02-13
Use Promise<T> instead of PromiseLike<T>

mhegazy:
Fix for https://github.com/Microsoft/TypeScript/issues/13947

Blocked on https://github.com/Microsoft/TypeScript/pull/14053


zhengbli:
👍

PR #194, opened 2017-02-14
Fix document.implementation.hasFeature signature

argelius:
https://github.com/Microsoft/TypeScript/issues/14062

The signature for this function is incorrect:
https://developer.mozilla.org/ja/docs/Web/API/DOMImplementation/hasFeature

### DIFF
--- a/inputfiles/browser.webidl.xml
+++ b/inputfiles/browser.webidl.xml
@@ -3241,10 +3241,7 @@
         <method name="createHTMLDocument" type="Document">
           <param name="title" type="DOMString" />
         </method>
-        <method name="hasFeature" type="boolean">
-          <param name="feature" nullable="1" type="DOMString" />
-          <param name="version" nullable="1" type="DOMString" />
-        </method>
+        <method name="hasFeature" type="boolean" />
       </methods>
     </interface>
     <interface name="DOMParser" extends="Object">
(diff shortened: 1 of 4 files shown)


inputfiles/browser.webidl.xml
```diff
@@ -3241,7 +3241,10 @@
         <method name="createHTMLDocument" type="Document">
           <param name="title" type="DOMString" />
         </method>
-        <method name="hasFeature" type="boolean" />
+        <method name="hasFeature" type="boolean">
```
mhegazy: Please do not edit these files. these are imported and used as inputs. you need to edit https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/overridingTypes.json instead

argelius: @mhegazy Thanks! Sorry about that, should have read the README more closely!

I have reverted the changes to the xml files and added an override.

argelius:
@mhegazy Thanks for the review. The issues should be fixed now.

mhegazy:
@argelius, can you please sign the CLA?

argelius:
@mhegazy When I signed it yesterday the bot didn't update the labels so I closed and opened the PR myself and it added the `cla-signed` label but the `cla-required` label was not removed.

I can try signing it again.

mhegazy:
thanks. seems like i did not see the second label. sorry about that.

PR #137, opened 2016-07-24
Add ParentNode interface

YuichiNukiyama:
fix [#4400](https://github.com/Microsoft/TypeScript/issues/4400), [#7528](https://github.com/Microsoft/TypeScript/issues/7528)

PR #199, opened 2017-02-22
Fix DocumentFragment interface

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/14223

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3448,7 +3448,7 @@ declare var Document: {
     new(): Document;
 }
 
-interface DocumentFragment extends Node, NodeSelector, ParentNode {
+interface DocumentFragment extends Node, NodeSelector {
 }
 
 declare var DocumentFragment: {
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -1214,11 +1214,6 @@
         "interface": "Document"
     },
     {
-        "kind": "extends",
-        "baseInterface": "ParentNode",
-        "interface": "DocumentFragment"
-    },
-    {
         "kind": "typedef",
         "name": "ScrollRestoration",
         "flavor": "DOM",


falsandtru:
ping @zhengbli @DanielRosenwasser @mhegazy

baselines/dom.generated.d.ts
```diff
@@ -3448,7 +3448,11 @@ declare var Document: {
     new(): Document;
 }
 
-interface DocumentFragment extends Node, NodeSelector, ParentNode {
+interface DocumentFragment extends Node {
+    querySelector<K extends keyof ElementTagNameMap>(selectors: K): ElementTagNameMap[K] | null;
```
mhegazy: so why not extend from NodeSelector?

falsandtru: As I specified in https://github.com/Microsoft/TypeScript/issues/14246, NodeSelector's properties are defined with ParentNode, not NodeSelector. So we must merge and remove NodeSelector by this pr and #201.

mhegazy: I understand. but why remove `NodeSelector` from the extends clause, and then define the functions inline? 

it would have been sufficient to just remove `ParentNode` from the extends clause.

falsandtru: ok, I updated my PRs.

mhegazy:
Looking at the histry again, this change has been added in https://github.com/Microsoft/TSJS-lib-generator/pull/137 by @YuichiNukiyama  to address fix [#4400](https://github.com/Microsoft/TypeScript/issues/4400), [#7528](https://github.com/Microsoft/TypeScript/issues/7528). 

Reverting this change will be a breaking change. the ParentNode properties seem to be defined in Chrome,  and Edge and referenced in the [MDN  documentation](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment).

This does not seem to be a worth while change, at least with the Selector API is still a working draft.

falsandtru:
This change doesn't affect selector apis.

falsandtru:
> the ParentNode properties seem to be defined in Chrome, and Edge and referenced in the MDN documentation.

As I specified in https://github.com/Microsoft/TypeScript/issues/14223, ParentNode properties are NOT defined (implemented) in Edge. Of course Element should inherit ParentNode for its actual apis, but DocumentFragment is NOT. These interfaces have different state of implemetations about ParentNode. Do you really believe ParentNode apis on DocumentFragment?

mhegazy:
I am not an expert on web APIs and browser compt. What i care about is not introducing unneeded breaks for users of the standard library. The changes in  #137 explicitly added `ParentNode` to the heritage clause of `DocumentFragment`, and these APIs seem supported in Firefox:
image

And chrome:
image

so unless I missed something, i am inclined to punt on this issue, and wait until the spec reaches the next stage before making any changes.

falsandtru:
So do you accept the error in Edge (to support the unclear spec)? What is your mind? I don't know about that.

mhegazy:
> So do you accept the error in Edge (to support the unclear spec)? What is your mind? I don't know about that.

From a user perspective, it is easier to have the API and not use it, than to use it and get a compiler error. IE for example does not support many of the APIs in lib.d.ts today.

falsandtru:
I agree about that in general terms. But MDN says ParentNode properties are not experimenal in Element, experimetal in DocumentFragment.

https://developer.mozilla.org/en-US/docs/Web/API/Element
https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment

So I feel uneasy about the correctness of ParentNode apis in DocumentFragment. Finally, can you believe that correctness?

mhegazy:
> So I feel uneasy about the correctness of ParentNode apis in DocumentFragment. Finally, can you believe that correctness?

I am not really an expert on web APIs, so i am not the best person to comment about the validity of these declarations. it seems that browsers do support them today; and i worry changing them would break valid code for no to little value.

falsandtru:
ok

PR #201, opened 2017-02-23
Fix ParentNode interface

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/14246


falsandtru:
@zhengbli Can you add the feature to remove the interfaces?

PR #207, opened 2017-03-01
Fix Event.currentTarget type

zhengbli:
This should fix https://github.com/Microsoft/TypeScript/issues/299


zhengbli:
Travis seems to have problems downloading mono bits. Will retry later.

baselines/webworker.generated.d.ts
```diff
@@ -399,9 +399,9 @@ declare var Event: {
 }
 
 interface EventTarget {
-    addEventListener(type: string, listener?: EventListenerOrEventListenerObject, useCapture?: boolean): void;
+    addEventListener(type: string, listener?: EventListenerOrEventListenerObject<this>, useCapture?: boolean): void;
```
mhegazy: in general we do not want to set `this` as it turns all the types that inherit from EventTarget (almost all DOM types) as generic, and every time you reference them we make an instantiation, more memory and more time spent checking the default lib file.

can we make EventTarget<T extends EventTarget> and pass the current type through?

zhengbli:
@mhegazy now there are no `<this>` as type parameters any more. Further comments?

DaSchTour:
Why is this still pending? What is the next step? How could I maybe help?

SaschaNaz:
This PR still has performance problem:

```
Files:             2
Lines:         19211
Nodes:         85546
Identifiers:   30004
Symbols:       96920
Types:         25399
Memory used: 101439K
I/O read:      0.00s
I/O write:     0.00s
Parse time:    0.36s
Bind time:     0.17s
Check time:    1.12s
Emit time:     0.02s
Total time:    1.67s
```

Compared to the current master:

```
Files:            2
Lines:        19316
Nodes:        84450
Identifiers:  29164
Symbols:      22004
Types:         5790
Memory used: 56757K
I/O read:     0.03s
I/O write:    0.00s
Parse time:   0.38s
Bind time:    0.16s
Check time:   0.64s
Emit time:    0.02s
Total time:   1.20s
```

DaSchTour:
Any progress on this? It's really annoying to have as `as HTMLElement` all over my code where I use EventListeners. That really messes up my code and every second week somebody asks me why "Events are broken" in TypeScript or I discover code where Events are "casted" to `any` just to get around this wrong type definition.

DaSchTour:
Why is this closed? This is still an issue! And explanation on why this is closed would be very helpful.

saschanaz:
I guess because the author stopped working on TS.

sandersn:
The PR is two years out of date, back when (1) the code was written in F# (2) the IDL source was IE or Edge.

Microsoft/TypeScript#299, the issue tracking this problem, is still open and has 31 upvotes right now.

DaSchTour:
Okay. Interesting news. I wonder why this isn't communicated more offensively. I guess many people would like to know more about such changes. It would be nice to see such changes and information in TypeScript Release Notes and Roadmap.

PR #219, opened 2017-03-26
Add missing methods for FormData type

NColey:
Fixes https://github.com/Microsoft/TypeScript/issues/14813 by updating the existing append method to better match the [Web APIs](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData) and adding missing methods. Please let me know if there is anything I should add or change, thanks!

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3796,12 +3796,12 @@ declare var FocusNavigationEvent: {
 }
 
 interface FormData {
-    append(name: string, value: string | blob, fileName?: string): void;
+    append(name: string, value: string | Blob, fileName?: string): void;
     delete(name: string): void;
     get(name: string): FormDataEntryValue | null;
     getAll(name: string): FormDataEntryValue[];
     has(name: string): boolean;
-    set(name: string, value: string | blob, fileName?: string): void;
+    set(name: string, value: string | Blob, fileName?: string): void;
 }
 
 declare var FormData: {
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3796,12 +3796,12 @@ declare var FocusNavigationEvent: {
 }
 
 interface FormData {
-    append(name: any, value: any, fileName?: string): void;
-    delete(name: any): void;
-    get(name: any): any;
-    getAll(name: any): any;
-    has(name: any): any;
-    set(name: any, value: any, fileName?: string): void;
+    append(name: string, value: any, fileName?: string): void;
+    delete(name: string): void;
+    get(name: string): any;
+    getAll(name: string): any[];
+    has(name: string): boolean;
+    set(name: string, value: any, fileName?: string): void;
 }
 
 declare var FormData: {
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3796,7 +3796,12 @@ declare var FocusNavigationEvent: {
 }
 
 interface FormData {
-    append(name: any, value: any, blobName?: string): void;
+    append(name: any, value: any, fileName?: string): void;
+    delete(name: any): void;
+    get(name: any): any;
+    getAll(name: any): any;
+    has(name: any): any;
+    set(name: any, value: any, fileName?: string): void;
 }
 
 declare var FormData: {
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3796,12 +3796,12 @@ declare var FocusNavigationEvent: {
 }
 
 interface FormData {
-    append(name: string, value: any, fileName?: string): void;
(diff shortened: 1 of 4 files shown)


baselines/dom.generated.d.ts
```diff
@@ -3796,7 +3796,12 @@ declare var FocusNavigationEvent: {
...
 
 interface FormData {
-    append(name: any, value: any, blobName?: string): void;
+    append(name: any, value: any, fileName?: string): void;
+    delete(name: any): void;
+    get(name: any): any;
+    getAll(name: any): any;
+    has(name: any): any;
+    set(name: any, value: any, fileName?: string): void;
 }
```
mhegazy: From the spec: https://xhr.spec.whatwg.org/#formdata the definition should look like:
```ts
type FormDataEntryValue = File or string;

interface FormData {
  append(name: string, value: string): void;
  append(name:string, blobValue:Blob, filename?:string):void;
  delete(name: string):void;
  get(name:string):  FormDataEntryValue | null;
  getAll(name: string): FormDataEntryValue[];
  has(name: string): boolean;
  set(name:string, value:string): void;
  set(name: string, blobValue:Blob, filename?: string): void;
};
```

Kovensky: You can add arbitrary Blob, but you can only get File back? interesting

NColey:
My apologies @mhegazy , I had been referencing the definitions here: https://developer.mozilla.org/en-US/docs/Web/API/FormData. I just updated the PR to match the appropriate definitions from the spec, thank you for sharing the link with me!

baselines/dom.generated.d.ts
```diff
@@ -3796,7 +3796,12 @@ declare var FocusNavigationEvent: {
...
 
 interface FormData {
-    append(name: any, value: any, blobName?: string): void;
+    append(name: string, value: any, fileName?: string): void;
```
mhegazy: Still have a few `any`s left:

```ts
type FormDataEntryValue = File | string;

interface FormData {
  append(name: string, value: string): void;
  append(name:string, blobValue:Blob, filename?:string):void;
  delete(name: string):void;
  get(name:string):  FormDataEntryValue | null;
  getAll(name: string): FormDataEntryValue[];
  has(name: string): boolean;
  set(name:string, value:string): void;
  set(name: string, blobValue:Blob, filename?: string): void;
}
````

Comment (NColey) [inline baselines/dom.generated.d.ts]: Ah sorry, I have a quick question before I update my PR, if I can't find a FormDataEntryValue interface in the generated file does that make it ok to add it as a type anyway? Or should that be added first?

Comment (Kovensky) [inline baselines/dom.generated.d.ts]: That is correct, it's just a `type` so it should not be in the generated code, but it should be available to users to use as a type.

Comment (NColey) [inline baselines/dom.generated.d.ts]: Thanks so much, that makes sense! I just updated the PR with the appropriate types added in.

Comment (Kovensky) [inline baselines/dom.generated.d.ts]: Ah, looking in the full diff I understand better what you meant.

It seems you need to add FormDataEntryValue to the json file as a typedef (search "typedef" in it to find examples)

**Comment by NColey** [comment]
@Kovensky Is there another area beyond just as a typedef where it might
need to be added? I just updated my PR with the typedef for
FormDataEntryValue but the test is failing with the following error `Error:
The baseline file verifyDefaultLIb_dom.errors.txt has changed.` which is
both a file that I don't seem to have and also one I suspect I shouldn't
touch. I can't seem to figure out what in my changes would have prompted
that but I'm guessing there is another area of the code where I would need
to add FormDataEntryValue?

File: baselines/dom.generated.d.ts (hunk 2 above)
Comment (NColey) [inline baselines/dom.generated.d.ts]: @Kovensky Is there another area beyond just as a typedef where it might need to be added? I just updated my PR with the typedef for FormDataEntryValue but the test is failing with the following error `Error: The baseline file verifyDefaultLIb_dom.errors.txt has changed.` which is both a file that I don't seem to have and also one I suspect I shouldn't touch. I can't seem to figure out what in my changes would have prompted that but I'm guessing there is another area of the code where I would need to add FormDataEntryValue?

File: baselines/dom.generated.d.ts [hunk 3]
```diff
@@ -3796,7 +3796,12 @@ declare var FocusNavigationEvent: {
 }
 
 interface FormData {
-    append(name: any, value: any, blobName?: string): void;
+    append(name: string, value: string | blob, fileName?: string): void;
```
mhegazy: I think you mean `Blob`, my guess is this is what is causing the test failure.

NColey: You're totally right, thanks so much for catching that. I just fixed it and it seems to be passing now. Thank you!

mhegazy:
thanks!

NColey:
Thank you!

PR #220, opened 2017-03-27
Improve method signatures returning Node type

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/14863

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3208,7 +3208,7 @@ interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEven
       * Gets or sets the version attribute specified in the declaration of an XML document.
       */
     xmlVersion: string | null;
-    adoptNode(source: Node): Node;
+    adoptNode<T extends Node>(source: T): T;
     captureEvents(): void;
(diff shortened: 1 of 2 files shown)


baselines/dom.generated.d.ts
```diff
@@ -8249,20 +8249,20 @@ interface Node extends EventTarget {
     readonly previousSibling: Node | null;
     textContent: string | null;
     appendChild<T extends Node>(newChild: T): T;
-    cloneNode(deep?: boolean): Node;
+    cloneNode(deep?: boolean): this;
```
mhegazy: This will make all the Node hierarchy generic, something we have avoided for perf reasons. see https://github.com/Microsoft/TSJS-lib-generator/pull/166 for more details.

The alternative here is for every interface that extends from Node, we can redefine `cloneNode` to return the containing type.

mhegazy: The other alternative is to just leave `cloneNode` the way it is and take the other changes.

falsandtru: Agreed, I expected this fix to work as an easy workaround. Done.

falsandtru:
@mhegazy

mhegazy:
thanks!

PR #223, opened 2017-03-28
add URL typing to webworker

VincentBel:
`URL` is also available in worker. ([URL IDL spec](https://url.spec.whatwg.org/#idl-index)). Fix https://github.com/Microsoft/TypeScript/issues/14878.

---

**Questions**:

1. `URL` tying is already defined in [`inputfiles/browser.webidl.xml#L11367-L11395`](https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/browser.webidl.xml#L11367-L11395), so it is only need to added to `worker`. So I added `"flavor": "Worker"`, but it seems doesn't work.(same as `"flavor": "All"` for `URLSearchParams`) 😅
2. I'm not sure whether the `URL` and `URLSearchParams` should be added to `WorkerGlobalScope`.  It is signed as `Exposed=(Window,Worker)` in the [spec IDL](https://url.spec.whatwg.org/#idl-index)
3. The spec IDL use `USVString`, but I use `string` here (keep it same as previous added `URLSearchParams` typing). Doesn't it matters?

**Updates**:

1. Just add `URL` and `URLSearcParams` to `knownWorkerInterfaces.json`.
2. `URL` and `URLSearchParams` shouldn't be added to `WorkerGlobalScope`.(The same as [fetch related typing](https://fetch.spec.whatwg.org/#request-class))

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -13859,25 +13859,6 @@ interface ImageBitmap {
     close(): void;
 }
 
-interface URL {
-    href: string;
-    readonly origin: string;
-    protocol: string;
-    username: string;
-    password: string;
-    host: string;
-    hostname: string;
-    port: string;
-    pathname: string;
-    search: string;
-    toJSON(): string;
-}
-
-declare var URL: {
-    prototype: URL;
-    new (url: string, base?: string): URL;
-}
-
 interface URLSearchParams {
     /**
       * Appends a specified key/value pair as a new search parameter.
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -13232,6 +13232,7 @@ interface Window extends EventTarget, WindowTimers, WindowSessionStorage, Window
     readonly top: Window;
     readonly window: Window;
     URL: typeof URL;
+    URLSearchParams: typeof URLSearchParams;
     Blob: typeof Blob;
     customElements: CustomElementRegistry;
     alert(message?: any): void;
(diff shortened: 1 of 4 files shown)


baselines/dom.generated.d.ts
```diff
@@ -13858,6 +13859,25 @@ interface ImageBitmap {
     close(): void;
 }
 
+interface URL {
```
mhegazy: interface `URL` is already defined in this file. no need to redefine it.

baselines/webworker.generated.d.ts
```diff
@@ -1484,6 +1486,25 @@ interface ImageBitmap {
     close(): void;
 }
 
+interface URL {
```
mhegazy: just edit `inputfiles\knownWorkerInterfaces.json` to add `URL` and `URLSearcParams`.

VincentBel:
@mhegazy updated. But ci is still failing. I have see some other PRs, and they seems failed with the same error.

inputfiles/knownWorkerInterfaces.json
```diff
@@ -91,11 +91,13 @@
     "SyncEventInit",
     "SyncManager",
     "USVString",
+    "URL",
+    "URLSearchParams",
```
mhegazy: looks like you need `ObjectURLOptions` as well

VincentBel: Added.

mhegazy:
thanks!

PR #224, opened 2017-03-29
Fix the EventTarget.addEventListener and EventTarget.removeEventListener useCapture argument - related to issue #14188

Diullei:
According to https://dom.spec.whatwg.org/#interface-eventtarget

* The `EventTarget.addEventListener` `useCapture` argument must be (AddEventListenerOptions or boolean)
* The `EventTarget.removeEventListener` `useCapture` argument must be (EventListenerOptions or boolean)

Fix issue https://github.com/Microsoft/TypeScript/issues/14188

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3701,9 +3701,9 @@ declare var Event: {
 }
 
 interface EventTarget {
-    addEventListener(type: string, listener?: EventListenerOrEventListenerObject, useCapture?: boolean): void;
+    addEventListener(type: string, listener?: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
     dispatchEvent(evt: Event): boolean;
-    removeEventListener(type: string, listener?: EventListenerOrEventListenerObject, useCapture?: boolean): void;
+    removeEventListener(type: string, listener?: EventListenerOrEventListenerObject, options?: boolean | EventListenerOptions): void;
 }
 
 declare var EventTarget: {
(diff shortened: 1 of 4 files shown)


Diullei:
This `appvayor` error is not clear to me. Can someone help me to understand what I'm doing wrong?

zhengbli:
The appveyor error means when replacing the default lib.d.ts with your changed version in TypeScript repo, some TypeScript tests failed (in this case the `verifyDefaultLib_webworker` test. 

The test is a simple check: https://github.com/Microsoft/TypeScript/blob/master/tests/cases/compiler/verifyDefaultLib_webworker.ts
to see if the webworker library file contains error.

On a glance it looks like you defined the two new types only in the `dom.generated.d.ts` but not in the worker file, so when you use them in the worker file they could not be found.

Diullei:
Ok, thanks for your help! I'll fix this.

Diullei:
Hi @zhengbli the appveyor build has been fixed.

baselines/dom.generated.d.ts
```diff
@@ -3701,9 +3701,9 @@ declare var Event: {
 }
 
 interface EventTarget {
-    addEventListener(type: string, listener?: EventListenerOrEventListenerObject, useCapture?: boolean): void;
+    addEventListener(type: string, listener?: EventListenerOrEventListenerObject, useCapture?: boolean | AddEventListenerOptions): void;
```
zhengbli: According to the spec, the third argument should be named "options" instead of `useCapture`, can you also correct this?

baselines/dom.generated.d.ts
```diff
@@ -14178,6 +14178,15 @@ interface PromiseRejectionEventInit extends EventInit {
     reason?: any;
 }
 
+interface EventListenerOptions {
+    capture: boolean;
+}
+
+interface AddEventListenerOptions extends EventListenerOptions {
+    passive: boolean;
```
zhengbli: The fact that this type is a `dictionary` in the spec, meaning every property should be optional. Otherwise things like `{ once: true }` would not be accepted. Can you update? The same with `EventListenerOptions.capture`

Diullei:
@zhengbli done!

zhengbli:
Thanks @Diullei !

PR #227, opened 2017-04-02
lib from W3C/WHATWG

### LINKED ISSUE #101: Use WHATWG IDL instead of Edge IDL
I came from https://github.com/Microsoft/TypeScript/issues/3027

This IDL is used as de-facto standard for all major browsers
https://dom.spec.whatwg.org/#idl-index

TypeScript should use this IDL to add and ship features from DOM early from Living Standard.

SaschaNaz:
Fixes #101.

The input XML in this PR is generated by [webidl-serializer](https://github.com/SaschaNaz/webidl-serializer).

The changes in TS.fsx are to:
1. support new `namespace` types (67bcdd6)
2. support multiple constructor signatures (61070d8)
3. add parentheses on variadic union param types -> `someMethod(args: ...(Node | string)[])` (
e105830)
4. add missing IDL primitive type mappings
5. ensure that member conflict solver won't affect other unlisted interfaces (f1c26da)
6. use Flavor.Web (8875852)
7. filter by Exposed instead of Tags (853c7f6)
8. replace webworker.specidl.xml by browser.webidl.xml (103c39e)

Current remaining problems:
- [x] No Streams API spec as it provides no Web IDL. (whatwg/streams#45)
- [x] Type conflict on SVG 2 spec. (Microsoft/TypeScript#15015)
- [x] Merge #239 
- [ ] Check possibly missing standard APIs
- [x] Add element map
- [x] Add CSS property map
- [x] Filter out worker-only interfaces from dom.generated.d.ts
- [x] Fix NodeFilter callback interface being empty

* Note: No SVG 1.1 spec as it is written with OMG IDL which is not compatible with Web IDL. (and probably not type-compatible with SVG 2)


SaschaNaz:
This PR contains several types from Working Draft and Editor's Draft to be compatible with current lib.d.ts where pre-CR specs e.g. Web Audio, Touch Events Level 2, Intersection Observer are already included.

mhegazy:
@SaschaNaz sorry for the delay. i am not sure what is the best way to move on this. it is a bit hard to see the impact of the change due to the size of the change. i wounder if there is a way we make this more managable..

for instance, can we change the current generated file to first have the same order as the new file, the do the change, at least we can limit the diff noise..

SaschaNaz:
I'm not sure there is a good way to reorder current generated file to match this PR, since there is no rule for the order. Instead I can reorder new file by sorting them alphabetically.

mhegazy:
I just need a way to see the changes and asses the differences. We can do changes to the existing repo to reorder first, then you can use the same order well, this way we can compare.

SaschaNaz:
Yes, I mean the existing one is already sorted (though callback-functions are not) so better to fix this PR to match it. The last commit is not sufficient, currently I'm testing locally to make seeing diff easy.

SaschaNaz:
>Yes, I mean the existing one is already sorted

Not true, some are sorted and others are not. Will push another PR to sort them.

SaschaNaz:
I think now it's ready to get some review as I don't see any major/minor problems on code diff itself. If you see some problems please tell me :wink:

falsandtru:
ping @mhegazy

falsandtru:
@mhegazy Are you reviewing this now???
cc @sandersn @DanielRosenwasser

SaschaNaz:
It seems reviewing this PR is too time-consuming because of the diff size. A progressive approach would be great but I'm not sure how as of now.

SaschaNaz:
My current thought: Do not replace XML side, instead add an addedTypes.json-like progressive data.

NaridaL:
I've looked at the dom.generated.d.ts... the order is nearly the same, but some things seem to have moved, e.g. ScrollOptions has moved up 11k lines.

An issue is that the spec source doesn't seem to be quite as detailed in some respects. For example, the static members on WebGlContext are missing. (tbh, I didn't find them in the spec, but all the browsers have them...). Also, some constructors are missing, e.g. for SpeechSynthesisEvent

A lot of interfaces are missing, but it looks like they are mostly Microsoft specific things. Maybe a first step would be to add those types to the removedTypes.json to see if it breaks anything. Note that some types have just been renamed slightly, so it may be necessary to check if the type is being used elsewhere in the file before removing it.

SaschaNaz:
@NaridaL Thank you for your review 👍 

>e.g. ScrollOptions has moved up 11k lines.

Yeah, addedTypes.json adds its types on different position, so this movement occurs.

>For example, the static members on WebGlContext are missing. (tbh, I didn't find them in the spec, but all the browsers have them...).

I'm not sure what you mean, I couldn't find an interface named WebGlContext.

>Also, some constructors are missing, e.g. for SpeechSynthesisEvent

You're right, this is basically because [the spec](https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html) (which hasn't been updated for years) doesn't have it :( Someone [filed a bug](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28867) about this.

>Maybe a first step would be to add those types to the removedTypes.json to see if it breaks anything.

Yeah, probably a separate PR to remove the MS-specific things is the right way to do, and then the next thing is moving them to DefinitelyTyped :smile:

NaridaL:
@SaschaNaz WebGLRenderingContext, sorry. All the constants are available on the global `WebGLRenderingContext` in browsers. (And in the current generated.d.ts)

SaschaNaz:
Ah, now I understand. They are defined in `WebGLRenderingContextBase` which `WebGLRenderingContext` depends on.

NaridaL:
@SaschaNaz The interface is fine, but they're missing on the declared variable. https://github.com/Microsoft/TSJS-lib-generator/pull/227/files#diff-a97a9d0dbf71ec2f96be57f7fb9f70c9L12517

SaschaNaz:
Hmm, it seems that's because `emitConstants` currently only checks the child interface members. I didn't catch that. Thanks!

SaschaNaz:
Superceded by #300.

PR #229, opened 2017-04-04
Add missing optional constructor param for Text

alexeagle:
/cc @samccone

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -11587,7 +11587,7 @@ interface Text extends CharacterData {
 
 declare var Text: {
     prototype: Text;
-    new(): Text;
+    new(data?: string): Text;
 }
 
 interface TextEvent extends UIEvent {
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -1066,5 +1066,12 @@
         "signatures": [
             "new(type: string, touchEventInit?: TouchEventInit): TouchEvent"
         ]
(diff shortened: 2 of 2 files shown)


SaschaNaz:
[The file must not be modified](https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/README.md):
image

alexeagle:
Thanks - in that case, what is the mechanism to change the lib.dom.d.ts
that is distributed with TypeScript? Does it require a change to Edge where
this IDL originates?

SaschaNaz:
You'll want to look at addedTypes.json to add a constructor to an interface.

SaschaNaz:
Make sure you get `pass` on the build script to solve the CI failure! The `generated` directory will have new files and you have to copy-paste them to `baselines` directory.

alexeagle:
The CI only says that the generated file doesn't match the baseline, without printing the diff. I guess I'll have to install Mono so I can generate the files locally.

alexeagle:
Thanks for the help @SaschaNaz !

samccone:
Thanks @alexeagle

mhegazy:
Link to documentation: https://dom.spec.whatwg.org/#text

For future purposes, please file an issue on the TS repo along with the change.

mhegazy:
thanks!

mhegazy:
and thanks @SaschaNaz!

samccone:
thank you for landing @mhegazy

PR #231, opened 2017-04-14
add FormData typing to webworker

VincentBel:
`FormData` is also available in worker. ([IDL spec](https://xhr.spec.whatwg.org/#interface-formdata)). 

### Problem

The `FormData` declaration in DOM is

```ts
declare var FormData: {
    prototype: FormData;
    new (form?: HTMLFormElement): FormData;
}
```

But `HTMLFormElement` is not available in worker. So I think we should change the constructor to:
 (but can't find related info in [w3c spec](https://xhr.spec.whatwg.org/#interface-formdata))

```ts
new (): FormData;
```

form-data

I add a `flavor` to `inputfiles/overridingTypes.json`, but it seems not work. How can I achieve this？

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -699,7 +699,7 @@ module Emit =
     let extendedTypes =
         ["ArrayBuffer";"ArrayBufferView";"Int8Array";"Uint8Array";"Int16Array";"Uint16Array";"Int32Array";"Uint32Array";"Float32Array";"Float64Array"]
 
-    let integerTypes = 
+    let integerTypes =
         ["byte";"octet";"short";"unsigned short";"long";"unsigned long";"long long";"unsigned long long"]
 
     /// Get typescript type using object dom type, object name, and it's associated interface name
@@ -731,7 +731,7 @@ module Emit =
                 // Name of an interface / enum / dict. Just return itself
                 if allInterfacesMap.ContainsKey objDomType ||
                     allCallbackFuncs.ContainsKey objDomType ||
-                    allDictionariesMap.ContainsKey objDomType || 
+                    allDictionariesMap.ContainsKey objDomType ||
                     allEnumsMap.ContainsKey objDomType then
                     objDomType
                 // Name of a type alias. Just return itself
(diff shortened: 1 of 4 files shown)


inputfiles/overridingTypes.json
```diff
@@ -338,6 +338,7 @@
     {
         "kind": "constructor",
         "interface": "FormData",
+        "flavor": "DOM",
```
mhegazy: i think you mean `"web"`

VincentBel: I change it to `Web`, but seem still not work.

VincentBel: And I guess the origin `FormData` constructor definition is in[`inputfiles/browser.webidl.xml#L4303`](https://github.com/VincentBel/TSJS-lib-generator/blob/18e8e68d6e77f7399ce7f96b5f11da1a2f41e50b/inputfiles/browser.webidl.xml#L4303).

baselines/webworker.generated.d.ts
```diff
@@ -450,6 +450,15 @@ declare var FileReader: {
     new(): FileReader;
 }
 
+interface FormData {
+    append(name: string, value: string | Blob, fileName?: string): void;
+}
+
+declare var FormData: {
+    prototype: FormData;
+    new (form?: HTMLFormElement): FormData;
```
mhegazy: Tis is failing the CI tests:

> generated/webworker.generated.d.ts(459,17): error TS2304: Cannot find name 'HTMLFormElement'.

you will need to update this for WebWorkers only to be `string` i presume.

TS.fsx
```diff
@@ -1086,13 +1086,13 @@ module Emit =
...
-    let EmitConstructorSignature (i:Browser.Interface) =
+    let EmitConstructorSignature flavor (i:Browser.Interface) =
         let emitConstructorSigFromJson (c: InputJsonType.Root) =
             c.Signatures |> Array.iter (Pt.Printl "%s;")
 
-        let removedCtor = getRemovedItems ItemKind.Constructor Flavor.All  |> Array.tryFind (matchInterface i.Name)
+        let removedCtor = getRemovedItems ItemKind.Constructor flavor  |> Array.tryFind (matchInterface i.Name)
         if Option.isNone removedCtor then
-            let overriddenCtor = getOverriddenItems ItemKind.Constructor Flavor.All  |> Array.tryFind (matchInterface i.Name)
+            let overriddenCtor = getOverriddenItems ItemKind.Constructor flavor  |> Array.tryFind (matchInterface i.Name)
```
VincentBel: @mhegazy After looking into this generating script, I found out the reason why `"flavor": "Web"`  in `inputfiles/overridingTypes.json` is ignored is that it always use `Flavor.All` to EmitConstructorSignature.

baselines/webworker.generated.d.ts
VincentBel: @mhegazy [The IDL spec](https://xhr.spec.whatwg.org/#interface-formdata) says `Constructor(optional HTMLFormElement form)`. So I think it should be empty rather than a `string`.

PR #234, opened 2017-04-19
Update ParentNode interface

falsandtru:
Fixes Microsoft/TypeScript#14246

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -14129,6 +14129,10 @@ interface ParentNode {
     readonly firstElementChild: Element | null;
     readonly lastElementChild: Element | null;
     readonly childElementCount: number;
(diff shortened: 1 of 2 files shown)


baselines/dom.generated.d.ts
```diff
@@ -14129,6 +14129,10 @@ interface ParentNode {
     readonly firstElementChild: Element | null;
     readonly lastElementChild: Element | null;
     readonly childElementCount: number;
+    querySelector<K extends keyof ElementTagNameMap>(selectors: K): ElementTagNameMap[K] | null;
```
mhegazy: why not just make `ParentNode` extends from `NodeSelector`?

SaschaNaz: NodeSelector is removed from the spec and merged into ParentNode.

mhegazy:
closing for now. please reopen with updated files.

PR #235, opened 2017-04-20
Support implicit iterable interfaces

SaschaNaz:
This generates a new `dom.es6.generated.d.ts` which contains `[Symbol.iterator]` index signature for some interfaces. ('es6' because it should ultimately support `setlike` and `maplike` which will add some methods including `keys()` `values()` `entries()` etc.)

Copying from Microsoft/TypeScript#5329:

>[WebIDL says](https://heycam.github.io/webidl/#es-iterators):
>
>>If the interface has any of the following:
>>    * an iterable declaration
>>    * an indexed property getter and an integer-typed attribute named “length”
>>    * a maplike declaration
>>    * a setlike declaration
>>
>>then a property must exist whose name is the `@@iterator` symbol, with attributes `{ [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true }` and whose value is a function object.
>
>[Chromium on April 2016](https://bugs.chromium.org/p/chromium/issues/detail?id=538558#c23) implemented and merged the change based on the criteria number 2.
>
>This means we can add `[Symbol.iterator]` key for some interfaces and also means that TSJS-lib-generator can automatically add the key but without methods e.g. `values()`.


TS.fsx
```diff
@@ -1466,8 +1468,8 @@ module Emit =
         Pt.Reset()
         Pt.Printl "/////////////////////////////"
         match flavor with
-        | Worker -> Pt.Printl "/// IE Worker APIs"
-        | _ -> Pt.Printl "/// IE DOM APIs"
+        | Worker -> Pt.Printl "/// MSEdge Worker APIs"
```
mhegazy: i would remove `MSEdge` / `IE` alltogether.

mhegazy:
@SaschaNaz we need to come up with a way to merge this with dom.iterable.d.ts in src/lib in the ts repo.

baselines/dom.es6.generated.d.ts
```diff
@@ -0,0 +1,104 @@
...
+
+interface CSSStyleDeclaration {
+    [Symbol.iterator](): IterableIterator<string>
+}
+
+interface ClientRectList {
+    [Symbol.iterator](): IterableIterator<ClientRect>
+}
+
+interface DOMStringList {
```
mhegazy: does not seem itrable, https://www.w3.org/TR/DOM-Level-3-Core/core.html#DOMStringList

SaschaNaz: It meets the standard on WHATWG spec:

```WebIDL
interface DOMStringList {
  readonly attribute unsigned long length;
  getter DOMString? item(unsigned long index);
  boolean contains(DOMString string);
};
```

It has 1. an indexed property getter and 2. an integer-typed attribute named “length”.

SaschaNaz:
>we need to come up with a way to merge this with dom.iterable.d.ts in src/lib in the ts repo.

That file should ultimately be completely replaced by generated one, once we get #227 merged (as current input xml file does not have the data). I think we have two options, we can:

1. have two files temporarily and merge it on TS builder side
2. support manual type adding as current addedTypes.json does and remove dom.iterable.d.ts completely

SaschaNaz:
I'm not sure what happened on Travis side.

mhegazy:
new mono 5.0, not sure why..

PR #236, opened 2017-04-23
Add typing for Navigator.getBattery()

Jack-Works:
https://github.com/Microsoft/TypeScript/issues/15314
https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API
https://www.w3.org/TR/battery-status

Signed-off-by: Jack Works <zjwpeter@gmail.com>

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -8209,7 +8209,7 @@ declare var NavigationEventWithReferrer: {
     new(): NavigationEventWithReferrer;
 }
 
-interface Navigator extends Object, NavigatorID, NavigatorOnLine, NavigatorContentUtils, NavigatorStorageUtils, NavigatorGeolocation, MSNavigatorDoNotTrack, MSFileSaver, NavigatorBeacon, NavigatorConcurrentHardware, NavigatorUserMedia {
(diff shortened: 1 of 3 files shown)


noomorph:
@Jack-Works , isn't it deprecated?

https://www.fxsitecompat.com/en-CA/docs/2016/battery-status-api-has-been-removed/

AFAIK, only privileged apps can use it.

Jack-Works:
Umm...
This API is sound removed.
https://webkit.org/status/#specification-battery-status-api

Edge is considering support this API: http://status.modern.ie/batterystatusapi
Chrome is shipped.
Firefox used to shipped but removed it (unless in privileged code).
Safari does not support this.

So this is a Chrome-Only API now.
But I think this should be added to the navigator.
Maybe I can add some API limited warning on it?

mhegazy:
If this is not a standard feature, nor on the standard track, it should not be included.

Jack-Works:
Alright

elasim:
Here is up to date status.

Draft has been updated recently. (Draft 12 June 2018)
Chrome is shipped long times ago.
Firefox is shipped
WebKit is removed because of privacy matter.

@mhegazy @Jack-Works 
Can we discuss this issue again?

Jack-Works:
I see there is no sign that battery api come back again. According to the principle of lib.d I guess it has no way to merge in.

stefann01:
Why is this not introduced? From what I see it can be used in some browsers, and it is specified in the mdn docs.
https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API
https://caniuse.com/?search=battery.

mathe42:
It is only supported in chromium based browsers (= 1 implementation) but we require 2 implementations to support it to add it. And that will probably not happen (in near future).

---
Chrome, Edge, Opera all the same implementation.

PR #240, opened 2017-04-27
Improve type safety of `name` global variable

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/15424


mhegazy:
as noted in https://github.com/Microsoft/TypeScript/issues/15424#issuecomment-301207034, `never` was added for a specific reason. It is much more common for users to use the global `name` by mistake vs intentionally. this change was meant to alert users in these cases; unfortunately it breaks other cases, but we believe this gets the 90% case right. for other cases, `window.name`.

PR #243, opened 2017-04-30
Update IndexedDB API

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/14872, https://github.com/Microsoft/TypeScript/issues/15483

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -7010,7 +7010,7 @@ interface IDBDatabase extends EventTarget {
     close(): void;
     createObjectStore(name: string, optionalParameters?: IDBObjectStoreParameters): IDBObjectStore;
     deleteObjectStore(name: string): void;
-    transaction(storeNames: string | string[], mode?: string): IDBTransaction;
+    transaction(storeNames: string | string[], mode?: IDBTransactionMode): IDBTransaction;
     addEventListener(type: "versionchange", listener: (ev: IDBVersionChangeEvent) => any, useCapture?: boolean): void;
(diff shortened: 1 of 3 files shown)


baselines/dom.generated.d.ts
```diff
@@ -7112,7 +7112,7 @@ interface IDBRequestEventMap {
 }
 
 interface IDBRequest extends EventTarget {
-    readonly error: DOMError;
+    readonly error: DOMException | DOMError;
```
SaschaNaz: DOMError is deprecated, isn't DOMException enough?

falsandtru: I cannot decide it. @mhegazy ?

SaschaNaz: Structurally DOMException is a subtype of DOMError so it shouldn't cause any compatibility issue.

falsandtru: These two have an important difference.

```
> new DOMError('') instanceof Error
false
> new DOMException('') instanceof Error
true
```

SaschaNaz: Right but that's 100% browser dependent behavior and TS type does not affect it.

falsandtru: Then we lose to know this difference through TS type. Type names are important info too.

falsandtru: @mhegazy What do you think?

falsandtru:
@mhegazy

baselines/dom.generated.d.ts
mhegazy: I would say `DOMException`. from usage perspective `DOMException` has the same props as `DOMError`, so it is not going to break existing code. but it allows for users using new browsers to use the new APIs.

falsandtru: Done, but travis seems to have a problem. Please restart the build after it is fixed.

PR #246, opened 2017-05-06
Add an 'InsertPosition' string literal type, to be used by the element insertion methods (insertAdjacentElement / insertAdjacentHTML / insertAdjacentText)

noobiept:
Here's the associated report: https://github.com/Microsoft/TypeScript/issues/15620

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -1075,12 +1075,6 @@
         ]
     },
     {
-        "kind": "typedef",
-        "name": "InsertPosition",
-        "flavor": "Web",
-        "type": "\"beforebegin\" | \"afterbegin\" | \"beforeend\" | \"afterend\""
-    },
-    {
         "kind": "method",
         "interface": "Element",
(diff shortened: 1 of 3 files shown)


noobiept:
The method changes in the 'overridingTypes.json' aren't updating the 'dom.generated.d.ts' file after running the build script. 
I'm probably missing a step somewhere.

inputfiles/overridingTypes.json
```diff
@@ -1073,5 +1073,23 @@
         "signatures": [
             "new(data?: string): Text"
         ]
+    },
+    {
```
mhegazy: these methods are acctually defined in `inputFiles/addedTypes.json` see https://github.com/noobiept/TSJS-lib-generator/blob/3855f80a4f24e558633d620bfb22f5b67495af97/inputfiles/addedTypes.json#L1225, so i would move these there instead.

You should also see the change in dom.generated.d.ts above.

noobiept:
I think its correct now, thanks for the help.

PR #247, opened 2017-05-09
Fix Console types in browser.webidl.xml

carloslfu:
Add `optionalParams` to `group` and `groupCollapsed` methods of interface `Console`


mhegazy:
Do not change this file, change https://github.com/Microsoft/TSJS-lib-generator/blob/fb07b7023898620da5315be14b17352e4a0c6374/inputfiles/overridingTypes.json#L286-L327

PR #249, opened 2017-05-09
Updated HTMLMediaElement.play return type to Promise<void>

majames:
Closes issue #15691

https://github.com/Microsoft/TypeScript/issues/15691

* Previously the return type was void
* According to https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-play it should be Promise<void>

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5725,7 +5725,7 @@ interface HTMLMediaElement extends HTMLElement {
     /**
       * Loads and starts playback of a media resource.
       */
-    play(): void;
+    play(): Promise<void>;
     setMediaKeys(mediaKeys: MediaKeys | null): Promise<void>;
     readonly HAVE_CURRENT_DATA: number;
     readonly HAVE_ENOUGH_DATA: number;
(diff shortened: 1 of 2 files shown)


mhegazy:
@SaschaNaz can you please review this change?

SaschaNaz:
I'm not 100% sure about the policy, but if we prefer WHATWG spec over W3C one then LGTM to me, if not I think it should be `Promise<void> | void` for W3C/MSEdge compatibility.

majames:
I'd be happy to go with `Promise<void> | void` 👍

mhegazy:
what do current browsers support?

SaschaNaz:
Firefox: Promise<void>
Chrome: Promise<void>
Edge: void

mhegazy:
I would say it should be `Promise<void>` then. Edge has been moving to align with the spec lately, so this seems like a point-in-time issue.

SaschaNaz:
LGTM then :smile:

mhegazy:
thanks @majames and thanks @SaschaNaz for the review

PR #250, opened 2017-05-11
Override Console types

carloslfu:
Add `optionalParams` to `group` and `groupCollapsed` methods of interface `Console`


mhegazy:
you need to run `build` locally, then copy generated/* to baselines/* and include them in your PR.

PR #255, opened 2017-05-13
Apply Flavor.Web and remove MSApp interfaces

SaschaNaz:
Current Flavor.All causes dom.generate.d.ts include types marked as worker-only. This PR applies Flavor.Web and remove them.

(Latest specs have some worker-only interfaces e.g. [FileReaderSync](http://www.w3.org/TR/FileAPI/#FileReaderSync))

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -1547,7 +1547,7 @@ module Emit =
         target.Close()
 
     let EmitDomWeb () =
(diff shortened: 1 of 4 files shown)


SaschaNaz:
Probably better to merge this PR to WHATWG lib PR.

PR #269, opened 2017-06-19
Add valid init types to URLSearchParams constructor

jacobrask:
See https://url.spec.whatwg.org/#interface-urlsearchparams
Added sequence<sequence<USVString>> and record<USVString, USVString>.

Fixes https://github.com/Microsoft/TypeScript/issues/15338

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -182,7 +182,7 @@
         "kind": "interface",
         "name": "URLSearchParams",
         "constructorSignatures": [
-            "new (init?: string | URLSearchParams): URLSearchParams"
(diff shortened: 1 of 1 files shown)


Kovensky:
You also need to update the baselines (run `./build.sh`, copy `generated/*` into `baselines/`).

I also checked and it doesn't support `Iterable` :cry:

mhegazy:
As @Kovensky [noted](https://github.com/Microsoft/TSJS-lib-generator/pull/269#issuecomment-312148457), you need to update the baselines for the CI server to pass. feel free to reopen with the missing files.

PR #270, opened 2017-06-29
Add isIntersecting key to IntersectionObserverEntry

Kovensky:
It's missing in the IDL file.

https://wicg.github.io/IntersectionObserver/#intersection-observer-entry

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -207,6 +207,7 @@ interface IDBObjectStoreParameters {
 }
 
 interface IntersectionObserverEntryInit {
+    isIntersecting: boolean;
     boundingClientRect?: DOMRectInit;
     intersectionRect?: DOMRectInit;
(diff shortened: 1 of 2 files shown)


baselines/dom.generated.d.ts
```diff
@@ -207,6 +207,7 @@ interface IDBObjectStoreParameters {
 }
 
 interface IntersectionObserverEntryInit {
+    isIntersecting: boolean;
     boundingClientRect?: DOMRectInit;
```
Kovensky: These keys are incorrectly marked as optional; they're `required` in the spec.

It turns out that the generator script doesn't support `required` so I fixed this in https://github.com/Microsoft/TSJS-lib-generator/pull/271

Kovensky:
The official polyfill now supports adding `isIntersecting` to browsers that support `IntersectionObserver` but don't support it.

PR #272, opened 2017-07-01
Added dialog element

arjunyel:
Fixes https://github.com/Microsoft/TypeScript/issues/16880


kitsonk:
Experimental technologies that aren't widely adopted in browsers shouldn't be added, IMO.

PR #276, opened 2017-07-28
Change type from Element to HTMLElement for getElementsByClassName

mooman219:
Looks like this was changed from NodeList to HTMLCollection, but the element type was never changed from Element to HTMLElement.

Relevant issues: https://github.com/Microsoft/TypeScript/issues/13297 https://github.com/Microsoft/TypeScript/issues/17470

Thanks @weswigham for the help with this.


DanielRosenwasser:
Just as a heads up, you may want to register your email with GitHub, or amend the email address on your commits.

mooman219:
@DanielRosenwasser Registered that email.
Also apparently Element is the correct return according to this: https://msdn.microsoft.com/en-us/library/ms536439(v=vs.85).aspx closing the PR

PR #277, opened 2017-08-03
Add Path2D string constructor

thw0rted:
Per https://github.com/Microsoft/TypeScript/issues/14660, Path2D should have a constructor overload that takes a single `string.  This [must be added here](https://github.com/Microsoft/TSJS-lib-generator#when-should-a-dom-api-be-included-here) to be consistent with [the WhatWG Canvas spec](https://html.spec.whatwg.org/multipage/canvas.html#dom-path2d), which is a Living Standard.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -8542,6 +8542,7 @@ interface Path2D extends Object, CanvasPathMethods {
 declare var Path2D: {
     prototype: Path2D;
     new(path?: Path2D): Path2D;
+    new(d: string): Path2D;
 };
 
 interface PaymentAddress {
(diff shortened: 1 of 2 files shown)


thw0rted:
I'm passing the CLA to my management, hopefully it's something they can clear quickly.  Sorry for the wait.

mhegazy:
@thw0rted we will not be able to proceed with this PR without the CLA.

thw0rted:
I will bug my management about this again, I had no idea they would take so long getting a response to me.  Sorry!

mhegazy:
> I will bug my management about this again, I had no idea they would take so long getting a response to me. Sorry!

just a reminder you do not need your employers sign off if you do this on your own time.

PR #280, opened 2017-08-12
Fetch API Redirected Property

arjunyel:
Fixes https://github.com/Microsoft/TypeScript/issues/17758


kitsonk:
Just to note it was only recently added to the [living standard](https://www.chromestatus.com/feature/6667148676038656).  It is not present in Edge.

PR #282, opened 2017-08-17
Added HTML main, details, and summary elements

arjunyel:
Fixes https://github.com/Microsoft/TypeScript/issues/17872

All 3 elements are living standards


arjunyel:
@mhegazy if you could help me understand one thing, I started adding the constructors with this change because I saw the other elements had them, is this something I should be doing? Thanks!

mhegazy:
if you add a new type that is supposed to be construct-able, then yes. it all depends on the spec.

PR #284, opened 2017-08-23
Update to WHATWG DOM Living Standard for Events

KamiQuasi:
Referencing this issue - https://github.com/Microsoft/TypeScript/issues/17974

[WHATWG DOM Living Standard](https://dom.spec.whatwg.org/) includes [composed](https://dom.spec.whatwg.org/#dom-event-composed) and [composedPath](https://dom.spec.whatwg.org/#dom-event-composedpath) properties instead of scoped and deepPath. This will allow proper syntax for a case like the following:

```ts
this.dispatchEvent(new CustomEvent('update-term', { detail: { term: this.term }, bubbles: true, composed: true }));
```

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -137,7 +137,7 @@ interface ErrorEventInit extends EventInit {
 }
 
 interface EventInit {
-    scoped?: boolean;
+    composed?: boolean;
     bubbles?: boolean;
     cancelable?: boolean;
 }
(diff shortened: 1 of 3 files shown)


kitsonk:
(just trying to help triage stuff to help @mhegazy, but it is just my opinion)

All of these changes were added ShadowDOM APIs and the changes reflect the current ShadowDOM standards, so I think 👍

atishpatel:
Is there a reason this isn't being accepted?

PR #297, opened 2017-09-16
MediaTrackConstraintSet updates and fixes

chris-morgan:
This gets the definition of that dictionary in line with the current candidate recommendation.


chris-morgan:
The CI failure is spurious (Paket failed to verify the www.nuget.org TLS certificate).

PR #301, opened 2017-09-27
Added interfaces of WebGL extensions in recommended specification status

kyasbal-1994:
This issue is related to  https://github.com/Microsoft/TypeScript/issues/18633 .

Several extensions in draft status are ignored. I just added several interfaces of specification which has recommended status on specification published by Khronos.

See these specifications for more details.

* https://www.khronos.org/registry/webgl/extensions/ANGLE_instanced_arrays/
* https://www.khronos.org/registry/webgl/extensions/EXT_blend_minmax/
* https://www.khronos.org/registry/webgl/extensions/EXT_frag_depth/
* https://www.khronos.org/registry/webgl/extensions/EXT_sRGB/
* https://www.khronos.org/registry/webgl/extensions/EXT_shader_texture_lod/
* https://www.khronos.org/registry/webgl/extensions/EXT_texture_filter_anisotropic/
* https://www.khronos.org/registry/webgl/extensions/OES_vertex_array_object/
* https://www.khronos.org/registry/webgl/extensions/WEBGL_color_buffer_float/
* https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_astc/
* https://www.khronos.org/registry/webgl/extensions/WEBGL_debug_shaders/
* https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_s3tc_srgb/
* https://www.khronos.org/registry/webgl/extensions/WEBGL_draw_buffers/
* https://www.khronos.org/registry/webgl/extensions/WEBGL_lose_context/


inputfiles/addedTypes.json
```diff
@@ -1765,5 +1765,568 @@
...
+                    "drawBuffersWEBGL(buffers: number[]): void"
+                ]
+            }
+        ]
+    },
+    {
+        "kind": "interface",
+        "name": "WEBGL_lose_context",
+        "flavor": "DOM",
+        "constructorSignatures": [
```
mhegazy: i do not think you need these. these are types only and do not exisit at runtime.

baselines/dom.generated.d.ts
```diff
@@ -14262,6 +14284,188 @@ declare var HTMLSummaryElement: {
     new(): HTMLSummaryElement;
 };
 
+interface EXT_blend_minmax {
+    readonly MIN_EXT: number;
+    readonly MAX_EXT: number;
+}
+
+declare var EXT_blend_minmax: {
```
mhegazy: not needed

SaschaNaz: `EXT_blend_minmax` is declared as `[NoInterfaceObject]`, so the interface declaration itself is needed but `declare var` part is not needed.

inputfiles/addedTypes.json
SaschaNaz: Only `constructorSignatures` should be removed.

kyasbal-1994:
Sorry for late to fix this PR. I have fixed 2 reviewed points.
Could you review again?

baselines/dom.generated.d.ts
```diff
@@ -12479,7 +12479,29 @@ interface WebGLRenderingContext {
...
+    getExtension(extensionName: "OES_vertex_array_object"): OES_vertex_array_object | null;
+    getExtension(extensionName: "WEBGL_color_buffer_float"): WEBGL_color_buffer_float | null;
+    getExtension(extensionName: "WEBGL_compressed_texture_astc"): WEBGL_compressed_texture_astc | null;
+    getExtension(extensionName: "WEBGL_compressed_texture_s3tc_srgb"): WEBGL_compressed_texture_s3tc_srgb | null;
+    getExtension(extensionName: "WEBGL_debug_shaders"): WEBGL_debug_shaders | null;
+    getExtension(extensionName: "WEBGL_draw_buffers"): WEBGL_draw_buffers | null;
+    getExtension(extensionName: "WEBGL_lose_context"): WEBGL_lose_context | null;
+    getExtension(extensionName: "WEBGL_depth_texture"): WEBGL_depth_texture|null;
+    getExtension(extensionName: "WEBGL_debug_renderer_info"): WEBGL_debug_renderer_info|null;
+    getExtension(extensionName: "WEBGL_compressed_texture_s3tc"): WEBGL_compressed_texture_s3tc|null;
```
mhegazy: nit. can you make sure there is a space before and after `|`.

baselines/dom.generated.d.ts
```diff
@@ -12479,7 +12479,29 @@ interface WebGLRenderingContext {
     getBufferParameter(target: number, pname: number): any;
     getContextAttributes(): WebGLContextAttributes;
     getError(): number;
-    getExtension(name: string): any;
+    getExtension(extensionName: "EXT_blend_minmax"): EXT_blend_minmax | null;
```
mhegazy: why do you need ` | null`? should not all these extensions be supported?

baselines/dom.generated.d.ts
```diff
@@ -14627,6 +14649,178 @@ declare var HTMLSummaryElement: {
...
 
+interface EXT_blend_minmax {
+    readonly MIN_EXT: number;
+    readonly MAX_EXT: number;
+}
+
+interface EXT_frag_depth {
+}
+
+declare var EXT_frag_depth: {
```
mhegazy: you do not need all the `declare var`s. please remove them all.

baselines/dom.generated.d.ts
kyasbal-1994: `getExtension` can return null if the extension is not supported on the environment.
This is not only depends on which browser the user using, but also depends on GPU chips, display drivers or browser extension flags.
Therefore, developer should consider the situation the returned value is null.

baselines/dom.generated.d.ts
kyasbal-1994: I fixed this with last commit.

baselines/dom.generated.d.ts
kyasbal-1994: Fixed also.

PR #303, opened 2017-10-03
add HTMLFormElement.reportValidity

SimonSchick:
As seen on https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-reportvalidity and https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reportValidity 

I hope I did everything correctly!

Fixes https://github.com/Microsoft/TypeScript/issues/18733

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4838,6 +4838,7 @@ interface HTMLFormElement extends HTMLElement {
      * Fires when a FORM is about to be submitted.
      */
     submit(): void;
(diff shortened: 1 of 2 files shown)


chdh:
Shouldn't `reportValidity()` also be added to `HTMLInputElement`, `HTMLSelectElement`, etc.? The [first link you provided](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-cva-reportvalidity) is for the "submittable elements", not for the `HTMLFormElement` itself.

PR #304, opened 2017-10-06
Parameterizes CustomEvent and CustomEventInit

Todberg:
In relation to the open issue, https://github.com/Microsoft/TypeScript/issues/14785.
This guy @arjunyel creates a commit, but this never makes it to a PR. So here it is!

NOTE that I was not able to run the tests sucessfully when I ran `build.cmd`.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -82,8 +82,8 @@ interface ConstrainVideoFacingModeParameters {
     ideal?: VideoFacingModeEnum | VideoFacingModeEnum[];
 }
 
-interface CustomEventInit extends EventInit {
-    detail?: any;
+interface CustomEventInit<T = any> extends EventInit {
+    readonly detail?: T;
 }
 
 interface DeviceAccelerationDict {
@@ -2366,9 +2366,9 @@ declare var CSSSupportsRule: {
     new(): CSSSupportsRule;
 };
 
-interface CustomEvent extends Event {
-    readonly detail: any;
-    initCustomEvent(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: any): void;
+interface CustomEvent<T = any> extends Event {
+    readonly detail: T;
+    initCustomEvent(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: T): void;
 }
 
 declare var CustomEvent: {


arjunyel:
Thank you for finishing it!

Todberg:
No problem! Do you know why the tests are failing? I followed the instructions in the readme..

arjunyel:
You have to add the type information to inputfiles/addedTypes.json or inputfiles/overridingTypes.json I'm not sure which for this pull request, that's why I didn't submit it :D

You need to add the typing information to the JSON and then change the file in baselines/dom.generated.d.ts

When you run the build.cmd it creates the file in generated/dom.generated.d.ts using the info in the json files, then it compares it to what you manually created in the baselines folder. That test needs to pass in order for the change to be successful.

Here's an example of one of my pull requests https://github.com/Microsoft/TSJS-lib-generator/pull/282/files

Todberg:
Thanks for the explaination. I manged to figure out how to override the properties and methods, however, I'm having difficulties figuring out how to override optional properties and interfaces.

## Optional Properties
Changes to `CustomEventInit` in `baselines/dom.generated.d.ts`:
```TypeScript
interface CustomEventInit extends EventInit {
    readonly detail?: T;
}
```
Changes to `overridingTypes.json`:
```json
{
  "kind": "property",
  "interface": "CustomEventInit",
  "readonly": true,
  "name": "detail",
  "type": "T"
}
```

I suspect that it doesn't work because of the optional `?` in the baseline file! How can I state that in the json?

## Interfaces
Changes to `CustomEventInit` in `baselines/dom.generated.d.ts`:
```TypeScript
interface CustomEventInit<T = any> extends EventInit {
    readonly detail?: T;
}
```
Changes to `overridingTypes.json`:
```json
{
  "kind": "interface",
  "interface": "CustomEventInit",
  "name": "CustomEventInit<T = any>"
 }
```
For this have no idea to what could be wrong? I have looked through some of the merged pull-requests, however, there are no examples of this.

jthoms1:
I wanted this same change, so started doing some research and trying to read the f sharp code. It appears that there is no way to accomplish this in the compiler. I am able to add new interfaces, but the code to modify or remove interfaces does not exist.

Is there another route that should be taken or is a PR needed to update the TS.fsx file? @mhegazy

mhegazy:
For adding the generic type parameter, we will need a new property in `overridingTypes.json`, call it `typeParameter` to be added, and then pick that up and write in `ts.fsx` while we are doing the transformation.

jthoms1:
I will make a PR to allow for overridingTypes with `typeParameter` to work in the ts.fsx file.

mhegazy:
The format of the input files has changed a bit since this PR was opened. closing for now, please reopen after merging with master.

PR #310, opened 2017-10-16
Fix supportedMethods property type

1000ch:
According to https://www.w3.org/TR/payment-request/#paymentmethoddata-dictionary, https://www.w3.org/TR/payment-request/#paymentdetailsmodifier-dictionary.

```ts
interface PaymentMethodData {
    data?: any;
    supportedMethods?: string[];
}
```

should be

```ts
interface PaymentMethodData {
    data?: any;
    supportedMethods?: string;
}
```


1000ch:
@mhegazy Thank you for review.

PR #311, opened 2017-10-20
Fix Microsoft/TypeScript#19277

Kingwl:
1. remove readonly  
2. type to Location | string

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2705,7 +2705,7 @@ interface Document extends Node, GlobalEventHandlers, NodeSelector, DocumentEven
     /**
      * Contains information about the current URL.
      */
-    readonly location: Location;
+    location: Location | string;
     msCapsLockWarningOff: boolean;
     msCSSOMElementFloatMetrics: boolean;
     /**
@@ -13112,7 +13112,7 @@ interface Window extends EventTarget, WindowTimers, WindowSessionStorage, Window
     readonly innerWidth: number;
     readonly isSecureContext: boolean;
     readonly length: number;
-    readonly location: Location;
+    location: Location | string;
     readonly locationbar: BarProp;
     readonly menubar: BarProp;
     readonly msContentScript: ExtensionScriptApis;
@@ -14515,7 +14515,7 @@ declare var innerHeight: number;
 declare var innerWidth: number;
 declare var isSecureContext: boolean;
 declare var length: number;
-declare var location: Location;
+declare var location: Location | string;
 declare var locationbar: BarProp;
 declare var menubar: BarProp;
 declare var msContentScript: ExtensionScriptApis;
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -33,6 +33,12 @@
         "type": "HTMLScriptElement | SVGScriptElement | null"
     },
     {
+        "kind": "property",
+        "interface": "Document",
+        "name": "location",
+        "type": "Location | string"
+    },
+    {
         "kind": "method",
         "interface": "Document",
         "name": "importNode",
@@ -986,6 +992,12 @@
          "name": "opener",
          "type": "any"
     },
+    {
+        "kind": "property",
+        "interface": "Window",
+        "name": "location",
+        "type": "Location | string"
+   },
     {
         "kind": "method",
         "interface": "History",


SaschaNaz:
While the setter allows `string` the getter always returns `Location` not `string`. I think forcing `location.href` is reasonable enough.

Kingwl:
@SaschaNaz  

1. 'get' and 'set' accessor must have the same type  
2. but that is strange

SaschaNaz:
>'get' and 'set' accessor must have the same type

In TypeScript. Not in raw JavaScript world, sadly.

mhegazy:
i am not sure this is the right fix here... we may need to make this `any` instead.. 

we have favored reading over writing before, favoring writing over reading seems to just change the problem instead of fixing it.. seems like `any` is the only "fix" for both cases..

amoshydra:
Wanted to highlight that this PR is reverted in this commit https://github.com/microsoft/TSJS-lib-generator/commit/9e88bfd7f01253b71fd266bcb5e11796942d377e

As of today `typescript@3.7.4` it is not possible to assign type `string` to `Location`

```ts
window.location = '/'
```

Error:
```
Type 'string' is not assignable to type 'Location'.ts(2322)
```

saschanaz:
Is there any reason to not use `window.location.href`?

amoshydra:
> Is there any reason to not use window.location.href?

Apparently yes. There seems to be some differences between setting `location.href` vs `location`. 

Quoting 
> I note that in browsers today, setting window.location = "some string" has special behaviour, see here: [stackoverflow.com/questions/2383401/…](https://stackoverflow.com/questions/2383401/javascript-setting-location-href-versus-location) - see the comments regarding same-site, same-origin and XHR behaviour. – [Dai Dec 16 '19 at 5:54](https://stackoverflow.com/questions/13106950/set-window-location-with-typescript/13106968#comment104897866_13106955)

Was trying to create an issue about this in this repo, however, I am not able to setup an environment to proof those differences mention in the discussion.

saschanaz:
Have you seen that behavior yourself? The mentioned web page doesn't say the href setter will throw (only the getter will), and AFAIK setting `window.location` simply forwards its value to `window.location.href`. (See `PutForwards=href` in the [IDL](https://html.spec.whatwg.org/multipage/window-object.html#window).)

amoshydra:
## 1
> The mentioned web page doesn't say the href setter will throw (only the getter will), 
Not sure what you mean by "throw" (throw error)

Hmm.. not sure what you mean by "href setter will throw". 

What I wanted to highlight from the quoted links are:

> setting location.href mail fail because of same-origin-policy: [javascript.info/tutorial/…](http://javascript.info/tutorial/same-origin-security-policy#can-t-get-but-can-set)
– [Taha Jahangir Aug 25 '14 at 7:49](https://stackoverflow.com/questions/2383401/javascript-setting-location-href-versus-location#comment39768323_2383401)

> window.location.href is somewhat treated like an XHR request. If fired from within an XHR's success callback, window.location.href will be treated as an XHR while window.location emulates clicking on a link.
– [Akshay Raje Jul 6 '16 at 10:35](https://stackoverflow.com/questions/2383401/javascript-setting-location-href-versus-location#comment63865970_2383427)

> window.location does not require SAME ORIGIN.
– [Swiss Mister Jul 4 '14 at 13:55](https://stackoverflow.com/questions/2383401/javascript-setting-location-href-versus-location#comment18225061_10016109)

These commenters claimed to observed different behaviours between the setting `location` and `location.href`

## 2
> Have you seen that behavior yourself?

Unfortunately no. 
1. I was NOT able to reproduce the claim that "window.location.href is somewhat treated like XHR request" in Chrome 79
2. I do not currently have time to create an environment to test out the "same-origin" requirement.

## 3
> and AFAIK setting `window.location` simply forwards its value to `window.location.href`. (See `PutForwards=href` in the [IDL](https://html.spec.whatwg.org/multipage/window-object.html#window).)

Noted.

I made the [original comment](https://github.com/microsoft/TSJS-lib-generator/pull/311#issuecomment-571424203) in hope to document the reversion of this PR. 

I was in the midst of converting existing JS codebase to TS. Our codebase uses `window.location` and I was not expecting the code to throw error in Typescript.

Upon researching I found this issue https://github.com/microsoft/TypeScript/issues/19277 and this PR, thinking that we are able to assign type `string` to `location` again. But that was not the case in my machine.


## 4 My thought/concern on this issue
`location = '/'` is a valid assignment in JavaScript ([according to w3 spec](https://html.spec.whatwg.org/multipage/history.html#location-object-setter-navigate)). However, I am not sure how TypeScript can fascilitate this kind of assignment given "'get' and 'set' accessor must have the same type".

My current requirement allow me to change assignment from `location` to `location.href`, so I am ok with the current workaround at the moment.

At this moment, I am just concerned about the validity of the claims made in the [stackoverflow comments](https://stackoverflow.com/questions/2383401/javascript-setting-location-href-versus-location)

amoshydra:
@saschanaz I now understood what you meant by "doesn't say the href setter will throw". I wasn't reading the [mentioned web page](http://javascript.info/tutorial/same-origin-security-policy#can-t-get-but-can-set) in detail earlier.

Upon testing, it is possible to redirect an iframe by setting both `location` and `location.href` in context where same-origin policy is enforced. 
<details>
<summary>I have tested this behaviour in different browsers.</summary>
<br><table><tr><td>

- Chrome 79
- Firefox
- IE 11 (Also emulated IE 5, 7, 8, 9, 10)
- ~~Edge 44.18362,449~~ (it crashed for both case, seems to be a separate bug with Edge itself)

Test: https://codepen.io/amoshydra/full/YzPeyYb

All browsers are able to redirect an iframe of different origin by setting its `location.href`

</td></tr></table>

</details>

It seems like all those [claims](https://stackoverflow.com/questions/2383401/javascript-setting-location-href-versus-location) about the special behaviour of setting `location` made are invalid / unfounded.

---

I agree with you. Right now I see no reason to not use `window.location.href`

PR #312, opened 2017-10-21
Generate typed signatures for removeEventListenter

thebanjomatic:
These overload signatures are primarily there to suppress errors when strictFunctionTypes is enabled.

https://github.com/Microsoft/TypeScript/issues/19386


mhegazy:
thanks!

PR #315, opened 2017-11-02
Fixed fetch API's HeadersInit, RequestInit and ResponseInit typedef

### LINKED ISSUE #289: Fetch headers as object literal?
Our definition of `RequestInit` includes:
```ts
headers?: Headers | string[][]
```

From [mdn](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch), the headers should be "a Headers object or an object literal with ByteString values". The [isomorphic-fetch](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cb2dbfbb50afbff23de701e8d630f5cfde57832e/types/isomorphic-fetch/isomorphic-fetch-tests.ts) tests are now failing; they try to use `fetch` like:
```ts
fetch("http://my.url.com", { headers: { 'Content-Type': 'application/json' } });
```

This indicates that the second option should be `{ [key: string]: string }`, not `string[][]`.
In TypeScript `2.4` we simply had `headers?: any;`, so the test failures are new.

There are also failing tests for [`chocolatechipjs`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cb2dbfbb50afbff23de701e8d630f5cfde57832e/types/chocolatechipjs/chocolatechipjs-tests.ts).

timocov:
Fixes #289
Fixes Microsoft/TypeScript#19656

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -1381,7 +1381,7 @@
         "kind": "constructor",
         "interface": "Headers",
         "signatures": [
-            "new(init?: Headers | string[][] | object): Headers"
+            "new(init?: HeadersInit): Headers"
         ]
     },
     {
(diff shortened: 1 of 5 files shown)


inputfiles/overridingTypes.json
```diff
@@ -1381,13 +1381,25 @@
...
+    {
+        "kind": "property",
+        "interface": "RequestInit",
+        "name": "headers?",
+        "type": "HeadersInit"
+    },
+    {
+        "kind": "property",
+        "interface": "ResponseInit",
+        "name": "headers?",
```
timocov: Seems it does not work

mhegazy:
Should be superseded by https://github.com/Microsoft/TSJS-lib-generator/pull/316

timocov:
Thanks!

PR #318, opened 2017-11-03
Add generic type to closest method of Element.

jthoms1:
This pull request will allow the Element `closest` method to infer the HTML Element type rather than returning a generic Element.

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -462,6 +462,13 @@
         ]
     },
     {
+        "kind": "method",
+        "interface": "Element",
+        "signatures": [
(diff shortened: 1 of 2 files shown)


jthoms1:
This fixes https://github.com/Microsoft/TypeScript/issues/17550

inputfiles/addedTypes.json
```diff
@@ -462,6 +462,13 @@
         ]
     },
     {
+        "kind": "method",
```
mhegazy: this should come before the string overload. overload resolution is order dependent and `string` is a supertype of all literal strings.

jthoms1:
Ok updated PR, should be corrected now. Thanks

PR #319, opened 2017-11-04
Add overrides for interface types and add CustomEvent types for testing.

jthoms1:
This PR will fix https://github.com/Microsoft/TypeScript/issues/14785 and replace https://github.com/Microsoft/TSJS-lib-generator/pull/304.

In order to override interfaces and add types this PR updates the TS.fsx and adds a new override type called `typedinterface`.  The naming convention was provided here https://github.com/Microsoft/TSJS-lib-generator/pull/304#issuecomment-341561311

This is my first time writing F# so please let me know if I need to make changes to the convention or if I missed an interfaces beyond Dictionaries or Interface Declarations.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -83,7 +83,7 @@ interface ConstrainVideoFacingModeParameters {
 }
 
 interface CustomEventInit<T = any> extends EventInit {
-    detail: T;
+    detail?: T;
 }
 
 interface DeviceAccelerationDict {
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2376,12 +2376,12 @@ declare var CSSSupportsRule: {
 
 interface CustomEvent<T = any> extends Event {
     readonly detail: T;
-    (typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: T): void;
+    initCustomEvent(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: T): void;
 }
 
 declare var CustomEvent: {
     prototype: CustomEvent;
-    new(typeArg: string, eventInitDict?: CustomEventInit): CustomEvent;
+    new<T>(typeArg: string, eventInitDict?: CustomEventInit<T>): CustomEvent<T>;
 };
 
 interface DataCue extends TextTrackCue {
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -82,8 +82,8 @@ interface ConstrainVideoFacingModeParameters {
     ideal?: VideoFacingModeEnum | VideoFacingModeEnum[];
 }
 
-interface CustomEventInit extends EventInit {
-    detail?: any;
+interface CustomEventInit<T = any> extends EventInit {
+    detail: T;
 }
 
 interface DeviceAccelerationDict {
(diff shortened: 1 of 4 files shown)


inputfiles/overridingTypes.json
```diff
@@ -1,5 +1,40 @@
 [
     {
+        "kind": "typedinterface",
```
mhegazy: i do not think you need a new kind here.. just add a property `typeParameters`

jthoms1: So would `kind` be `interface` and then `parameters` would become `typeParameters`?

jthoms1:
Much better approach.  I have updated the code to reflect the requested changes.

inputfiles/overridingTypes.json
```diff
@@ -1,5 +1,40 @@
...
+        "readonly": true,
+        "name": "detail",
+        "type": "T"
+    },
+    {
+        "kind": "method",
+        "interface": "CustomEvent",
+        "name": "initCustomEvent",
+        "signatures": [
+            "(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: T): void"
```
mhegazy: you need to include the name here. i.e. `initCustomEvent(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: T)`

baselines/dom.generated.d.ts
```diff
@@ -2374,9 +2374,9 @@ declare var CSSSupportsRule: {
     new(): CSSSupportsRule;
 };
 
-interface CustomEvent extends Event {
-    readonly detail: any;
-    initCustomEvent(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: any): void;
```
mhegazy: name of the method is missing.

baselines/dom.generated.d.ts
```diff
@@ -2374,9 +2374,9 @@ declare var CSSSupportsRule: {
...
+interface CustomEvent<T = any> extends Event {
+    readonly detail: T;
+    (typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, detailArg: T): void;
 }
 
 declare var CustomEvent: {
```
mhegazy: you need to change this one too, this is your entry point to an event.. it should be:

```ts
declare var CustomEvent: {
prototype: CustomEvent;
     new<T>(typeArg: string, eventInitDict?: CustomEventInit<T>): CustomEvent<T>;
 };
```

jthoms1:
`initCustomEvent` is back in and I also added the type information to the constructor.

baselines/dom.generated.d.ts
```diff
@@ -82,8 +82,8 @@ interface ConstrainVideoFacingModeParameters {
     ideal?: VideoFacingModeEnum | VideoFacingModeEnum[];
 }
 
-interface CustomEventInit extends EventInit {
-    detail?: any;
+interface CustomEventInit<T = any> extends EventInit {
+    detail: T;
```
mhegazy: `detail` should still be optional? or is this intentional?

mhegazy: looks good otherwise. thanks!

jthoms1:
It is now optional. Thanks! Good learning experience I will likely have more PRs in the future.

arjunyel:
@jthoms1 thank you for stepping up and seeing this through! I got stuck at the F# stuff and never made it through

jthoms1:
@arjunyel No worries, I really wanted this. Your work started it! 😄

PR #329, opened 2017-11-21
Add back element tag name map

mhegazy:
https://github.com/Microsoft/TSJS-lib-generator/pull/326 removed `ElementTagNameMap`. turns our types in [Definitelytyped depended on it](https://github.com/DefinitelyTyped/DefinitelyTyped/search?utf8=%E2%9C%93&q=ElementTagNameMap&type=). Putting it back.

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -831,6 +831,10 @@ module Emit =
         Pt.Printl "}"
         Pt.Printl ""
 
+    let EmitElementTagNameMap () =
+        Pt.Printl "interface ElementTagNameMap extends HTMLElementTagNameMap, SVGElementTagNameMap { }"
+        Pt.Printl ""
+
     /// Emit overloads for the createEvent method
     let EmitCreateEventOverloads (m: Browser.Method) =
         if matchSingleParamMethodSignature m "createEvent" "Event" "string" then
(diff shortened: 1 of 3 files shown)


TS.fsx
```diff
@@ -831,6 +831,10 @@ module Emit =
         Pt.Printl "}"
         Pt.Printl ""
 
+    let EmitElementTagNameMap () =
+        Pt.Printl "interface ElementTagNameMap extends HTMLElementTagNameMap, SVGElementTagNameMap { }"
```
falsandtru: Should extend only `HTMLElementTagNameMap`. If not, we can't define other SVG elements such as SVGAElement by extending the interface. So `ElementTagNameMap` should be just an alias of `HTMLElementTagNameMap`.

```
extends HTMLElementTagNameMap { }
```

falsandtru: This is a reasonable compromise as you said.

falsandtru: And `ElementTagNameMap` should be deprecated.

falsandtru:
Why you don't update DefinitelyTyped?

mhegazy:
> Why you don't update DefinitelyTyped?

The only way i could think of is adding some forward references to the new types.. but this seems ugly.. This also indicates that there are other users for the type out there.. having the type in the library is better than breaking users.

PR #332, opened 2017-11-24
Allow URL Object as constructor's base parameter

### LINKED ISSUE #331: new URL() constructor does not take URL object as base parameter
[https://url.spec.whatwg.org/#example-5434421b](https://url.spec.whatwg.org/#example-5434421b)

> A URL object can be used as base URL (while IDL requires a string as argument, a URL object stringifies to its href attribute value):
`var url = new URL("🏳️‍🌈", new URL("https://pride.example/hello-world"))`

Now it's only support DOMString as base parameter, need explicit call toString method on base to avoid compile error.
default
Related: Microsoft/TypeScript#15246

Nullizer:
Fix #331

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -12213,7 +12213,7 @@ interface URL {
 
 declare var URL: {
     prototype: URL;
-    new(url: string, base?: string): URL;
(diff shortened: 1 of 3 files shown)


mhegazy:
thanks!

PR #341, opened 2017-12-12
Type definitions for EventSource

Pappa:
Typescript issue: https://github.com/Microsoft/TypeScript/issues/13666


mhegazy:
merged manually.

PR #342, opened 2017-12-18
Add typemustmatch property on <object>

Vinnl:
As reported in TypeScript issue #20756:
https://github.com/Microsoft/TypeScript/issues/20756

Spec here:
https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-object-typemustmatch


DanielRosenwasser:
Thank you!

PR #352, opened 2018-01-10
Fix callback interfaces

### LINKED ISSUE #351: Wrong signature for WebKitEntriesCallback/WebKitErrorCallback/WebKitFileCallback
I was trying to override callback interface signature for `WebKitEntriesCallback`/`WebKitErrorCallback`/`WebKitFileCallback`, added below `json` in `overridingTypes.json`:
```
    {
        "kind": "callback",
        "interface": "WebKitEntriesCallback",
        "signatures": [
            "(entities: WebKitEntry[]): void"
        ]
    },
    {
        "kind": "callback",
        "interface": "WebKitErrorCallback",
        "signatures": [
            "(err: DOMError): void"
        ]
    },
    {
        "kind": "callback",
        "interface": "WebKitFileCallback",
        "signatures": [
            "(file: File): void"
        ]
    }
```
based on [.file()](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileEntry/file) and [.readEntries()](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryReader/readEntries). Related Typescript repo issue [#18139](https://github.com/Microsoft/TypeScript/issues/18139)

Then I changed `baselines/dom.generated.d.ts` to:
```
interface WebKitEntriesCallback {
    (entries: WebKitEntry[]): void;
}

interface WebKitErrorCallback {
    (err: DOMError): void;
}

interface WebKitFileCallback {
    (file: File): void;
}
```
But when I run `./build` to build and test, `dom.generated.d.ts` is still the same, the snapshot test for it failed.
I am not familiar with F# here, is my `json` input wrong in this case, or it is a bug from the build script?

eg: in `browser.webidl.xml`, these callbacks defined as:
```
  <callback-interfaces>
    ...
    <interface name="WebKitEntriesCallback" extends="Object">
      <methods>
        <method name="handleEvent" type="void">
          <param name="entries" type="WebKitEntry[]"/>
        </method>
      </methods>
    </interface>
    <interface name="WebKitErrorCallback" extends="Object">
      <methods>
        <method name="handleEvent" type="void">
          <param name="err" type="DOMError"/>
        </method>
      </methods>
    </interface>
    <interface name="WebKitFileCallback" extends="Object">
      <methods>
        <method name="handleEvent" type="void">
          <param name="file" type="File"/>
        </method>
      </methods>
    </interface>
  </callback-interfaces>
```

saschanaz:
Fixes #351.

Callback interfaces including WebKitEntriesCallback have incorrect signatures so this fixes them.

IMO the WebKit-specific types should ultimately excluded, but anyway this change will help future WebIDL works. (NodeFilter is a callback interface by standard specification)

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -701,7 +701,6 @@ module Emit =
         | "DOMString" -> "string"
         | "DOMTimeStamp" -> "number"
         | "EndOfStreamError" -> "number"
-        | "EventListener" -> "EventListenerOrEventListenerObject"
         | "double" | "float" -> "number"
         | "object" -> "any"
         | "ReadyState" -> "string"
@@ -1084,7 +1083,7 @@ module Emit =
 
         let emitStringEventHandler (addOrRemove: string) =
             Pt.Printl
-                "%s%sEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | %s): void;"
+                "%s%sEventListener(type: string, listener: EventListener, options?: boolean | %s): void;"
                 fPrefix addOrRemove (getOptionsType addOrRemove)
 
         let tryEmitTypedEventHandlerForInterface (addOrRemove: string) =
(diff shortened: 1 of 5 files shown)


mhegazy:
do we need to remove `EventListenerObject` now? also `EventListenerOrEventListenerObject`.

saschanaz:
We don't need it, good catch!

mhegazy:
thanks!

PR #353, opened 2018-01-11
Code cleanup

saschanaz:
This removes F# lint errors by:

1. deleting unused functions/constants
2. replacing unused identifiers to `_` or `__` (where `_` is not applicable)
3. using linter-suggested simpler code


mhegazy:
sorry for the delay.

PR #356, opened 2018-01-24
Remove alt attribute from HTMLObjectElement interface

benbraou:
alt is not part of the HTML Standard for HTMLObjectElement

Fixes Microsoft/TypeScript#21386

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5961,10 +5961,6 @@ declare var HTMLModElement: {
 interface HTMLObjectElement extends HTMLElement, GetSVGDocument {
     align: string;
     /**
-     * Sets or retrieves a text alternative to the graphic.
-     */
-    alt: string;
-    /**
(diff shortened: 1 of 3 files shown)


mhegazy:
Can you also remove the comment from https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/comments.json#L2811

benbraou:
Done. Thanks!

mhegazy:
u also need to update the baselines after removing the comment.

benbraou:
Done. Thanks for taking time to review :) I see now how comments.json and removedTypes.json play together to generate the baseline

benbraou:
@mhegazy does it make sense to you to update the section in`README` 
```
Add missing elements to inputfiles/addedTypes.json, overriding elements to inputfiles/overridingTypes.json, or elements to remove to inputfiles/removedTypes.json.
```
with an indication that that may also imply to add/remove comments from `inputfiles/comments.json` (depending on the matching types that have been added or removed)

I can do a PR for that

PR #357, opened 2018-01-25
fix querySelector typings

k1r0s:


HolgerJeromin:
Why is that PR needed?
`document.querySelector("svg")`
returns the first [SVGSVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGSVGElement) of the document. This is not an instance of [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement).

k1r0s:
This method is overloaded, other elements can be retrieved with that method, not only SVG instances. My PR fix wrong typings when retriving other elements such as <div>, <span> .. whatever. Basically typings retrieve the wrong Type which missmatch with the retrieved runtime instance.

mhegazy:
> This method is overloaded, other elements can be retrieved with that method, not only SVG instances.

but you removed the generic one that was a catch-all for both `HTMLElement` and `SVGElement`.


I am not sure i understand the issue, can you elaborate on what this change achieves?

Also for future references, please file a bug under https://github.com/Microsoft/TypeScript/issues first.

k1r0s:
Problem is that `Element` doesn't contain some of properties of `HTMLElement`, say `style` for instance. I was wondering how to achieve this. I wasn't aware of `SVGElement`.

Any comments? maybe problem is on `Element` interface :\

saschanaz:
You can do `querySelectorAll<HTMLElement>(".if .you .need")`.

k1r0s:
IMO using generics isn't a good solution.

k1r0s:
By the way HTMLElement as a generic doesn't work either bcoz you have to specificly say what type you expect

k1r0s:
I just checked that both `HTMLElement` and `SVGSVGElement` have an `style` property which should be and abstract one on Element, which now doesn't exist. 

The real problem is on `Element` class

k1r0s:
Maybe I should close this PR and create another

$ inputfiles/addedTypes.json
```
    {
        "kind": "property",
        "interface": "Element",
        "name": "style",
        "type": "CSSStyleDeclaration"
    },
```

HolgerJeromin:
@k1r0s [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) is no TypeScript invention!
Its defined in [DOM](https://dom.spec.whatwg.org/#interface-element).
You see `Element` as equivalent to `HTMLElement | SVGElement` which is the reality for current Browsers, but not according to the standard.

k1r0s:
Because generics aren't working with that signature

The only workarround is to use a Type alias. `document.querySelector(".hello") as HTMLElement`

And isn't enough for me

HolgerJeromin:
@k1r0s 
`document.querySelector<HTMLElement>(".hello")`
is better than the type assertion.

k1r0s:
@HolgerJeromin try it by yourself. Isn't working because you have this signature:

`querySelector<K extends keyof ElementTagNameMap>(selectors: K): ElementTagNameMap[K] | null`

HolgerJeromin:
@k1r0s I did with ts2.6.3 with strictNullCheck active

`let hello = document.querySelector<HTMLElement>(".hello");`
hello is `HTMLElement|null` and transpiled JS is
`var hello = document.querySelector(".hello");`

k1r0s:
yep @HolgerJeromin didn't try latest version. Thanks

HolgerJeromin:
@k1r0s This is [new with 2.6 ](https://github.com/Microsoft/TypeScript/issues/12568)

k1r0s:
very nice. Indeed!

PR #358, opened 2018-01-27
Fix WebSocket.send method

YuichiNukiyama:
fix https://github.com/Microsoft/TypeScript/issues/21266

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -1707,7 +1707,7 @@
         "name": "getBoundingClientRect",
         "signatures": [
             "getBoundingClientRect(): ClientRect | DOMRect"
-        ]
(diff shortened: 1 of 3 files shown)


inputfiles/overridingTypes.json
```diff
@@ -1822,5 +1822,13 @@
         "signatures": [
             "getAttributeNode(name: string): Attr | null"
         ]
+    },
+    {
+        "kind": "method",
+        "interface": "WebSocket",
+        "name": "send",
+        "signatures": [
+            "send(data: USVString | ArrayBuffer | Blob | ArrayBufferView): void;"
```
saschanaz: signatures in `overridingTypes.json` doesn't need semicolons as they are automatically added.

PR #359, opened 2018-01-27
adds onvisibilitychange to Document

DCtheTall:
for TypeScript issue [#21353](https://github.com/Microsoft/TypeScript/issues/21353)

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -2713,5 +2713,11 @@
                 "type": "boolean"
             }
         ]
+    },
+    {
+        "kind": "property",
+        "interface": "Document",
+        "name": "onvisibilitychange",
+        "type": "(this: Document, ev: Event) => any"
     }
 ]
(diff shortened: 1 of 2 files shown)


inputfiles/addedTypes.json
```diff
@@ -2713,5 +2713,11 @@
                 "type": "boolean"
             }
         ]
+    },
+    {
+        "kind": "property",
+        "interface": "Document",
+        "name": "onvisibilitychange",
+        "type": "(this: Document, ev: Event) => any"
```
HolgerJeromin: really `any` or better `void`?

DCtheTall: @HolgerJeromin I just followed the pattern I saw other event listeners on the Document type, it appeared the other type definitions had `any` as the return type.

If you think `void` is more appropriate happy to make that change.

HolgerJeromin: Oh, i have not checked that. I would keep it consistent, too :)

DCtheTall:
@mhegazy I have yet to add support for adding an event listener for `visibilitychange` by using `addEventListener`. Going to try my hand at allowing the `DocumentEventMap` to be modified, I will open another PR if I get something working.

PR #365, opened 2018-01-31
add web-animations types

benbraou:
Added web animation types based on [this](https://developer.mozilla.org/en-US/docs/Web/API/Animation/Animation)

Fixes Microsoft/TypeScript#21492

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4687,6 +4687,7 @@ interface HTMLElement extends Element {
     dragDrop(): boolean;
     focus(): void;
     msGetInputContext(): MSInputMethodContext;
+    animate(keyframes: AnimationKeyFrame | AnimationKeyFrame[], options: number | AnimationOptions): Animation;
     addEventListener<K extends keyof HTMLElementEventMap>(type: K, listener: (this: HTMLElement, ev: HTMLElementEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
     addEventListener(type: string, listener: EventListener, options?: boolean | AddEventListenerOptions): void;
     removeEventListener<K extends keyof HTMLElementEventMap>(type: K, listener: (this: HTMLElement, ev: HTMLElementEventMap[K]) => any, options?: boolean | EventListenerOptions): void;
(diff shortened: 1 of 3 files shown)


baselines/dom.generated.d.ts
```diff
@@ -14852,6 +14853,59 @@ interface EventSourceInit {
     readonly withCredentials: boolean;
 }
 
+interface AnimationOptions {
+    id?: DOMString;
```
mhegazy: we so not use this type else where. just use `string`.

baselines/dom.generated.d.ts
```diff
@@ -14852,6 +14853,59 @@ interface EventSourceInit {
...
+    delay?: number;
+    direction?: "normal" | "reverse" | "alternate" | "alternate-reverse";
+    duration?: number;
+    easing?: string;
+    endDelay?: number;
+    fill?: "none" | "forwards" | "backwards" | "both"| "auto";
+    iterationStart?: number;
```
mhegazy: missing `iterations: number`

baselines/dom.generated.d.ts
```diff
@@ -15360,6 +15414,8 @@ type ScrollRestoration = "auto" | "manual";
 type FormDataEntryValue = string | File;
 type InsertPosition = "beforebegin" | "afterbegin" | "beforeend" | "afterend";
 type HeadersInit = Headers | string[][] | { [key: string]: string };
+type DOMString = string;
+type AnimationKeyFrame = {[key: string]: string | number | [string | number, string | number] | undefined};
```
mhegazy: the spec does not say it has to be a tuple.. there are examples https://drafts.csswg.org/web-animations/#example-be6aa998 where it is an array of strings.

mhegazy: `null` seems to be allowed as well

baselines/dom.generated.d.ts
```diff
@@ -14852,6 +14853,59 @@ interface EventSourceInit {
...
+    readonly finished: Promise<Animation>;
+    id: string;
+    readonly pending: boolean;
+    readonly playState: "idle" | "running" | "paused" | "finished";
+    playbackRate: number;
+    readonly ready: Promise<Animation>;
+    startTime: number;
+    timeline: AnimationTimeline;
+    oncancel() => void;
+    oncancel() => void;
```
mhegazy: repeated.

baselines/dom.generated.d.ts
```diff
@@ -14852,6 +14853,59 @@ interface EventSourceInit {
...
+    effect: AnimationEffectReadOnly;
+    readonly finished: Promise<Animation>;
+    id: string;
+    readonly pending: boolean;
+    readonly playState: "idle" | "running" | "paused" | "finished";
+    playbackRate: number;
+    readonly ready: Promise<Animation>;
+    startTime: number;
+    timeline: AnimationTimeline;
+    oncancel() => void;
```
mhegazy: oncancel(): void;

mhegazy: missing `onFinish`

baselines/dom.generated.d.ts
benbraou: ok => commit will be amended accordingly

baselines/dom.generated.d.ts
benbraou: ok => commit will be amended accordingly

baselines/dom.generated.d.ts
benbraou: ok => commit will be amended accordingly

baselines/dom.generated.d.ts
benbraou: To be more specific about the event type
`oncancel: (this: Animation, ev: AnimationPlaybackEvent) => any;`
I will add the interface `AnimationPlaybackEvent `

baselines/dom.generated.d.ts
benbraou: Done: I defined the type
```typescript
{offset?: number | null | (number | null)[]} & {[key: string]: string | number | number[] | string[]};
```

benbraou:
@mhegazy I have updated the PR to address your review (I added on extra commit, please let me know if I need to squash)
Thanks

PR #366, opened 2018-02-07
HTML Elements in addedTypes.json are mapped tagName -> Interface

e111077:
Fixing Microsoft/TypeScript#21669

See diff without whitespace changes [here](https://github.com/Microsoft/TSJS-lib-generator/compare/master...e111077:master?w=1).

Possible controversial topics in this PR:

1. I modified `inputfiles/sample.json` to include an HTML element with a `tagNames` field.
2. I added a `tagNames` field to `inputfiles/addedTypes.json`'s `HTMLSlotElement`
    1. I figured this was the correct-most way to find associations between tag names and interfaces
    2. It causes duplication of the `slot` property. See [here](https://github.com/e111077/TSJS-lib-generator/blob/3f8d2427b24442d071b2a684ed21f168c03a6861/inputfiles/addedTypes.json#L1541).
3. I added the "addedTypes" to `iNameToIDependList `'s `getExtendList` but not the `getImplementList`
    1. It was required to get through the condition in `EmitHTMLElementTagNameMap`
    2. No `InputJson.ItemKind.Interface` has an implements. This would have required more edits to `inputfiles/sample.json`.
    3. I was on the fence on whether to add it to one or none at all.

My F# is 🙅‍♂️ sharp so thanks for the review!

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -428,9 +428,20 @@ module Data =
             | name when Seq.contains name iNames -> name
             | _ -> raise (Exception("Element conflict occured! Typename: " + tagName))
 
-        [ for i in GetNonCallbackInterfacesByFlavor Flavor.All do
-            yield! [ for e in i.Elements do
-                        yield (e.Name, i.Name) ] ]
+        let nativeTagNamesToInterface =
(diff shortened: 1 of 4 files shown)


mhegazy:
thanks!

PR #369, opened 2018-02-10
Add focusin, focusout events to DocumentEventMap/HTMLElementEventMap

lankaapura:
Fixes https://github.com/Microsoft/TypeScript/issues/21822

### DIFF
--- a/TS.fsx
+++ b/TS.fsx
@@ -1278,6 +1278,13 @@ module Emit =
                 getEventTypeInInterface eHandler.EventName i
             Pt.Printl "\"%s\": %s;" eHandler.EventName eventType
 
+        let emitJsonProperty (p: InputJsonType.Root) =
+            let readOnlyModifier =
+                match p.Readonly with
+                | Some(true) -> "readonly "
+                | _ -> ""
+            Pt.Printl "%s%s: %s;" readOnlyModifier p.Name.Value p.Type.Value
+
         let ownEventHandles = if iNameToEhList.ContainsKey i.Name && not iNameToEhList.[i.Name].IsEmpty then iNameToEhList.[i.Name] else []
         if ownEventHandles.Length > 0 then
             Pt.Printl "interface %sEventMap" i.Name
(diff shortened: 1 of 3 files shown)


inputfiles/browser.webidl.xml
```diff
@@ -3560,6 +3560,8 @@
         <property name="onended" event-handler="ended" type="EventHandler"/>
         <property name="onerror" event-handler="error" type="EventHandler"/>
         <property name="onfocus" event-handler="focus" type="EventHandler"/>
+        <property name="focusin" event-handler="focusin" type="EventHandler"/>
```
mhegazy: these files are autogrenerated and we do not manually change them.So we will need to add support for adding even handler maps to the generation script and add them to overridingTypes.json instead.

lankaapura: @mhegazy thanks, I'll update the PR.

lankaapura: Hi @mhegazy 

I have added following to addedTypes.json but it doesn't generate anything. Do I have to update somewhere else too? I suspect that it's the double quotes in DocumentEventMap properties. (ex: `"blur": FocusEvent;` ) but not sure.

```JavaScript
{
        "kind": "property",
        "interface": "DocumentEventMap",
        "name": "\"focusin\"",
        "type": "FocusEvent"
    },
    {
        "kind": "property",
        "interface": "DocumentEventMap",
        "name": "\"focusout\"",
        "type": "FocusEvent"
    }

```

saschanaz:
Type additions for `*EventMap` interfaces are currently ignored, this is why the test passes without any baseline updates.

lankaapura:
Thanks @saschanaz so do I need to update TS.fsx file to include it? or any other suggestions to fix the issue?

saschanaz:
TS.fsx update is needed.

Sent from my Windows 10 phone

lankaapura:
@saschanaz @mhegazy  PR updated. Pls review.

dumistoklus:
Why this PR not in master branch still?

chicoxyzzy:
They are still not there https://github.com/Microsoft/TSJS-lib-generator/blob/master/src/emitter.ts#L27-L38

stevenyxu:
I also want focusin/focusout because it's inconvenient to keep on writing addEventListener('focusin' as any as keyof HTMLElementEventMap, ...).

But HTMLElementEventMap is generated from https://github.com/Microsoft/TSJS-lib-generator which requires that the event be in a W3C candidate recommendation or further, but I could only find focusin in a working draft https://www.w3.org/TR/uievents/#event-type-focusin. I don't know that it will make it based on the current standard for inclusion.

anyexinglu:
So guys, which version has the 'focusin' and 'focusout'?

saschanaz:
#696 added them again, so I think the next release will have it.

PR #370, opened 2018-02-12
fix: add `escape` to `CSS`

alan-agius4:
Closes: https://github.com/Microsoft/TypeScript/issues/21090

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1819,6 +1819,7 @@ declare var CryptoKeyPair: {
 };
 
 interface CSS {
+    escape(value: string): string;
     supports(property: string, value?: string): boolean;
 }
 declare var CSS: CSS;


baselines/dom.generated.d.ts
```diff
@@ -1819,6 +1819,7 @@ declare var CryptoKeyPair: {
 };
 
 interface CSS {
+    escape(value?: string): string;
```
alan-agius4: to be honest I am not sure 100% that the value should be a `string` as while the input is a `DOMString` from the example it seems that It can accept a number.
Example
```ts
CSS.escape(".foo#bar")        // "\.foo\#bar"
CSS.escape("()[]{}")          // "\(\)\[\]\{\}"
CSS.escape('--a')             // "--a"
CSS.escape(0)                 // "\30 ", the Unicode code point of '0' is 30
CSS.escape('\0')              // "\ufffd", the Unicode REPLACEMENT CHARACTER
```

mhegazy:
The file is automatically generated. please make your edits in https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/overridingTypes.json instead.
See https://github.com/Microsoft/TSJS-lib-generator#contribution-guidelines for more details.

alan-agius4:
@mhegazy 

Question: In noticed that the `CSS.escape` is a `WD` and bases on https://github.com/Microsoft/TSJS-lib-generator#contribution-guidelines. It shouldn't be added here either correct?

mhegazy:
We have been accepting these lately.

alan-agius4:
I think you the new method should go in `addedTypes.json` rather than `overridingTypes.json` as this method is not yet declared.

However, when I do so I end up with something like this;
```json

interface CSS {
    escape(value: string): string;
}

declare var CSS: {
    supports(property: string, value?: string): boolean;
};
```

Maybe you can point me to the right direction?

mhegazy:
yes. you are correct, addedTypes is the right place.

PR #371, opened 2018-02-12
fix: add `escape` to `CSS`

alan-agius4:
Closes:
https://github.com/Microsoft/TypeScript/issues/21090

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1819,9 +1819,12 @@ declare var CryptoKeyPair: {
 };
 
 interface CSS {
-    supports(property: string, value?: string): boolean;
+    escape(value: string): string;
 }
-declare var CSS: CSS;
+
+declare var CSS: {
+    supports(property: string, value?: string): boolean;
+};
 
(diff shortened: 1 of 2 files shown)


baselines/dom.generated.d.ts
```diff
@@ -1819,9 +1819,12 @@ declare var CryptoKeyPair: {
 };
 
 interface CSS {
```
alan-agius4: I'd expect that something like the below to be the generated output

```ts

interface CSS {
    supports(property: string, value?: string): boolean;
    escape(value: string): string;
}

declare var CSS: CSS;
```

However I cannot seem to make it like that.

HolgerJeromin:
This is still a draft. [Really ready](https://github.com/Microsoft/TSJS-lib-generator#w3c) to be included?

baselines/dom.generated.d.ts
mhegazy: the generation script is not handling added static properties correctly.. i think i have a fix for that.

mhegazy:
Should be fixed by https://github.com/Microsoft/TSJS-lib-generator/pull/383

PR #372, opened 2018-02-12
screen: add `lockOrientation` and `unlockOrientation`

alan-agius4:
Closes: https://github.com/Microsoft/TypeScript/issues/20489

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -2926,7 +2926,7 @@
             {
                 "name": "timeline",
                 "type": "AnimationTimeline"
-            }          
+            }
         ],
         "methods": [
             {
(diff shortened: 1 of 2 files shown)


inputfiles/addedTypes.json
```diff
@@ -2980,5 +2980,21 @@
         "signatures": [
             "animate(keyframes: AnimationKeyFrame | AnimationKeyFrame[], options: number | AnimationOptions): Animation"
         ]
+    },
+    {
+        "kind": "method",
+        "name": "lockOrientation",
+        "interface": "Screen",
+        "signatures": [
+            "lockOrientation(orientations: string | string[]): boolean"
```
mhegazy: seems that these are fixed set of strings. https://w3c.github.io/screen-orientation/#orientationtype-enum

PR #374, opened 2018-02-14
Alias EventListenerOrEventListenerObject to EventListener

kube:
This PR puts back `EventListenerOrEventListenerObject` as an alias to `EventEmitter`, which was removed in https://github.com/Microsoft/TSJS-lib-generator/pull/352.

This solves issues with Electron definitions, as stated in:
https://github.com/Microsoft/TypeScript/issues/21922


kube:
Thanks for the feedback.
I'm gonna close this PR though as it's now fixed in two other different PRs. (Which fix the issue in a cleaner way)

PR #375, opened 2018-02-15
Fix few issues with EventListener and EventListenerOrEventObject

mhegazy:
https://github.com/Microsoft/TSJS-lib-generator/pull/352 introduced a few issues, this PR addresses them.

Fix Microsoft/TypeScript#21948
Fix Microsoft/TypeScript#21922

Supersedes https://github.com/Microsoft/TSJS-lib-generator/pull/374


mhegazy:
There are too many dependencies on `EventListener` being an object type and  not a union in our Real World code bases that i do not think it make sense to remove it. it will be a massive breaking change for very little value.
I have added special handling to `EventListener` as a call back interface to avoid the break.

PR #376, opened 2018-02-16
Add proper parentheses + use any return type

saschanaz:
Closes #374, closes #375 

It seems the core issue is the incorrect parentheses (Sorry 😱) and by whatever reason some definitions want to return something.

This fixes parenthesizing and the return type problem.

cc: @mhegazy


mhegazy:
I have additional changes in https://github.com/Microsoft/TSJS-lib-generator/pull/375 as well.

mhegazy:
closing in favor of https://github.com/Microsoft/TSJS-lib-generator/pull/375

PR #380, opened 2018-02-27
update MutationRecord type

den-bozhyk:
According to MDN: https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord

https://github.com/Microsoft/TypeScript/issues/22179


NN---:
@den-bozhyk  What a fast PR :)

PR #383, opened 2018-03-01
Convert generation script to TS, and update the source webidl

mhegazy:
This change includes two parts:
1. A complete rewrite of the generator script:
  * Moving from F# to TS to simplify contribution and code review
  * Using JSON files for all inputs instead of xml
2. An update to the Edge webidl file 
  * The new file is more inline with the standard [webidl](https://www.w3.org/TR/WebIDL-1/) format, which will allow us to reuse the script to generate directly from webidl
  * The new file comes in JSON format with no need for preprocessing
  * The new file has multiple changes that bring us closer to the spec, with removals of MS-specific types and adding new definitions

A note about methodology:
I have converted the original F# script to TS first, maintain the input file source, and testing as I went to ensure the output was the same. when that was complete, i then plugged in the new file definition file and continued from there. Unfortunately the new input file is different enough from the new one that it made splitting this change into two more expensive.

A note for reviewers:
There are two things to look out for, first and foremost the changes in the generated files, and second the generation script. 

Testing:
* I have run the local TS tests, as well as the RWC tests. and addressed the issues that came up,
* I still need to run the user tests and DefinitelyTyped on the new generated files.


mhegazy:
@saschanaz  Please take a look at this change. With this in it should be much easier migrate to the standard webidl files. We can add one webidl file at a time, merge the declarations with the ones we have already, and at the end remove the browser webidl file all together.

mhegazy:
@weswigham  and @andy-ms can you please review this change

baselines/dom.generated.d.ts
```diff
@@ -3696,49 +4450,53 @@ interface Element extends Node, GlobalEventHandlers, ElementTraversal, NodeSelec
...
+    readonly tagName: string;
+    attachShadow(shadowRootInitDict: ShadowRootInit): ShadowRoot;
+    closest<K extends keyof HTMLElementTagNameMap>(selector: K): HTMLElementTagNameMap[K] | null;
+    closest<K extends keyof SVGElementTagNameMap>(selector: K): SVGElementTagNameMap[K] | null;
+    closest(selector: string): Element | null;
+    getAttribute(qualifiedName: string): string | null;
     getAttributeNS(namespaceURI: string, localName: string): string;
     getAttributeNode(name: string): Attr | null;
     getAttributeNodeNS(namespaceURI: string, localName: string): Attr | null;
     getBoundingClientRect(): ClientRect | DOMRect;
```
HolgerJeromin: Probably i missed something, but i can not find the definition of ClientRect. This is available in TS master:
https://github.com/Microsoft/TypeScript/blob/bb05122a50fa294ad9ab8da8b19ecbd76fe193e1/lib/lib.dom.d.ts#L1691
I think many are using the ClientRect definition, so this should survive.

baselines/dom.generated.d.ts
```diff
@@ -3846,12 +4613,39 @@ declare var Event: {
...
+    new(url: string, eventSourceInitDict?: EventSourceInit): EventSource;
 };
 
+interface EventSourceInit {
+    readonly withCredentials: boolean;
+}
+
 interface EventTarget {
-    addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
+    addEventListener(type: string, listener?: EventListenerOrEventListenerObject | null, options?: AddEventListenerOptions | boolean): void;
```
HolgerJeromin: Second parameter is not optional.

baselines/dom.generated.d.ts
mhegazy: not sure i understand the question. the definition of `ClientRect` is at https://github.com/Microsoft/TSJS-lib-generator/blob/8d0ab24514f187d005028c8bd77c0439bd36dfe6/baselines/dom.generated.d.ts#L2892

HolgerJeromin: The was no change, so it was collapsed in the github webinterface. I just found it by myself. Sorry for the noise.

baselines/webworker.generated.d.ts
```diff
@@ -172,22 +202,28 @@ declare var Blob: {
...
 
 interface Cache {
-    add(request: RequestInfo): Promise<void>;
-    addAll(requests: RequestInfo[]): Promise<void>;
-    delete(request: RequestInfo, options?: CacheQueryOptions): Promise<boolean>;
-    keys(request?: RequestInfo, options?: CacheQueryOptions): Promise<Request[]>;
-    match(request: RequestInfo, options?: CacheQueryOptions): Promise<Response>;
-    matchAll(request?: RequestInfo, options?: CacheQueryOptions): Promise<Response[]>;
-    put(request: RequestInfo, response: Response): Promise<void>;
+    add(request: Request | string): Promise<void>;
```
weswigham: It seems like `RequestInfo` became `Request | string` (does `RequestInfo` still exist?). Is this a break we should avoid?

baselines/webworker.generated.d.ts
```diff
@@ -578,13 +663,16 @@ declare var Headers: {
     new(init?: HeadersInit): Headers;
 };
 
+interface IDBArrayKey extends Array<number | string | Date | IDBArrayKey> {
+}
+
 interface IDBCursor {
     readonly direction: IDBCursorDirection;
-    key: IDBKeyRange | IDBValidKey;
+    readonly key: IDBKeyRange | number | string | Date | IDBArrayKey;
```
weswigham: Looks like `IDBValidKey` became `number | string | Date | IDBArrayKey` - should is be an alias for it?

baselines/webworker.generated.d.ts
```diff
@@ -1077,8 +1157,8 @@ declare var PushManager: {
 interface PushMessageData {
     arrayBuffer(): ArrayBuffer;
     blob(): Blob;
-    json(): JSON;
-    text(): USVString;
+    json(): any;
```
weswigham: We should avoid using `any` in the lib, right? Meaning this should probably return `object | number | string | Function` or something?

baselines/webworker.generated.d.ts
```diff
@@ -1143,9 +1234,9 @@ interface Request extends Object, Body {
     readonly redirect: RequestRedirect;
     readonly referrer: string;
     readonly referrerPolicy: ReferrerPolicy;
+    readonly signal: any;
```
weswigham: This went from a strongly typed `AbortSignal` to `any`; should it be at least `object`? `any` feels unsafe.

baselines/webworker.generated.d.ts
```diff
@@ -1154,32 +1245,32 @@ declare var Request: {
...
-    readonly redirected: boolean;
     clone(): Response;
 }
 
 declare var Response: {
     prototype: Response;
-    new(body?: any, init?: ResponseInit): Response;
-    error: () => Response;
-    redirect: (url: string, status?: number) => Response;
+    new(body?: Blob | Int8Array | Int16Array | Int32Array | Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array | DataView | ArrayBuffer | FormData | string | null, init?: ResponseInit): Response;
```
weswigham: Isn't there a type alias somewhere we could be using in place of `Int8Array | Int16Array | Int32Array | Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array` here? I think I remember one getting added awhile back.

baselines/webworker.generated.d.ts
```diff
@@ -1287,18 +1380,50 @@ interface URL {
...
-    username: string;
     readonly searchParams: URLSearchParams;
+    username: string;
     toString(): string;
 }
 
 declare var URL: {
     prototype: URL;
-    new(url: string, base?: string | URL): URL;
+    new(url: string, base?: string): URL;
```
weswigham: This looks like a regression. We don't allow passing `URL` as a string, since it isn't one, but passing a URL as the second parameter here seems important.

baselines/webworker.generated.d.ts
```diff
@@ -1530,363 +1652,53 @@ declare var XMLHttpRequestUpload: {
...
 declare var isSecureContext: boolean;
 declare var location: WorkerLocation;
-declare var onerror: (this: DedicatedWorkerGlobalScope, ev: ErrorEvent) => any;
+declare var onerror: ((this: DedicatedWorkerGlobalScope, ev: ErrorEvent) => any) | null;
 declare var performance: Performance;
 declare var self: WorkerGlobalScope;
-declare function msWriteProfilerMark(profilerMarkName: string): void;
 declare function createImageBitmap(image: ImageBitmap | ImageData | Blob, options?: ImageBitmapOptions): Promise<ImageBitmap>;
 declare function createImageBitmap(image: ImageBitmap | ImageData | Blob, sx: number, sy: number, sw: number, sh: number, options?: ImageBitmapOptions): Promise<ImageBitmap>;
+declare function msWriteProfilerMark(profilerMarkName: string): void;
```
weswigham: Is `msWriteProfilerMark` not tagged as an IE/MS-only thing? It looks like one.

baselines/webworker.generated.d.ts
mhegazy: We used to serialize the `type-original` property in the past, and now we use `type`. `type-original` used to have ` or ` in them and `<>` for type arguments, etc.. `type` is more rich, e.g. an array in case of union, and type argumetns being separate, etc.. but they do not use type aliases..

long story short, i will add the type aliases again.

baselines/webworker.generated.d.ts
mhegazy: The alias is still there, i have explicitly added it.. but the references to it have been inlined.

baselines/webworker.generated.d.ts
mhegazy: > We should avoid using any in the lib, right?

well.. i do nto know.. some of these are from the spec, so if not an accurate type, i would not want to mess up with them..

baselines/webworker.generated.d.ts
mhegazy: I can change that.

baselines/webworker.generated.d.ts
mhegazy: BufferSoruce. it should still be there.

baselines/webworker.generated.d.ts
mhegazy: will fix it

baselines/webworker.generated.d.ts
mhegazy: the input file has bugs as well :D. I did talk to the authors in the Edge team and they are aware of some of these untagged declarations, and plan to address them.

I was planning on filtering out all "ms*" and "MS*" declarations, but will get that in a separate change.

baselines/webworker.generated.d.ts
weswigham: I'm saying we should use the union of all possible base types instead of `any`, that way no `any`'s seep into user code from the lib. What was `JSON` defined as before, anyway?

mhegazy:
> It seems generally good; but style nit: will we convert the code to use lowerCamelCase consistently instead of UpperCamelCase? Or are we going to leave it in its mixed state for now?

I tried to convert most of it to lower, if i missed something let me know and i will fix it.

mhegazy:
> Oh, and why are arrays printed as Array<A | B> instead of (A | B)[]?

:) easier to serialize.. let me see if i can enhance that.

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+import * as path from "path";
+import { filter, merge, filterProperties, mapToArray, distinct, map, toNameMap } from "./helpers";
+
+enum Flavor {
+    Web,
+    Worker,
+    ES6Iterators
+}
+
+function EmitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+    // Used to decide if a member should be emitted given its static property and
+    // the intended scope level.
+    function matchScope(scope: EmitScope, x: Browser.Method) {
+        if (scope === EmitScope.All) return true;
+        else if (x.static) return scope === EmitScope.StaticOnly;
+        else return scope === EmitScope.InstanceOnly;
+    }
+
+    /// Parameter cannot be named "default" in JavaScript/Typescript so we need to rename it.
+    function AdjustParamName(name: string) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+                output += newLine;
+                lineStart = true;
+            }
+        }
+
+        reset();
+
+        return {
+
+            Reset: reset,
```
weswigham: `lowerCamelCase` all these members

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+                // only emit the string event handler if we just emited a typed handler
+                emitStringEventHandler(addOrRemove);
+            }
+        }
+
+        emitEventHandler("add");
+        emitEventHandler("remove");
+    }
+
+    function EmitConstructorSignature(i: Browser.Interface) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        // Emit constructor signature
+        if (constructor) {
+            emitSignatures(constructor, "", "new", s => printer.Printl(s));
+        }
+        else {
+            printer.Printl(`new(): ${i.name};`);
+        }
+    }
+
+    function EmitConstructor(i: Browser.Interface) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        EmitConstructorSignature(i);
+        emitConstants(i);
+        emitMembers("", EmitScope.StaticOnly, i);
+
+        printer.DecreaseIndent();
+        printer.Printl("};");
+        printer.Printl("");
+    }
+
+    function EmitNamedConstructor(i: Browser.Interface) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        const nc = i["named-constructor"];
+        if (nc) {
+            printer.Printl(`declare var ${nc.name}: {`);
+            nc.signature.forEach(s => printer.PrintWithAddedIndent(`new(${s.param ? paramsToString(s.param) : ""}): ${i.name};`));
+            printer.Printl(`};`);
+        }
+    }
+
+    /// Emit all the named constructors at root level
+    function EmitNamedConstructors() {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        getElements(webidl.interfaces, "interface")
+            .sort(compareName)
+            .forEach(EmitNamedConstructor);
+    }
+
+    function EmitInterfaceDeclaration(i: Browser.Interface) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+            .map(processIName);
+
+        if (finalExtends && finalExtends.length) {
+            printer.Print(` extends ${finalExtends.join(", ")}`);
+        }
+        printer.Print(" {");
+    }
+
+    /// To decide if a given method is an indexer and should be emited
+    function ShouldEmitIndexerSignature(i: Browser.Interface, m: Browser.Method) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+                    if (mTypes.length === 0 && amTypes.length === 1 && pTypes.length === 0) return amTypes[0] === sig.type;
+                    if (mTypes.length === 1 && amTypes.length === 1 && pTypes.length === 0) return mTypes[0] === amTypes[0] && amTypes[0] === sig.type;
+                    if (mTypes.length === 0 && amTypes.length === 1 && pTypes.length === 1) return amTypes[0] === pTypes[0] && amTypes[0] === sig.type;
+                    if (mTypes.length === 1 && amTypes.length === 1 && pTypes.length === 1) return mTypes[0] === amTypes[0] && amTypes[0] === pTypes[0] && amTypes[0] === sig.type;
+            }
+        }
+        return false;
+    }
+
+    function EmitIndexers(emitScope: EmitScope, i: Browser.Interface) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+            iNameToEhList[i.name]
+                .sort(compareName)
+                .forEach(emitInterfaceEventMapEntry);
+            printer.DecreaseIndent();
+            printer.Printl("}");
+            printer.Printl("");
+        }
+    }
+
+    function EmitInterface(i: Browser.Interface) {
```
weswigham: `lowerCamelCase`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        printer.Printl("");
+
+        if (!printer.StackIsEmpty()) {
+            printer.PrintStackContent();
+            printer.Printl("}");
+            printer.Printl("");
+        }
+    }
+
+    function EmitStaticInterface(i: Browser.Interface) {
```
weswigham: `lowerCamelCase`

baselines/webworker.generated.d.ts
```diff
@@ -619,16 +707,14 @@ interface IDBDatabaseEventMap {
 interface IDBDatabase extends EventTarget {
     readonly name: string;
     readonly objectStoreNames: DOMStringList;
-    onabort: (this: IDBDatabase, ev: Event) => any;
-    onerror: (this: IDBDatabase, ev: Event) => any;
-    version: number;
-    onversionchange: (ev: IDBVersionChangeEvent) => any;
+    onabort: ((this: IDBDatabase , ev: Event) => any) | null;
```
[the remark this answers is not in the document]
mhegazy: fixed

baselines/webworker.generated.d.ts
mhegazy: Added the aliases back

src/index.ts
mhegazy: fixed.

src/index.ts
mhegazy: fixed

src/index.ts
mhegazy: done.

src/index.ts
mhegazy: done.

src/index.ts
mhegazy: done.

src/index.ts
mhegazy: done.

src/index.ts
mhegazy: done.

src/index.ts
mhegazy: done.

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+            webidl.typedefs.typedef
+                .forEach(emitTypeDef);
+        }
+    }
+
+    function compareName(c1: { name: string }, c2: { name: string }) {
+        return c1.name < c2.name ? -1 : c1.name > c2.name ? 1 : 0;
+    }
+
+    function emitTheWholeThing() {
```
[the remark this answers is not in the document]
mhegazy: done.

baselines/webworker.generated.d.ts
mhegazy: done.

baselines/webworker.generated.d.ts
mhegazy: done.

baselines/dom.generated.d.ts
mhegazy: fixed.

mhegazy:
Fixed `Array` to always be `[]`.

baselines/webworker.generated.d.ts
mhegazy: This one specifically came from the webidl file. i would argue it is better than `object | string ..` similar to `JSON.parse()`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+
+        return Object.keys(eventsMap).sort();
+    })();
+
+    /// Interface name to its related eventhandler name list map
+    /// Note:
+    /// In the xml file, each event handler has
+    /// 1. eventhanlder name: "onready", "onabort" etc.
+    /// 2. the event name that it handles: "ready", "SVGAbort" etc.
+    /// And they don't NOT just differ by an "on" prefix!
```
andy-ms: *do not

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        function getParentsWithEventHandler(i: Browser.Interface) {
+            function getParentEventHandler(i: Browser.Interface): Browser.Interface[] {
+                return hasHandler(i) ? [i] : getParentsWithEventHandler(i);
+            }
+
+            const extendedParentWithEventHandler = allInterfacesMap[i.extends] && getParentEventHandler(allInterfacesMap[i.extends]) || [];
+
+            const implementedParentsWithEventHandler =
+                i.implements
+                    ? i.implements.reduce<Browser.Interface[]>((acc, i) => {
```
andy-ms: `flatMap`?

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+            result[i.name] = getParentsWithEventHandler(i);
+        }
+        return result;
+    })();
+
+    return flavor === Flavor.ES6Iterators ? emitES6DomIterators() : emit();
+
+    // Used to decide if a member should be emitted given its static property and
+    // the intended scope level.
+    function matchScope(scope: EmitScope, x: Browser.Method) {
```
andy-ms: Seems like this and several other functions don't depend on any state. Would be nice to move them out of the closure so it's clearer what *is* staeful.

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        if (scope === EmitScope.All) return true;
```
andy-ms: `return scope === EmitScope.All || (scope === EmitScope.StaticOnly) === !!x.static;`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        else if (x.static) return scope === EmitScope.StaticOnly;
+        else return scope === EmitScope.InstanceOnly;
+    }
+
+    /// Parameter cannot be named "default" in JavaScript/Typescript so we need to rename it.
+    function adjustParamName(name: string) {
```
andy-ms:
```ts
const tsKeywords = new Set(["default", "delete", "continue"]);
function adjustParamName(name: string) {
    return tsKeywords.has(name) ? `_${name}` : name;
}
```

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+    function tryGetMatchingEventType(eName: string, i: Browser.Interface) {
+        if (i.events) {
+            const event = i.events.event.find(e => e.name === eName);
+            return event && event.type;
+        }
+        return undefined;
+    }
+
+    function getEventTypeInInterface(eName: string, i: Browser.Interface) {
+        if (eName === "abort" && (i.name === "IDBDatabase" || i.name === "IDBTransaction" || i.name === "MSBaseReader" || i.name === "XMLHttpRequestEventTarget")) return "Event";
```
andy-ms: Might be better to `switch (i.name)`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        else {
+            const ownEventType = tryGetMatchingEventType(eName, i);
+            return ownEventType || eNameToEType[eName] || "Event";
+        }
+    }
+
+    /// Determine if interface1 depends on interface2
+    function dependsOn(i1Name: string, i2Name: string) {
+        return iNameToIDependList[i2Name] && iNameToIDependList[i1Name]
+            ? iNameToIDependList[i1Name].indexOf(i2Name) > -1
```
andy-ms: `iNameToIDependList[i1Name].includes(i2Name)`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        else if (i.name === "XMLHttpRequest") return "ProgressEvent";
+        else {
+            const ownEventType = tryGetMatchingEventType(eName, i);
+            return ownEventType || eNameToEType[eName] || "Event";
+        }
+    }
+
+    /// Determine if interface1 depends on interface2
+    function dependsOn(i1Name: string, i2Name: string) {
+        return iNameToIDependList[i2Name] && iNameToIDependList[i1Name]
```
andy-ms: Does `iNameToIDependList[i2Name]` matter?

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+            ? iNameToIDependList[i1Name].indexOf(i2Name) > -1
+            : i2Name === "Object";
+    }
+
+    // Some params have the type of "(DOMString or DOMString [] or Number)"
+    // we need to transform it into [“DOMString", "DOMString []", "Number"]
+    function decomposeTypes(t: string) {
+        return t.replace(/[\(\)]/g, "").split(" or ");
```
andy-ms: Maybe to be a bit safer:
```ts
if (t[0] !== "(") return [t];
assert(t[t.length - 1] === ")");
return t.slice(0, t.length - 1).split(" or ");
```

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+    }
+
+    function createTextWriter(newLine: string) {
```
andy-ms: Might be better to print to nodes eventually?

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+        const subtypeString = subtype ? subtype.nullable ? makeNullable(subtype.name) : subtype.name : undefined;
+
+        return {
+            name: (type.name === "Array" && subtypeString) ? makeArrayType(subtypeString) : `${type.name}${subtypeString ? `<${subtypeString}>` : ""}`,
+            nullable: type.nullable
+        };
+    }
+
+    function makeArrayType(elementType: string): string {
+        return elementType.indexOf("|") > -1 ? `(${elementType})[]` : `${elementType}[]`;
```
andy-ms: `.includes("|")`

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+    /// Get typescript type using object dom type, object name, and it's associated interface name
+    function convertDomTypeToTsType(obj: Browser.Typed): string {
+        if (obj["override-type"]) return obj["override-type"]!;
+        if (!obj.type) throw new Error("Missing type " + JSON.stringify(obj));
+        const type = convertDomTypeToTsTypeWorker(obj);
+        return type.nullable ? makeNullable(type.name) : type.name;
+    }
+
+    function convertDomTypeToTsTypeWorker(obj: Browser.Typed): { name: string; nullable: boolean } {
+        let type;
```
andy-ms: Looks like `obj["type-original"]` is never used?

This function seems to be equivalent to:
```ts
if (typeof obj.type === "string") {
    const name = covertDomTypeToTsTypeSimple(obj.type);
    const typeArg = obj.subtype && convertDomTypeToTsType(obj.subtype);
    return {
        name: typeArg ? name === "Array" ? makeArrayType(typeArg) : `${name}<${typeArg}>` : name,
        nullable: !!obj.nullable,
    };
}
else {
    assert(!obj.subtype);
    const types = obj.type.map(convertDomTypeToTsTypeWorker);
    return {
        name: types.map(t => t.name).join(" | "),
        nullable: types.some(t => t.nullable),
    };
}
```

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+            nullable: type.nullable
+        };
+    }
+
+    function makeArrayType(elementType: string): string {
+        return elementType.indexOf("|") > -1 ? `(${elementType})[]` : `${elementType}[]`;
+    }
+
+    function covertDomTypeToTsTypeSimple(objDomType: string): string {
+        if (typeof objDomType !== "string") {
```
andy-ms: Was this debugging code?

package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "TSJS-lib-generator",
+  "author": "Microsoft Corp.",
+  "version": "1.0.0",
+  "description": "AppVeyor Status: [![Build status](https://ci.appveyor.com/api/projects/status/8oj3j7u6nvag1xvu/branch/master?svg=true)](https://ci.appveyor.com/project/zhengbli/tsjs-lib-generator/branch/master)\r Travis CI Status: [![Build Status](https://travis-ci.org/Microsoft/TSJS-lib-generator.svg?branch=master)](https://travis-ci.org/Microsoft/TSJS-lib-generator)",
+  "main": "index.js",
+  "scripts": {
+    "build": "tsc --p ./tsconfig.json && node ./lib/index.js",
+    "baseline-accept": "copy /Y generated\\* baselines\\",
```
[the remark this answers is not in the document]
mhegazy: done.

src/helpers.ts
```diff
@@ -0,0 +1,98 @@
...
+            }
+            else {
+                src[k] = target[k];
+            }
+        }
+    }
+    return src;
+}
+
+function mergeNamedArrays<T extends { name: string }>(srcProp: T[], targetProp: T[]) {
```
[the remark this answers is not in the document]
mhegazy: i do not think there is harm from the extra polymorphism

src/index.ts
```diff
@@ -0,0 +1,1368 @@
+import * as Browser from "./types";
+import * as fs from "fs";
+import * as path from "path";
+import { filter, merge, filterProperties, mapToArray, distinct, map, toNameMap } from "./helpers";
+
+enum Flavor {
```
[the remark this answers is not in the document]
mhegazy: done.

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+            return eType === "Event" || dependsOn(eType, "Event")
+                ? eType
+                : defaultEventType;
+        }
+
+        // Get all the event handlers from an interface and also from its inherited / implemented interfaces
+        function getEventHandler(i: Browser.Interface) {
+            const ownEventHandler =
+                i.properties
+                    ? mapToArray(i.properties.property).filter(p => p["event-handler"]).map(p => ({
```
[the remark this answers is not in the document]
mhegazy: done.

src/index.ts
```diff
@@ -0,0 +1,1368 @@
...
+    type EventHandler = { name: string; eventName: string; eventType: string };
+
+    /// Decide which members of a function to emit
+    enum EmitScope {
+        StaticOnly,
+        InstanceOnly,
+        All
+    }
+
+    const pollutor = getElements(webidl.interfaces, "interface").find(i => flavor === Flavor.Web ? !!i["primary-global"] : !!i.global);
```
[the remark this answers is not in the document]
mhegazy: The idea is that the same script can be used on multiple idl files instead of just this one.

src/index.ts
mhegazy: do not think so. so far our transformations are not that complicated..

src/index.ts
mhegazy: yes. i wanted to leave it for now to see what other idl files have

src/index.ts
mhegazy: We should remove this code path altogether anyways, just leaving it for now until we consume other webidl files.

mhegazy:
@weswigham and @andy-ms any more comments?

src/index.ts
andy-ms: @mhegazy missed one

src/index.ts
andy-ms: Could still apply the above refactoring though?

kenchris:
Btw, just wanted to point out (not sure if relevant) that WebIDL is seeing many changes lately after being abandoned for quite some years. This happens on the editors draft though and it not in stable yet:

https://heycam.github.io/webidl/

package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "TSJS-lib-generator",
```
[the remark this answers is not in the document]
mhegazy: done.

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+
+function isEventHandler(p: Browser.Property) {
+    return p.type === "EventHandlerNonNull" || p.type === "EventHandler";
+}
+
+export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
+    // Global print target
+    const printer = createTextWriter("\n");
+
+    const pollutor = getElements(webidl.interfaces, "interface").find(i => flavor === Flavor.Web ? !!i["primary-global"] : !!i.global);
```
andy-ms: Why is this called `pollutor`? Looks like the global interface?

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+            return pollutor ? `this: ${pollutor.name}, ` : "";
+        }
+    }
+
+    // A covariant  EventHandler is one that is defined in a parent interface as then redefined in current interface with a more specific argument types
+    // These patterns are unsafe, and flagged as error under --strictFunctionTypes.
+    // Here we know the property is already defined on the interface, we elide its declaration if the parent has the same handler defined
+    function isCovariantEventHandler(i: Browser.Interface, p: Browser.Property) {
+        return isEventHandler(p) &&
+            iNameToEhParents[i.name] && iNameToEhParents[i.name].length > 0 &&
```
andy-ms: Nit: could use local variables for `iNameToEhParents[i.name]` and `iNameToEhList[i.name]`. Also avoid shadowing `i` so close to the first definition.

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+                }
+            }
+            const requiredModifier = !p.required || p.required === "1" ? "" : "?";
+            pType = p.nullable ? makeNullable(pType) : pType;
+            const readOnlyModifier = p["read-only"] && prefix === "" ? "readonly " : "";
+            printLine(`${prefix}${readOnlyModifier}${p.name}${requiredModifier}: ${pType};`);
+        }
+    }
+
+    function emitComments(entity: { comment?: string; deprecated?: 1 } | undefined, print: (s: string) => void) {
```
andy-ms: Don't think it's helpful to handle undefined input here -- in two of three callsites the input is definitely defined, in the other we're  calling `emitComments` immediately before an `if` and could just move it into the `if`.

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+            if (entity.deprecated) {
+                print(`/** @deprecated */`);
+            }
+        }
+    }
+
+    function emitProperties(prefix: string, emitScope: EmitScope, i: Browser.Interface, conflictedMembers: Set<string>) {
+        // Note: the schema file shows the property doesn't have "static" attribute,
+        // therefore all properties are emited for the instance type.
+        if (emitScope !== EmitScope.StaticOnly) {
```
andy-ms: Nit: double `if` could be replaced with `&&`

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+                method["additional-signatures"]!.forEach(s => printLine(`${prefix}${s};`));
+            }
+            method.signature.forEach(sig => emitSignature(sig, prefix, name, printLine));
+        }
+    }
+
+    function emitMethods(prefix: string, emitScope: EmitScope, i: Browser.Interface, conflictedMembers: Set<string>) {
+        // If prefix is not empty, then this is the global declare function addEventListener, we want to override this
+        // Otherwise, this is EventTarget.addEventListener, we want to keep that.
+        function mFilter(m: Browser.Method) {
```
andy-ms: Would inline this fn

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+
+        // The window interface inherited some methods from "Object",
+        // which need to explicitly exposed
+        if (i.name === "Window" && prefix === "declare function ") {
+            printer.printLine("declare function toString(): string;");
+        }
+    }
+
+    /// Emit the properties and methods of a given interface
+    function emitMembers(prefix: string, emitScope: EmitScope, i: Browser.Interface) {
```
andy-ms: Would move `prefix` to the end and make it optional.
Also, the only `prefix` value currently (besides `""`) is `"declare var"`, so maybe the parameter should be just `inNamespace = false`.

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+
+        for (const relatedIName of iNameToIDependList[i.name]) {
+            const i = allInterfacesMap[relatedIName];
+            if (i) {
+                emitAllMembers(i);
+            }
+        }
+    }
+
+    function emitEventHandlers(prefix: string, i: Browser.Interface) {
```
andy-ms:
```ts
function emitEventHandlers(i: Browser.Interface, inNamespace = false) {
    for (const addOrRemove of ["add", "remove"]) {
        const optionsType = addOrRemove === "add" ? "AddEventListenerOptions" : "EventListenerOptions";
        const fPrefix = inNamespace ? "declare function " : "";
        if (tryEmitTypedEventHandlerForInterface()) {
            // only emit the string event handler if we just emited a typed handler
            printer.printLine(`${fPrefix}${addOrRemove}EventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | ${optionsType}): void;`);
        }

        function emitTypedEventHandler(iParent: Browser.Interface) {
            printer.printLine(`${fPrefix}${addOrRemove}EventListener<K extends keyof ${iParent.name}EventMap>(type: K, listener: (this: ${i.name}, ev: ${iParent.name}EventMap[K]) => any, options?: boolean | ${optionsType}): void;`);
        }

        function tryEmitTypedEventHandlerForInterface() {
            if (iNameToEhList[i.name] && iNameToEhList[i.name].length) {
                emitTypedEventHandler(i);
                return true;
            }
            if (iNameToEhParents[i.name] && iNameToEhParents[i.name].length) {
                iNameToEhParents[i.name]
                    .sort(compareName)
                    .forEach(emitTypedEventHandler);
                return true;
            }
            return false;
        }
    }
}
```

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+                emitStringEventHandler(addOrRemove);
+            }
+        }
+
+        emitEventHandler("add");
+        emitEventHandler("remove");
+    }
+
+    function emitConstructorSignature(i: Browser.Interface) {
+        const constructor = typeof i.constructor === "object" ? i.constructor : undefined;
```
andy-ms: Was there some bad input this is avoiding?

mhegazy: no this is cause all objects have a `constructor` property on them :D

andy-ms: Oh yeah :man_facepalming:

src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    if (webidl["callback-interfaces"]) worker["callback-interfaces"]!.interface = filterProperties(webidl["callback-interfaces"]!.interface, isKnownWorkerName);
+    if (webidl.dictionaries) worker.dictionaries!.dictionary = filterProperties(webidl.dictionaries.dictionary, isKnownWorkerName);
+    if (webidl.enums) worker.enums!.enum = filterProperties(webidl.enums.enum, isKnownWorkerName);
+    if (webidl.mixins) worker.mixins!.mixin = filterProperties(webidl.mixins.mixin, isKnownWorkerName);
+    if (webidl.interfaces) worker.interfaces!.interface = filterProperties(webidl.interfaces.interface, isKnownWorkerName);
+    if (webidl.typedefs) worker.typedefs!.typedef = webidl.typedefs.typedef.filter(t => knownWorkerTypes.has(t["new-type"]));
+
+    const result = emitWebIDl(worker, Flavor.Worker);
+    fs.writeFileSync(tsWorkerOutput, result);
+    return;
```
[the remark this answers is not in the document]
mhegazy: it is intentinal to get an error if i have any code mixed with the helpers.

src/index.ts
```diff
@@ -0,0 +1,162 @@
+import * as Browser from "./types";
+import * as fs from "fs";
+import * as path from "path";
+import { filter, merge, filterProperties } from "./helpers";
+import { Flavor, emitWebIDl } from "./emitter";
+
+function emitDomWorker(webidl: Browser.WebIdl, knownWorkerTypes: Set<string>, tsWorkerOutput: string) {
+    const worker: Browser.WebIdl = {
```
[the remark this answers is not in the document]
mhegazy: it is the filtered webidl file to represent the webworker APIs, simiar to the other one, `browser`.

src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+
+function emitDomWeb(webidl: Browser.WebIdl, tsWebOutput: string) {
+    const browser = filter(webidl, o => {
+        return !(o && typeof o.exposed === "string"
+            && o.exposed.includes("Worker") && !o.exposed.includes("Window"));
+    });
+
+    const result = emitWebIDl(browser, Flavor.Web);
+    fs.writeFileSync(tsWebOutput, result);
+    return;
```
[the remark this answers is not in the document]
mhegazy: same as above

src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+                "mixin": {}
+            },
+            "typedefs": {
+                "typedef": []
+            }
+        };
+
+        if (obj["callback-functions"]) result["callback-functions"]!["callback-function"] = filterProperties(obj["callback-functions"]!["callback-function"], (cb) => !(template["callback-functions"] && template["callback-functions"]!["callback-function"][cb.name]));
+        if (obj["callback-interfaces"]) result["callback-interfaces"]!.interface = filterInterface(obj["callback-interfaces"]!.interface, template["callback-interfaces"] && template["callback-interfaces"]!.interface);
+        if (obj.dictionaries) result.dictionaries!.dictionary = filterDictinary(obj.dictionaries.dictionary, template.dictionaries && template.dictionaries.dictionary);
```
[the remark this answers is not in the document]
mhegazy: I need to assign it something, so having them in the functions seems easier to read.

src/emitter.ts
```diff
@@ -0,0 +1,1126 @@
...
+        }
+    }
+
+    reset();
+
+    return {
+
+        reset: reset,
+
+        resetIndent() { indent = 0; },
```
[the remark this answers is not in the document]
mhegazy: not sure really why we have that. removed it.

src/emitter.ts
mhegazy: the spec refers to it as such. it is the object that "pollutes" the global namespace with its properties.

src/emitter.ts
mhegazy: yeah. i need to do this for all prefixes.. but will do this separately. no need to pass a string in.

src/emitter.ts
mhegazy: not sure how i feel about functions declared in a for loop. will refactor though.

timocov:
@mhegazy seems that this PR change the type of `touch*` events (`touchstart` for example) from `TouchEvent` to just `Event`.

See Microsoft/TypeScript#22565

mhegazy:
PRs welcomed.

PR #385, opened 2018-03-05
Improve type signature for Node#cloneNode

djcsdy:
The attached pull request improves the type signature for `Node#cloneNode`.

```
    cloneNode(deep?: boolean): Node;
```

becomes

```
    cloneNode(deep?: boolean): this;
```

The cloned Node will always have the same type as the original Node.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -8710,7 +8710,7 @@ interface Node extends EventTarget {
     readonly previousSibling: Node | null;
     textContent: string | null;
     appendChild<T extends Node>(newChild: T): T;
-    cloneNode(deep?: boolean): Node;
+    cloneNode(deep?: boolean): this;
     compareDocumentPosition(other: Node): number;
     contains(child: Node): boolean;
     hasAttributes(): boolean;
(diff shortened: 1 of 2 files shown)


baselines/dom.generated.d.ts
```diff
@@ -8710,7 +8710,7 @@ interface Node extends EventTarget {
     readonly previousSibling: Node | null;
     textContent: string | null;
     appendChild<T extends Node>(newChild: T): T;
-    cloneNode(deep?: boolean): Node;
+    cloneNode(deep?: boolean): this;
```
mhegazy: We have explicitly avoided adding `this` types in the library to avoid making all these interfaces generic. generic types have higher cost, and the library has quite a lot of these, and making the base type generic, makes them all generic.

djcsdy: I guess by “higher cost” you mean that every TypeScript program would take a lot longer to compile with this change.

I did not realise this. Fair enough! Sorry to have bothered you :-).

bfred-it: I'm new to types. They may become generic and they'd be more _right,_ what's the point of a typed language if its types are "loose"? Every time I have to use `as` I feel like the types aren't right.

PR #390, opened 2018-03-12
Use `extends Typed` for the appropriate types from type.d.ts

saschanaz:
I've found creating some `WebIdl` types by TypeScript can be painful because of the added `"override-type"?: string;` on `Typed`.

```ts
import * as webidl2 from "webidl2"
import * as Browser from "./types";

function convertIdlType(t: webidl2.IDLTypeDescription): Browser.Typed { /* ... */ }

function convertParam(t: webidl2.Argument): Browser.Param {
  return {
    /*
     * ... some Param specific properties ...
     */
    ...convertIdlType(t.idlType) // Error: override-type does not exist in Browser.Param
  }
}
```

Those types really just extend Typed, so this PR uses `extends Typed` syntax to allow the destructuring work.

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -152,7 +152,7 @@ export interface Interface {
         property: Record<string, Property>;
     }
     constructor?: Constructor;
-    "secure-context"?: string;
+    "secure-context"?: 1;
     implements?: string[];
     static?: 1;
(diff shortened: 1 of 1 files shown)


mhegazy:
thanks!

PR #393, opened 2018-03-25
Add HTMLImageElement.decoding

### LINKED ISSUE #392: Add the decoding property to Image
`Image.decoding` isn't present on the DOM typings.

I can't find the spec, but it's mentioned on different places :
- [WHATWG PR](https://github.com/whatwg/html/pull/3221)
- [on MDN, but you have to scroll or search for "decoding"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)
- [Chromium's Intent to Ship](https://groups.google.com/a/chromium.org/forum/#!topic/Blink-dev/MbXp16hQclY) (they say Edge supports it, isn't TypeScript lib from Edge's IDL?)
- [Chromium IDL](https://chromium.googlesource.com/chromium/src/+/master/third_party/WebKit/Source/core/html/HTMLImageElement.idl#43)

fathyb:
[WHATWG/attr-img-decoding](https://html.spec.whatwg.org/#attr-img-decoding)

> The decoding attribute indicates the preferred method to decode this image. The attribute, if present, must be an image decoding hint. This attribute's missing value default and invalid value default are both the auto state.

Closes #392

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -6132,6 +6132,7 @@ interface HTMLImageElement extends HTMLElement {
     readonly complete: boolean;
     crossOrigin: string | null;
     readonly currentSrc: string;
+    decoding: "async" | "sync" | "auto";
     /**
      * Sets or retrieves the height of the object.
      */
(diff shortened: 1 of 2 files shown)


mhegazy:
thanks!

PR #394, opened 2018-03-26
Adding touch event typings

jakearchibald:
I'm not entirely sure what I'm doing, but if I've done this right it fixes https://github.com/Microsoft/TypeScript/issues/22565 - a regression created in https://github.com/Microsoft/TSJS-lib-generator/pull/383#issuecomment-376132738.


mhegazy:
you also need to include the new baselines. just run `npm run baseline-accept` after `npm run build`

jakearchibald:
Done. I also fixed the baseline-accept script, which didn't do anything (on OSX at least).

mhegazy:
thanks!

PR #398, opened 2018-03-29
Add type to FileReader event target.

rch850:
This PR fixes https://github.com/Microsoft/TypeScript/issues/4163.

The [FileReader API spec](https://www.w3.org/TR/FileAPI/#events) says the event target is `FileReader` so I give `FileReader` type to `target` of `ev`.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4756,11 +4756,6 @@ interface FileReaderProgressEvent extends ProgressEvent {
     readonly target: FileReader | null;
 }
 
-declare var FileReaderProgressEvent: {
-    prototype: FileReaderProgressEvent;
-    new(): FileReaderProgressEvent;
-};
-
 interface FocusEvent extends UIEvent {
     readonly relatedTarget: EventTarget;
     initFocusEvent(typeArg: string, canBubbleArg: boolean, cancelableArg: boolean, viewArg: Window, detailArg: number, relatedTargetArg: EventTarget): void;
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4722,12 +4722,12 @@ interface FileReaderEventMap {
 
 interface FileReader extends EventTarget {
     readonly error: DOMException | null;
-    onabort: ((this: FileReader, ev: ProgressEvent) => any) | null;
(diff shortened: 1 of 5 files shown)


baselines/dom.generated.d.ts
```diff
@@ -4722,12 +4722,12 @@ interface FileReaderEventMap {
 
 interface FileReader extends EventTarget {
     readonly error: DOMException | null;
-    onabort: ((this: FileReader, ev: ProgressEvent) => any) | null;
-    onerror: ((this: FileReader, ev: ProgressEvent) => any) | null;
-    onload: ((this: FileReader, ev: ProgressEvent) => any) | null;
-    onloadend: ((this: FileReader, ev: ProgressEvent) => any) | null;
-    onloadstart: ((this: FileReader, ev: ProgressEvent) => any) | null;
-    onprogress: ((this: FileReader, ev: ProgressEvent) => any) | null;
+    onabort: ((this: FileReader, ev: ProgressEvent & {target: FileReader}) => any) | null;
```
mhegazy: i would create a new type `FilreReaderProgressEvent` that extends `ProgressEvent` and use that instead.

rch850: Thank you for reviewing. I have added `FileReaderProgressEvent`.

baselines/dom.generated.d.ts
```diff
@@ -4752,6 +4752,15 @@ declare var FileReader: {
     readonly LOADING: number;
 };
 
+interface FileReaderProgressEvent extends ProgressEvent {
+    readonly target: FileReader | null;
+}
+
+declare var FileReaderProgressEvent: {
```
mhegazy: this should not be added. set `"no-interface-object": 1,` on the new interface to avoid the creation of a var.

inputfiles/addedTypes.json
```diff
@@ -47,6 +47,19 @@
...
+                "extends": "ProgressEvent",
+                "properties": {
+                    "property": {
+                        "name": {
+                            "name": "target",
+                            "read-only": 1,
+                            "override-type": "FileReader | null"
+                        }
+                    }
+                }
```
mhegazy: add `"no-interface-object": 1,`

rch850:
Removed unnecessary `var FileReaderProgressEvent` and made test passed.

mhegazy:
Thanks

PR #405, opened 2018-04-04
Add Web IDL base types + Encoding Standard from their specs

saschanaz:
The new json type file tends to flatten typedefs, is it a requirement?


inputfiles/idl/Encoding.widl
```diff
@@ -0,0 +1,23 @@
+dictionary TextDecoderOptions {
```
mhegazy: where is the idl file comming from?

mhegazy: nevermind. found the link :)

src/fetcher.ts
```diff
@@ -0,0 +1,35 @@
...
+
+fetchIDLs();
+
+interface IDLSource {
+    url: string;
+    title: string;
+}
+
+async function fetchIDLs() {
+    const file = fs.readFileSync(`${__dirname}/../inputfiles/idlSources.json`, { encoding: "utf-8" });
```
mhegazy: nit: i would use `path.combine` here.

mhegazy: or consider using `require` to get it and avoid `JSON.parse`

src/fetcher.ts
```diff
@@ -0,0 +1,35 @@
...
+    for (const source of idlSources) {
+        const idl = await fetchIDL(source);
+        fs.writeFileSync(`${__dirname}/../inputfiles/idl/${source.title}.widl`, idl + '\n');
+    }
+}
+
+async function fetchIDL(source: IDLSource) {
+    const response = await fetch(source.url);
+    const dom = new JSDOM(await response.text());
+    const elements = Array.from(dom.window.document.querySelectorAll("pre.idl:not(.extract),code.idl-code"));
```
mhegazy: totally random, and not related to this change; it would be super cool if in the future we can extract the description from the widl file and put it as a comment on the declaration..

src/widlprocess.ts
```diff
@@ -0,0 +1,240 @@
...
+    if (i.union) {
+        return {
+            type: (i.idlType as webidl2.IDLTypeDescription[]).map(convertIdlType),
+            nullable: i.nullable ? 1 : undefined
+        };
+    }
+    throw new Error("Unsupported IDL type structure");
+}
+
+function createEmptyBrowserWebidl(): Browser.WebIdl {
```
mhegazy: we have this function already in index.ts, i would move it helpers and reuse it.

src/widlprocess.ts
```diff
@@ -0,0 +1,240 @@
+import * as webidl2 from "webidl2";
+import * as Browser from "./types";
+
+export function convert(text: string) {
+    const rootTypes = webidl2.parse(text);
+    const partialInterfaces: Browser.Interface[] = [];
+    const browser = createEmptyBrowserWebidl();
```
mhegazy: nit, `browser` might not be the best name, consider changing it to `webidl`

src/widlprocess.ts
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+    if (constructor.signature.length) {
+        return constructor;
+    }
+}
+
+function convertOperation(operation: webidl2.OperationMemberType): Browser.Method {
+    return {
+        name: operation.name!,
```
mhegazy: can it be `undefinded`? what happens then? should we throw?

src/widlprocess.ts
```diff
@@ -0,0 +1,240 @@
...
+        signature: [{
+            ...convertIdlType(operation.idlType!),
```
mhegazy: should we throe if idlType is not defined?

src/test.ts
```diff
@@ -26,7 +26,7 @@ function compareToBaselines() {
 
 function compileGeneratedFile(file: string) {
     try {
-        child_process.execSync(`node ${tscPath} --strict --lib es5 --noEmit ${path.join(outputFolder, file)}`);
+        child_process.execSync(`node ${tscPath} --strict --lib es5 --skipLibCheck --noEmit ${path.join(outputFolder, file)}`);
```
mhegazy: why? we want to make sure we are generating valid declaration file.. that is the whole point of the test.

saschanaz: This breaks `@types/jsdom` which depends on DOM types. I think we are still testing the input library file, this just turns off checks for JavaScript types.

mhegazy: Then Add ‘—types’ to exclude all @types packages.

saschanaz: Good idea, but how can I pass empty string to `--types`? `--types []`, `--types ""`, or `--types "''"` don't work.

mhegazy: just `--types`.. 

`--strict --lib es5 --types --noEmit`

mhegazy:
thanks!

PR #408, opened 2018-04-05
Change setValueCurveAtTime values to Float32Array

iccir:
Microsoft/TypeScript#23160

https://www.w3.org/TR/webaudio/#widl-AudioParam-setValueCurveAtTime-AudioParam-Float32Array-values-double-startTime-double-duration


mhegazy:
Pushed an alternative fix in https://github.com/Microsoft/TSJS-lib-generator/pull/409

mhegazy:
closing in favor of https://github.com/Microsoft/TSJS-lib-generator/pull/409

PR #410, opened 2018-04-05
Add missing methods to URLSearchParams

philnova:
Add typings for methods that are specified in the URLSearchParams API but are missing in the Typescript def.


mhegazy:
These are already included, see https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.iterable.d.ts#L118-L135.

Not sure what i am missing here?

mhegazy:
Please see my response in https://github.com/Microsoft/TypeScript/issues/23174

PR #412, opened 2018-04-05
Add support for AES keys in importKey

ColtonProvias:
Fixes key import to be more consistent with the WebCrypto API as specified at https://www.w3.org/TR/WebCryptoAPI/ under sections 25.7 (AES-CTR Operations), 26.4 (AES-CBC Operations), 27.4 (AES-GCM Operations), and 28.3 (AES-KW Operations).


mhegazy:
thanks

PR #422, opened 2018-04-09
 Extract comments from html files along with widl when possible

mhegazy:
Some of the input files have a non-normative descriptions of the interface meant for end-user use. This change extracts these comments and ads them to the generated files.


mhegazy:

tsconfig.json
```diff
@@ -2,7 +2,7 @@
     "compilerOptions": {
         "target": "es6",
```
saschanaz: Will there is a trouble if we just set target: es2016?

mhegazy:
Looks like not all of the widl files have the `domintro` section. i will need to enhance this logic a bit.

saschanaz:
Is there a blocking problem here?

mhegazy:
Some legal issues I need to sort out. Need to figure out if the license allows such use and what attribution is needed.

PR #424, opened 2018-04-10
Add RTCDataChannel interface

a-tarasyuk:
[TypeScript#23300](https://github.com/Microsoft/TypeScript/issues/23300)

[webrtc#dom-rtcdatachannel](https://www.w3.org/TR/webrtc/#dom-rtcdatachannel)

### DIFF
--- a/inputfiles/idl/WebRTC.widl
+++ b/inputfiles/idl/WebRTC.widl
@@ -1,0 +1,677 @@
+dictionary RTCConfiguration {
+             sequence<RTCIceServer>   iceServers;
+             RTCIceTransportPolicy    iceTransportPolicy = "all";
+             RTCBundlePolicy          bundlePolicy = "balanced";
+             RTCRtcpMuxPolicy         rtcpMuxPolicy = "require";
+             DOMString                peerIdentity;
+             sequence<RTCCertificate> certificates;
+             [EnforceRange]
+             octet                    iceCandidatePoolSize = 0;
+};
+
+enum RTCIceCredentialType {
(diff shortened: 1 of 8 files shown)


inputfiles/idl/WebRTC.widl
```diff
@@ -0,0 +1,39 @@
+enum RTCPriorityType {
```
mhegazy: this file should be added to https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/idlSources.json, so that next fetch would bring it too.

mhegazy:
@saschanaz can you please review as well.

inputfiles/idl/WebRTC.widl
a-tarasyuk: ok

a-tarasyuk:
@mhegazy I've checked `widl` is not stable for now, and generated type is not correct. I decided to define missing type in `addedTypes.json`.

mhegazy:
We are trying to move away from adding types directly, and instead use the widl. why is not that stable? can we add the widl, then update specific declarations/properties?

mhegazy:
@saschanaz can you review the latest version of the change?

saschanaz:
The code has a bug about partial dictionaries, I will open a separate PR for this.

a-tarasyuk:
?

saschanaz:
#428 is merged, please rebuild. Thanks! 👍

mhegazy:
thanks @saschanaz for the help!

PR #427, opened 2018-04-10
Add IDBValidKey to knownWorkerTypes

### LINKED ISSUE #426: IDBValidKey does not exist in lib.webworker.d.ts
This type exists in DOM types, but does not exist in Web Worker types. It should exist in both, as IndexedDB spec and implementation is the same for both.

It is defined [here](https://github.com/Microsoft/TSJS-lib-generator/blob/818f7457e6bffe0bd8255f4e00f1b1dd700cd425/inputfiles/addedTypes.json#L2876).

mmermerkaya:
Fixes #426

### DIFF
--- a/baselines/webworker.generated.d.ts
+++ b/baselines/webworker.generated.d.ts
@@ -1774,6 +1774,7 @@ type RequestInfo = Request | string;
 type PerformanceEntryList = PerformanceEntry[];
 type BufferSource = ArrayBufferView | ArrayBuffer;
 type FormDataEntryValue = string | File;
+type IDBValidKey = number | string | Date | IDBArrayKey;
 type AlgorithmIdentifier = string | Algorithm;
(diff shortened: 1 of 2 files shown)


mhegazy:
you also need to update the baselines. 
```sh
npm run build
npm run baseline-accept
```

mhegazy:
Thanks @mmermerkaya, we should be good to go, once we have the contributor licence agreement signed.

mmermerkaya:
Thanks :) I'll do that as soon as I get approval from my employer (@google).

mmermerkaya:
Yep, I've already emailed the relevant people. I'll ping this PR as soon as that's done.

saschanaz:
IDBValidKey is not in the spec anymore, and the current dom.generated.d.ts does not have it. I think we should remove it completely.

PR #432, opened 2018-04-12
Add XMLHttpRequest types

saschanaz:
Add `FormData.forEach`, remove obsolete `initProgressEvent`, fix `XMLHttpRequestEventTargetEventMap`, etc.


baselines/webworker.generated.d.ts
```diff
@@ -641,18 +641,20 @@ declare var FileReaderSync: {
 };
 
 interface FormData {
-    append(name: string, value: string | Blob, fileName?: string): void;
+    append(name: string, value: string): void;
+    append(name: string, blobValue: Blob, filename?: string): void;
```
mhegazy: the previous overload was better.. since you can pass a value of type `string | Blob`

baselines/webworker.generated.d.ts
```diff
@@ -641,18 +641,20 @@ declare var FileReaderSync: {
...
     delete(name: string): void;
     get(name: string): FormDataEntryValue | null;
     getAll(name: string): FormDataEntryValue[];
     has(name: string): boolean;
-    set(name: string, value: string | Blob, fileName?: string): void;
```
mhegazy: same here.

baselines/webworker.generated.d.ts
```diff
@@ -1663,30 +1663,35 @@ declare var XMLHttpRequest: {
...
+    "load": ProgressEvent;
     "loadend": ProgressEvent;
-    "loadstart": Event;
+    "loadstart": ProgressEvent;
     "progress": ProgressEvent;
     "timeout": ProgressEvent;
 }
 
-interface XMLHttpRequestEventTarget {
-    onabort: ((this: XMLHttpRequest, ev: Event) => any) | null;
```
mhegazy: the `this` type was intentional see https://github.com/Microsoft/TypeScript/issues/15101 and https://github.com/Microsoft/TSJS-lib-generator/pull/275

src/index.ts
```diff
@@ -111,7 +111,7 @@ function emitDom() {
         if (obj["callback-interfaces"]) result["callback-interfaces"]!.interface = filterInterface(obj["callback-interfaces"]!.interface, template["callback-interfaces"] && template["callback-interfaces"]!.interface);
         if (obj.dictionaries) result.dictionaries!.dictionary = filterDictionary(obj.dictionaries.dictionary, template.dictionaries && template.dictionaries.dictionary);
         if (obj.enums) result.enums!.enum = filterEnum(obj.enums.enum, template.enums && template.enums.enum);
-        if (obj.mixins) result.mixins!.mixin = filterInterface(obj.mixins.mixin, template.mixins && template.mixins.mixin);
+        if (obj.mixins) result.mixins!.mixin = filterProperties(obj.mixins.mixin, mixin => !(template.mixins && template.mixins!.mixin[mixin.name]));
```
mhegazy: not sure i understand this change?

baselines/webworker.generated.d.ts
saschanaz: I agree, but somehow the spec authors have decided to use two signatures. No idea why, maybe documentation purpose?

```
[Constructor(optional HTMLFormElement form),
 Exposed=(Window,Worker)]
interface FormData {
  void append(USVString name, USVString value);
  void append(USVString name, Blob blobValue, optional USVString filename);
  void delete(USVString name);
  FormDataEntryValue? get(USVString name);
  sequence<FormDataEntryValue> getAll(USVString name);
  boolean has(USVString name);
  void set(USVString name, USVString value);
  void set(USVString name, Blob blobValue, optional USVString filename);
  iterable<USVString, FormDataEntryValue>;
};
```

mhegazy: maybe we can override these two methods to avoid breaking existing users.

baselines/webworker.generated.d.ts
saschanaz: `XMLHttpRequestUpload` also extends `XMLHttpRequestEventTarget`, hmm. It doesn't have any properties so forcing XMLHttpRequest won't be an issue for now...

saschanaz: Done

baselines/webworker.generated.d.ts
saschanaz: Ah, I just noted this is a breaking change, but anyway `formData.append('str', 'str', 'str);` is invalid and throws on Firefox. I would say the spec signature is better to prevent unexpected runtime breakage.

saschanaz: But the union problem... looks like there is no ideal solution here, I'll just override them to keep previous behavior.

mhegazy: i agree.. but `append(string, sring|Blob)` should just work fine.

src/index.ts
saschanaz: Previous code removed mixin members, the changed one removes the mixin itself.

saschanaz: (XMLHttpRequestEventTarget was a mixin but now is an interface, so the previous one should be removed.)

saschanaz: It's really a copy of line 110.

```ts
if (obj["callback-functions"]) result["callback-functions"]!["callback-function"] = filterProperties(obj["callback-functions"]!["callback-function"], (cb) => !(template["callback-functions"] && template["callback-functions"]!["callback-function"][cb.name]));
```

PR #438, opened 2018-04-13
Add notifications/vibration types

### LINKED ISSUE #354: Readonly property `permission` is missing in `Notification` global variable
Hello.

Issue title pretty much describes the problem. Readonly property `permission` with status of permission for web notifications is [not emitted](https://github.com/Microsoft/TSJS-lib-generator/blob/master/baselines/webworker.generated.d.ts#L776) in global variable, but present in [interface](https://github.com/Microsoft/TSJS-lib-generator/blob/master/baselines/webworker.generated.d.ts#L766) declaration.

Unfortunately, I don't know F# at all... I've tried to fix this using `addedTypes.json`, but when I add this entry, readonly property `permission` gets emitted in every single one interface and global variable declaration. And in the docs there is note stating that modification of `browser.webidl.xml` is forbidden, because it's generated by Microsoft Edge.
```json
{
	"kind": "constant",
	"name": "permission",
	"type": "NotificationPermission"
}
```
I can make a PR showing my problem, should I?

saschanaz:
Adds some missing properties on `Notification`.

Fixes #354, fixes Microsoft/TypeScript#14701


mmermerkaya:
When can we expect this change to be released? It's not in TS 2.9.

mhegazy:
TS 3.0

mmermerkaya:
Thank you! :slightly_smiling_face:

PR #445, opened 2018-04-19
Support LegacyWindowAlias

saschanaz:
[Geometry Interfaces Module](https://drafts.fxtf.org/geometry/) (#421) defines older SVGMatrix, WebKitCSSMatrix, etc. as legacy aliases of DOM-prefixed interfaces by `[LegacyWindowAlias]`.

This PR reads it and emits corresponding aliases, and then removes older type definitions.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3511,6 +3511,12 @@ declare var DOMMatrix: {
     fromMatrix(other?: DOMMatrixInit): DOMMatrix;
 };
 
+type SVGMatrix = DOMMatrix;
+declare var SVGMatrix: typeof DOMMatrix;
+
+type WebKitCSSMatrix = DOMMatrix;
+declare var WebKitCSSMatrix: typeof DOMMatrix;
+
 interface DOMMatrixReadOnly {
     readonly a: number;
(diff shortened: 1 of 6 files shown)


baselines/dom.generated.d.ts
```diff
@@ -12318,31 +12330,6 @@ declare var SVGMaskElement: {
...
-    e: number;
-    f: number;
-    flipX(): SVGMatrix;
-    flipY(): SVGMatrix;
-    inverse(): SVGMatrix;
-    multiply(secondMatrix: SVGMatrix): SVGMatrix;
-    rotate(angle: number): SVGMatrix;
-    rotateFromVector(x: number, y: number): SVGMatrix;
-    scale(scaleFactor: number): SVGMatrix;
-    scaleNonUniform(scaleFactorX: number, scaleFactorY: number): SVGMatrix;
```
mhegazy: this does not seem o be defined on DOMMatrix

saschanaz: scaleNonUniform was there but has been removed by https://github.com/w3c/fxtf-drafts/commit/99e3212469026b2f2f50926a41912d110a1741b7, so this is expected.

PR #449, opened 2018-04-21
Autoexpose types by [Exposed]

### LINKED ISSUE #431: Micro-control member exposure
For example, `performance` is exposed for both `Window` and `Worker` but `performance.timing` and `performance.navigation` is exposed only for `Window`.

```webidl
[Exposed=Window]
partial interface Performance {
  [SameObject]
  readonly attribute PerformanceTiming timing;
  [SameObject]
  readonly attribute PerformanceNavigation navigation;
};
```

saschanaz:
This reads [Exposed] and use it to determine whether this type should be written to the target lib file, so that we won't forget to expose new types on webworker.generated.d.ts.

Fixes #431


mhegazy:
some merge conflicts need attention.

PR #452, opened 2018-04-24
Add Console types

saschanaz:
This introduces namespace types.

This removes:

* ~~`.memory`~~ (It exists on Chrome so keeping for compatibility)
* ~~`.exception()`~~ (It exists on Firefox so keeping for compatibility)
* `.markTimeline()`
* `.profile()` (It exists on Firefox and Chrome but is non-functional https://github.com/whatwg/console/issues/144#issuecomment-432184849)
* `.profileEnd()`
* ~~`.timeStamp()`~~ (Not specced yet but is widely implemented https://github.com/whatwg/console/issues/140)
* `.timeline()`
* `.timelineEnd()`

and adds:

* `.countReset()`
* `.timeLog()`

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -19094,7 +19094,7 @@ declare namespace console {
     function count(label?: string): void;
     function countReset(label?: string): void;
     function debug(...data: any[]): void;
-    function dir(item: any, options?: any): void;
+    function dir(item?: any, options?: any): void;
     function dirxml(...data: any[]): void;
     function error(...data: any[]): void;
     function exception(message?: string, ...optionalParams: any[]): void;
@@ -19103,7 +19103,7 @@ declare namespace console {
     function groupEnd(): void;
     function info(...data: any[]): void;
     function log(...data: any[]): void;
-    function table(tabularData: any, properties?: string[]): void;
+    function table(tabularData?: any, properties?: string[]): void;
     function time(label?: string): void;
     function timeEnd(label?: string): void;
     function timeLog(label?: string, ...data: any[]): void;
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3723,40 +3723,6 @@ interface ConcatParams extends Algorithm {
     publicInfo?: Uint8Array;
 }
 
-/** Provides access to the browser's debugging console (e.g. the Web Console in Firefox). The specifics of how it works varies from browser to browser, but there is a de facto set of features that are typically provided. */
-interface Console {
-    memory: any;
-    assert(condition?: boolean, message?: string, ...data: any[]): void;
-    clear(): void;
-    count(label?: string): void;
-    debug(message?: any, ...optionalParams: any[]): void;
-    dir(value?: any, ...optionalParams: any[]): void;
-    dirxml(value: any): void;
-    error(message?: any, ...optionalParams: any[]): void;
-    exception(message?: string, ...optionalParams: any[]): void;
-    group(groupTitle?: string, ...optionalParams: any[]): void;
-    groupCollapsed(groupTitle?: string, ...optionalParams: any[]): void;
-    groupEnd(): void;
-    info(message?: any, ...optionalParams: any[]): void;
-    log(message?: any, ...optionalParams: any[]): void;
-    markTimeline(label?: string): void;
-    profile(reportName?: string): void;
-    profileEnd(reportName?: string): void;
-    table(...tabularData: any[]): void;
-    time(label?: string): void;
-    timeEnd(label?: string): void;
-    timeStamp(label?: string): void;
-    timeline(label?: string): void;
-    timelineEnd(label?: string): void;
-    trace(message?: any, ...optionalParams: any[]): void;
-    warn(message?: any, ...optionalParams: any[]): void;
-}
-
-declare var Console: {
-    prototype: Console;
-    new(): Console;
-};
-
 interface ConstantSourceNode extends AudioScheduledSourceNode {
     readonly offset: AudioParam;
     addEventListener<K extends keyof AudioScheduledSourceNodeEventMap>(type: K, listener: (this: ConstantSourceNode, ev: AudioScheduledSourceNodeEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
@@ -18590,7 +18556,7 @@ interface WindowEventMap extends GlobalEventHandlersEventMap, WindowEventHandler
 }
 
 /** A window containing a DOM document; the document property points to the DOM document loaded in that window. */
-interface Window extends EventTarget, AnimationFrameProvider, GlobalEventHandlers, WindowConsole, WindowEventHandlers, WindowLocalStorage, WindowOrWorkerGlobalScope, WindowSessionStorage {
+interface Window extends EventTarget, AnimationFrameProvider, GlobalEventHandlers, WindowEventHandlers, WindowLocalStorage, WindowOrWorkerGlobalScope, WindowSessionStorage {
     readonly applicationCache: ApplicationCache;
     readonly clientInformation: Navigator;
     readonly closed: boolean;
@@ -18720,10 +18686,6 @@ declare var Window: {
     prototype: Window;
     new(): Window;
 };
-
-interface WindowConsole {
-    readonly console: Console;
-}
 
 interface WindowEventHandlersEventMap {
     "afterprint": Event;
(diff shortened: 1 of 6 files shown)


baselines/webworker.generated.d.ts
```diff
@@ -2049,6 +2010,26 @@ declare var XMLHttpRequestUpload: {
     new(): XMLHttpRequestUpload;
 };
 
+declare namespace console {
```
mhegazy: this is going to be a breaking change. i will need to run the tests against definitelytyped.

saschanaz: Would `type Console = typeof console;` mitigate the compat problem?

mhegazy: My concern is that augmentation to `Console` are now ignored. making it a type makes it an error. but does not address the issue.

saschanaz: My quick search on DefinitelyTyped (with a search query `interface Console `) says only [arangodb](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/arangodb/index.d.ts) augments the existing `Console` definition, other types including Node.js and react-native have their own ones.

saschanaz: `extends Console` also only shows core-js.

saschanaz:
@sandersn, I just refreshed this PR, could you review this?

baselines/webworker.generated.d.ts
saschanaz: The `arangodb` type is not really about augmenting existing browser console but about typing their own console module: https://github.com/arangodb/arangodb/blob/16a1fabf2057fb1da56a7abd44bd8e27ab1bead5/js/common/bootstrap/modules/console.js

`core-js` dropped their `core.log` that was using `extends Console` 4 years ago. (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36100)

baselines/dom.generated.d.ts
```diff
@@ -3519,40 +3519,6 @@ interface ConcatParams extends Algorithm {
...
-    timeStamp(label?: string): void;
-    timeline(label?: string): void;
-    timelineEnd(label?: string): void;
-    trace(message?: any, ...optionalParams: any[]): void;
-    warn(message?: any, ...optionalParams: any[]): void;
-}
-
-declare var Console: {
-    prototype: Console;
-    new(): Console;
```
sandersn: When was this valid, if ever? (At least, I tested it on V8 and didn't get anything; I don't have Firefox installed.)

baselines/dom.generated.d.ts
```diff
@@ -18878,6 +18840,29 @@ declare var webkitRTCPeerConnection: {
 
 declare type EventListenerOrEventListenerObject = EventListener | EventListenerObject;
 
+declare namespace console {
```
sandersn: I'll re-test @mhegazy's question. Here's how I'll do it in case you want to try it yourself.

```sh
$ git checkout console
$ npm test
$ cp baselines/* ~/ts/src/lib
$ cd ~/ts
$ gulp
$ cd ~/dtslint-runner
$ npm run build
$ node bin/index.js --localTs ~/ts/built/local
```

You need clones of Typescript, dtslint-runner and DefinitelyTyped to make this work.

baselines/dom.generated.d.ts
```diff
@@ -3519,40 +3519,6 @@ interface ConcatParams extends Algorithm {
     publicInfo?: Uint8Array;
 }
 
-/** Provides access to the browser's debugging console (e.g. the Web Console in Firefox). The specifics of how it works varies from browser to browser, but there is a de facto set of features that are typically provided. */
-interface Console {
-    memory: any;
```
sandersn: The following properties are now gone:
* memory
* exception
* [mark]timeline[End]

Do you know who this is likely to break? People with dependencies on old versions of IE? Somebody else?

sandersn:
The following packages had errors: angular-meteor, apollo-upload-client, atom-mocha-test-runner, atom, backlog-js, baconjs, blessed, blob-stream, blob-to-buffer, body-parser, bytebuffer, cacheable-request, carlo, cassanknex, chai-enzyme, chai-webdriverio, cheerio, clean-css, consul, core-js, cron, depd, documentdb, dom-inputevent, dom-to-image, dropboxjs, electron-clipboard-extended, electron-json-storage, electron-notify, electron-notifications, electron-spellchecker, electron-window-state, engine.io-client, dojo, enzyme, enzyme-adapter-react-16, enzyme-adapter-react-15, enzyme-to-json, event-to-promise, evernote, expect-puppeteer, feathersjs__authentication-client, frisby, got, got/v8, gulp, gulp-cheerio, html5-to-pdf, htmlparser2, imperium, jasmine-enzyme, jest-environment-puppeteer, jsdom, jsdom/v2, jszip, kefir, koa-compose, kue, lasso, leadfoot, line-by-line, listr, lobibox, magnet-uri, maker.js, marko, menubar, meteor-collection-hooks, meteor-astronomy, meteor-persistent-session, meteor-jboulhous-dev, meteor-prime8consulting-oauth2, meteor-publish-composite, meteor-roles, meteor, mozilla-readability, multy, n3, multer-s3, next-redux-wrapper, next-redux-saga, nightmare, node/v0, node, node/v6, node/v11, node/v7, node/v10, node/v8, node/v9, node-persist, nodemailer-ses-transport, nookies, nw.gui, nw.js, openfin, openfin/v37, openfin/v39, openfin/v34, openpgp, orchestrator, papaparse, parse, parse/v1, parse-torrent, parse-mockdb, parse-torrent-file, pouch-redux-middleware, pouchdb, pouchdb-adapter-fruitdown, pouchdb-adapter-http, pouchdb-adapter-idb, pouchdb-adapter-leveldb, pouchdb-adapter-localstorage, pouchdb-adapter-node-websql, pouchdb-adapter-memory, pouchdb-browser, pouchdb-adapter-websql, pouchdb-core, pouchdb-find, pouchdb-http, pouchdb-mapreduce, pouchdb-live-find, pouchdb-replication, pouchdb-node, pouchdb-upsert, promised-temp, promisify-supertest, protractor-browser-logs, puppeteer/v0, puppeteer, puppeteer-core, qrcode, rasha, react-bootstrap-table/v2, react-csv, react-hot-loader, react-router, react-timeout, redux-storage, roads, roads-server, sass-webpack-plugin, s3-download-stream, s3-upload-stream, shapefile, sharedb, simple-cw-node, simple-peer, snoowrap, socket.io-parser, sortablejs, spdy, sql.js, statsd-client, sqs-producer, sqs-consumer, stompjs, storybook__addon-a11y, styled-components/v3, stylus, superagent, superagent/v2, superagent-bunyan, superagent-prefix, superagent-no-cache, supertest, supertest-as-promised, tesseract.js, text-buffer, thrift, touch, vast-client, webappsec-credential-management, webgl2, webtorrent, whatwg-streams, wonder.js, xmlpoke, zipkin-context-cls, zipkin-transport-http, zui.

I'm going to look into some of them; a lot seemed quite similar.

sandersn:
The majority of the errors are from `@types/node`, which needs to switch to a namespace as well. After my follow-up run finishes, I'll post the list of `@types` that still fail with some errors.

sandersn:
Here's the errors after fixing node. I didn't fix node/v* so there are errors from there. Besides that, it looks like a few other packages have their own declarations of `console`.

[errors.txt](https://github.com/microsoft/TSJS-lib-generator/files/3278635/errors.txt)

baselines/dom.generated.d.ts
saschanaz: No `Console` public constructor has ever existed, at least since the spec was first published in 2014.

IE11 somehow has it, though.

baselines/dom.generated.d.ts
saschanaz: Seems that I have to keep `.memory` and `.exception` for now because they are in Edge and Firefox although not in spec. `.markTimeline` is only in Edge in non-functional state.

saschanaz:
Some errors are from the removed `window.console`, which can probably be fixed via globalThis.

orta:
👋 - I took a look over this PR, I'd love to see Console changes move to the IDL. 

Now that `globalThis` is available , is it feasible to get this PR to a mergeable state? If yes, we'll try re-run the DT checks and see if node is still broken.

saschanaz:
IMO ready to be reviewed again 👍

sandersn:
DT still has 199 failures. Almost all of them are because node still has a global `declare var console: Console` that will need to become a namespace (and of course add the new properties). I'm not sure how to make that happen without breaking old versions of Typescript though:

```ts
declare var console: Console
// and
declare namespace console {
  export function f();
}
```

Are not going to merge, but that's what you'll have when mixing an old node with a new Typescript, or vice versa.

Judging from the other errors:

* meteor tries to `declare var console`
* jsdom might try to extend the `Console` type.
* zipkin refers to the Console type, but I'm not sure if it tries to extend it.
* External packages `@jest/console` and meteor-typings do the same.

So the next step is to make sure that `@types/node` can be used with these new dom types, and, after being changed, still works with the old ones too. There might be an obvious fix that I'm missing, but I haven't seen it yet.

sandersn:
As-is, I think there are too many breaks to take this. Unfortunately, I think that the IDL-generated namespace is not going to work with @types/node, at least not without some solution I haven’t thought of before.

I’m going to close this and recommend adding the new functions to overridingTypes.json. @saschanaz if you have an idea how to make @types/node compatibility work, let me know and we can restart this effort.

saschanaz:
Well, I don't have any other workaround in mind, so how about we tweak the script a little bit to generate an interface instead, special casing `console`?

sandersn:
I thought about that, but that seems hardly different from entries in overridingTypes.json. And the code to support overridingTypes.json is already written.

PR #454, opened 2018-04-26
Emit iterator for iterable/maplike/setlike

saschanaz:
This emits iterators for interfaces with `iterable<>`, `maplike<>`, `setlike<>` declarations.

### DIFF
--- a/baselines/dom.es6.generated.d.ts
+++ b/baselines/dom.es6.generated.d.ts
@@ -38,12 +38,26 @@ interface FileList {
     [Symbol.iterator](): IterableIterator<File>
 }
 
+interface FormData {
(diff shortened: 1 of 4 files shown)


mhegazy:
Can you also add `NodeListOf`, `HTMLCollectionBase`, `HTMLCollectionOf`, so that we can replace https://github.com/Microsoft/TypeScript/blob/master/src/lib/dom.iterable.d.ts with this file.

mhegazy:
`NodeList.prototype[Symbol.iterator]` exits at runtime, but it does not show on `NodeList`

saschanaz:
CSSRuleList is already there, so I added others.

PR #455, opened 2018-04-26
Feature/add webworkers api widl

a-tarasyuk:
[Microsoft/TypeScript#/23653](https://github.com/Microsoft/TypeScript/issues/23653)
[Web Workers Spec](https://html.spec.whatwg.org/multipage/workers.html)


baselines/dom.generated.d.ts
```diff
@@ -16632,5 +16657,5 @@ type VRDisplayEventReason = "mounted" | "navigation" | "requested" | "unmounted"
 type VREye = "left" | "right";
 type VideoFacingModeEnum = "user" | "environment" | "left" | "right";
 type VisibilityState = "hidden" | "visible" | "prerender" | "unloaded";
-type WorkerType = "classic" | "module";
+type WorkerType = "classic" | "module" | "classic" | "module";
```
mhegazy: this needs to be filtered.

baselines/webworker.generated.d.ts
```diff
@@ -2067,42 +2120,14 @@ interface NotificationPermissionCallback {
...
-declare function msWriteProfilerMark(profilerMarkName: string): void;
-declare function dispatchEvent(evt: Event): boolean;
-declare var indexedDB: IDBFactory;
-declare var msIndexedDB: IDBFactory;
-declare var navigator: WorkerNavigator;
-declare function clearImmediate(handle: number): void;
-declare function clearInterval(handle: number): void;
-declare function clearTimeout(handle: number): void;
-declare function importScripts(...urls: string[]): void;
-declare function setImmediate(handler: any, ...args: any[]): number;
```
mhegazy: why are these removed?

baselines/dom.generated.d.ts
a-tarasyuk: done

baselines/webworker.generated.d.ts
saschanaz: Because widlprocess does not support `[Global]` yet 😅

saschanaz:
I was going to add full HTML spec in once, but being incremental is probably better 😁

baselines/webworker.generated.d.ts
saschanaz: #456 is merged, please rebuild!

saschanaz:
The multipaged worker spec lacks `WindowOrWorkerGlobalScope` which is the target from many specs to add global methods/properties. We may manually add an empty one.

saschanaz:
Oh, this is an oops situation. I expected the empty WindowOrWorkerGlobalScope would automatically get some types from already added specs, but:

1. We don't support `partial interface mixin`s yet
2. Even if we did, specs are incorrectly using it as `partial interface`. I'll file some issue for those specs.

a-tarasyuk:
@saschanaz should I remove it for now?

saschanaz:
It also has IndexedDB and other stable types. Worker will lose access for them without proper WindowOrWorkerGlobalScope definition. I think we should add those lost types manually for now.

a-tarasyuk:
@saschanaz Will add.

inputfiles/knownWorkerTypes.json
```diff
@@ -1,6 +1,126 @@
 [
+    "ClientQueryOptions",
```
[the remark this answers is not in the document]
saschanaz: I meant not only that one but all additions here, because now types are autoexposed.

baselines/webworker.generated.d.ts
```diff
@@ -2055,38 +2078,33 @@ interface EventHandlerNonNull {
...
-declare var indexedDB: IDBFactory;
-declare var msIndexedDB: IDBFactory;
-declare var navigator: WorkerNavigator;
-declare function clearImmediate(handle: number): void;
-declare function clearInterval(handle: number): void;
-declare function clearTimeout(handle: number): void;
 declare function importScripts(...urls: string[]): void;
-declare function setImmediate(handler: any, ...args: any[]): number;
-declare function setInterval(handler: any, timeout?: any, ...args: any[]): number;
-declare function setTimeout(handler: any, timeout?: any, ...args: any[]): number;
```
mhegazy: is that expected? should `setTimeout` not be available for workers?

saschanaz: It's needed, I think we should wait until we get `WindowOrWorkerGlobalScope`.

PR #459, opened 2018-04-28
Add HTML canvas/comms types 

saschanaz:
Especially adding many canvas types to worker thanks to the new OffscreenCanvas.


saschanaz:
This is done to add some dependent types before adding WindowOrWorkerGlobalScope for #455.

saschanaz:
Something is wrong here, `npm build` gives different result between Windows and Linux.

a-tarasyuk:
@saschanaz @mhegazy It would be great to review/merge it before merging #455, because OffscreenCanvas depends on many new types which better to automate and do not add them by manually. What do you think?

baselines/webworker.generated.d.ts
```diff
@@ -2095,6 +2274,10 @@ type HeadersInit = Headers | string[][] | Record<string, string>;
 type BodyInit = Blob | BufferSource | FormData | URLSearchParams | ReadableStream | string;
 type RequestInfo = Request | string;
 type BlobPart = BufferSource | Blob | string;
+type HTMLOrSVGImageElement = never | never;
```
mhegazy: these `never` do not look right.. 

so few things, first i think `object` is more natural type here than `never`. and second, we should firer duplicates in a union. so you do not get `object | object` or `object | BodyInit`

src/emitter.ts
```diff
@@ -323,6 +321,8 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
                     allEnumsMap[objDomType]) return objDomType;
                 // Name of a type alias. Just return itself
                 if (allTypeDefsMap.has(objDomType)) return objDomType;
+                // Known interface name but not for the target flavor
+                if (unfilteredInterfaceNames.has(objDomType)) return "never";
```
mhegazy: not sure i see why this is needed over `object` (Which is the fallback for this method anyways).

saschanaz: type X = Body | object;
var x: X = {}; // no error

saschanaz:
Now the only nevers are in typedefs. Hmm.

baselines/webworker.generated.d.ts
```diff
@@ -2095,6 +2272,10 @@ type HeadersInit = Headers | string[][] | Record<string, string>;
 type BodyInit = Blob | BufferSource | FormData | URLSearchParams | ReadableStream | string;
 type RequestInfo = Request | string;
 type BlobPart = BufferSource | Blob | string;
+type HTMLOrSVGImageElement = never;
```
mhegazy: do we need to write this? seems like noise at this point..

baselines/webworker.generated.d.ts
```diff
@@ -2103,14 +2284,19 @@ type BufferSource = ArrayBufferView | ArrayBuffer;
 type DOMTimeStamp = number;
 type FormDataEntryValue = File | string;
 type IDBValidKey = number | string | Date | BufferSource | IDBArrayKey;
-type MessageEventSource = object | MessagePort | ServiceWorker;
+type WindowProxy = never;
```
mhegazy: and here.

baselines/webworker.generated.d.ts
saschanaz: CanvasImageSource uses this, we currently don't have a way to make it conditional. I will think about it.

src/emitter.ts
```diff
@@ -323,11 +315,8 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
...
                 if (allTypeDefsMap.has(objDomType)) return objDomType;
-                // Union types
-                if (objDomType.includes(" or ")) {
-                    const allTypes: string[] = decomposeTypes(objDomType).map(t => convertDomTypeToTsTypeSimple(t.replace("?", "")));
-                    return allTypes.includes("any") ? "any" : allTypes.join(" | ");
-                }
+                // Known interface name but not for the target flavor
+                if (unfilteredInterfaceNames.has(objDomType)) return "never";
                 else {
                     // Check if is array type, which looks like "sequence<DOMString>"
```
mhegazy: you can also delete this section, do not think it is used anylonger.

src/emitter.ts
```diff
@@ -323,11 +315,8 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
                     allEnumsMap[objDomType]) return objDomType;
                 // Name of a type alias. Just return itself
                 if (allTypeDefsMap.has(objDomType)) return objDomType;
-                // Union types
-                if (objDomType.includes(" or ")) {
-                    const allTypes: string[] = decomposeTypes(objDomType).map(t => convertDomTypeToTsTypeSimple(t.replace("?", "")));
-                    return allTypes.includes("any") ? "any" : allTypes.join(" | ");
-                }
+                // Known interface name but not for the target flavor
+                if (unfilteredInterfaceNames.has(objDomType)) return "never";
```
mhegazy: do not think you need this `unfilteredInterfaceNames` map at all.. you can make `never` the fallback at the end of the function instead of the throw. and in `convertDomTypeToTsType`, add a flag to allow `never` as a type, if the flag is not set, then throw. we only set the flag in function parameters since we are handling `never` there.

saschanaz:
I'm starting to feel bad about that filtering feature is now mixed in emitter. I'll close this and open another PR.

PR #462, opened 2018-04-29
Add Form elements types

saschanaz:
Adds `labels`, `reportValidity()`, etc.


mhegazy:
Some merge conflicts requir attention

inputfiles/overridingTypes.json
```diff
@@ -1456,6 +1470,30 @@
                     }
                 }
             },
+            "HTMLButtonElement": {
```
mhegazy: so why is  not that in the spec?

mhegazy: i see. thanks.

mhegazy:
merge conflicts that need some attention.

PR #465, opened 2018-04-30
Update standard events

ldexterldesign:
Hey all,

Hope you're well

Thanks for software

Scraped from https://developer.mozilla.org/en-US/docs/Web/Events > "Standard events" section

Let me know if you have any issues - happy to help

Hope this is clear/useful

Thanks (in advance)


ldexterldesign:
Fixes https://github.com/Microsoft/TSJS-lib-generator/issues/333

mhegazy:
The file is generated from the scripts in src and not hand authored.

ldexterldesign:
Hey @mhegazy,

Thanks for reply

Apologies, I must admit I wasn't 100% sure this was the correct PR - can you provide any guidance so I can get the intellisense updated?

Specifically I'm looking for "transitionend" and related JS event properties

Hope to hear back

Yours faithfully

mhegazy:
You can take a look at https://github.com/Microsoft/TSJS-lib-generator/pull/460 and https://github.com/Microsoft/TSJS-lib-generator/pull/415 for a reference.

baselines/dom.generated.d.ts
```diff
@@ -3863,18 +3863,57 @@ interface DhKeyGenParams extends Algorithm {
...
+    "cut": ClipboardEvent;
     "dblclick": MouseEvent;
-    "deactivate": Event;
+    "devicechange": Event;
+    "devicelight": DeviceLightEvent;
+    "devicemotion": DeviceMotionEvent;
+    "deviceorientation": DeviceOrientationEvent;
+    "deviceproximity": DeviceProximityEvent;
+    "DOMActivate": UIEvent;
+    "DOMAttributeNameChanged": MutationNameEvent;
```
saschanaz: I think the current DOM spec does not include these mutation events.

ldexterldesign: Thanks for reply

TBH I'm not entirely sure where to reference the latest spec' these days..?

Specifically, which events are you referring to or all of them in the PR?

Cheers

ldexterldesign:
@mhegazy hmm, unless there's more documentation or someone to pair with I think this PR is beyond the scope of my JS/TS knowledge

baselines/dom.generated.d.ts
saschanaz: Check out this one https://dom.spec.whatwg.org/#goals

>By defining a replacement for the "Mutation Events" and "Mutation Name Event Types" chapters of DOM Level 3 Events [uievents-20031107] as the old model was problematic.

saschanaz:
You may want to:

1. Search MDN for event interfaces e.g. [AnimationEvent](https://developer.mozilla.org/en-US/docs/Web/API/AnimationEvent)
2. MDN points to [the proper spec](https://www.w3.org/TR/css-animations-1/)
3. Add them in inputfiles/idlSources.json
4. Run `npm run fetch && npm run build && npm run baseline-accept && npm test`
5. If test succeeds, open a PR

That's the recommended automatic addition way. We still have manual way, I would say you check this part https://github.com/Microsoft/TSJS-lib-generator/pull/438/files#diff-33c56fff5ba0262bfabf1ca82cfab269 for that. (Remember that automatic way is preferred.)

PR #466, opened 2018-05-01
Add obsolete types

### LINKED ISSUE #463: Do we want "Obsolete but conforming" types?
https://html.spec.whatwg.org/multipage/obsolete.html

This page includes obsolete elements and properties which are somehow still supported by browsers. Do we want to include, or exclude them?

saschanaz:
~~Used this script for `@deprecated`: https://gist.github.com/saschanaz/39f1604deef7bf0cb54e08d10d587157~~

Closes #463.


inputfiles/overridingTypes.json
```diff
@@ -120,6 +120,26 @@
                             "read-only": 1,
                             "name": "relList",
                             "type": "DOMTokenList"
+                        },
```
mhegazy: i do not think we should have these in the overriding types. i think we should mark them as part of generating the .d.ts files. this way less maintenance cost. ideally these files go away at the end module some DOM spec bugs.

saschanaz: >i think we should mark them as part of generating the .d.ts files.

Do you mean within widlprocess? I can copy-paste the script https://gist.github.com/saschanaz/39f1604deef7bf0cb54e08d10d587157 but I'm not sure how frequently we'll use that.

>ideally these files go away at the end module some DOM spec bugs.

I'm not sure I understand this?

mhegazy: I would include the script, and call it part of the widlprocess, i.e. do not write the output to overridingTypes.json, just update the idl object in memory as you are creating it by adding the deprecation bit, similar to how we merge the idls.
My intention here is that the .json files in the input folder would just be things that we manually override the spec with; in other words spec bugs. but nothing else, everything else is generated from the widls.

PR #468, opened 2018-05-01
Add CSS Animations types

saschanaz:
Adds `animation*` events, and also `touch*` events (there was a bug where old specs still not migrated to partial interface mixin syntax 62cdf09).


saschanaz:
I think we have a problem with GlobalEventHandlers. Maybe we should redefine every GlobalEventHandlers event for Document/Element/Window? What do you think? @mhegazy

mhegazy:
> I think we have a problem with GlobalEventHandlers. Maybe we should redefine every GlobalEventHandlers event for Document/Element/Window? What do you think? @mhegazy

yup. that should handle https://github.com/Microsoft/TypeScript/issues/12490

PR #470, opened 2018-05-04
Refactor

saschanaz:
This reduces some parts of code.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -16233,7 +16233,6 @@ interface HTMLElementTagNameMap {
 }
 
 interface SVGElementTagNameMap {
-    "a": SVGAElement;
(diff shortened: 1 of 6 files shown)


baselines/dom.generated.d.ts
```diff
@@ -16233,6 +16233,7 @@ interface HTMLElementTagNameMap {
 }
 
 interface SVGElementTagNameMap {
+    "a": SVGAElement;
```
mhegazy: this is gonna cause a build error, cause now the two map types are not matching.. i believe this is what is failing the build now. i remember we tried that before and chose not to do it.

saschanaz: This is because of old deprecated ElementTagNameMap extends both HTML/SVGElementTagNameMap...

PR #475, opened 2018-05-29
Add iterator to .attributes

a-tarasyuk:
[TypeScript#20937](https://github.com/Microsoft/TypeScript/issues/20937)


src/emitter.ts
```diff
@@ -1068,15 +1068,39 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
...
-            printer.printLine(`values(): IterableIterator<${valueType}>;`);
+
+            const methods = [{
+                name: 'entries',
+                definition: `IterableIterator<[${keyType}, ${valueType}]>`,
+            }, {
+                name: 'values',
+                definition: `IterableIterator<${valueType}>`,
+            }, {
+                name: 'keys',
```
mhegazy: nit. `keys` before `values` in order

src/emitter.ts
```diff
@@ -1068,15 +1068,39 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
...
+                if (comments && comments[m.name]) {
+                    printer.printLine(comments[m.name]);
+                }
+                printer.printLine(`${ m.name }(): ${ m.definition };`);
+            });
+
+            if (comments && comments['forEach']) {
+                printer.printLine(comments['forEach']);
+            }
+            emitIteratorForEach(i);
```
mhegazy: not sure why `forEach` should be in this file and not in the normal DOM declarations..

this file is only separate because it has references to ES6-specific types like `Symbol.Iterator` and `IterableIterator`. `forEach` seems not to have any of these, and i would just treat it like a normal method.

a-tarasyuk: @mhegazy ok, will remove it

baselines/dom.iterable.generated.d.ts
```diff
@@ -0,0 +1,188 @@
...
+    [Symbol.iterator](): IterableIterator<Element>;
+}
+
+interface HTMLCollectionBase {
+    [Symbol.iterator](): IterableIterator<Element>;
+}
+
+interface HTMLCollectionOf<T extends Element> {
+    [Symbol.iterator](): IterableIterator<T>;
+    entries(): IterableIterator<[number, T]>;
```
mhegazy: Do not think these exist at runtime. only `Symbol.iterator`

mhegazy: since this was already there, we can change that in a different PR.

mhegazy: image

PR #479, opened 2018-06-03
Fixed #147703 - MediaQueryListListener interface definition incorrect

irrationalRock:
This PR addresses this issue: https://github.com/Microsoft/TypeScript/issues/14703

I ported over @mkhazov code to addedTypes.json and overridingTypes.json.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4843,7 +4843,6 @@ interface DocumentEvent {
     createEvent(eventInterface: "MSMediaKeyNeededEvent"): MSMediaKeyNeededEvent;
     createEvent(eventInterface: "MediaEncryptedEvent"): MediaEncryptedEvent;
     createEvent(eventInterface: "MediaKeyMessageEvent"): MediaKeyMessageEvent;
-    createEvent(eventInterface: "MediaQueryListEvent"): MediaQueryListEvent;
     createEvent(eventInterface: "MediaStreamErrorEvent"): MediaStreamErrorEvent;
     createEvent(eventInterface: "MediaStreamEvent"): MediaStreamEvent;
     createEvent(eventInterface: "MediaStreamTrackEvent"): MediaStreamTrackEvent;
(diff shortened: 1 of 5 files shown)


baselines/dom.generated.d.ts
```diff
@@ -9820,6 +9821,25 @@ declare var MediaQueryList: {
...
+declare var MediaQueryListEvent: {
+    prototype: MediaQueryListEvent;
+    new(init: {matches: boolean, media: string}): MediaQueryListEvent;
+};
+
+interface MediaQueryListListener {
+    (mql: MediaQueryListEvent): void;
+}
+
+declare var MediaQueryListListener: {
```
mhegazy: this should not be here. only the type.

mhegazy: actually i am not sure why you even have this type.

mhegazy:
@saschanaz can you please review this change

saschanaz:
This type is from [the old 2013 CSSOM View spec](http://www.w3.org/TR/2013/WD-cssom-view-20131217/#mediaquerylistlistener), and [has been replaced](http://www.w3.org/TR/cssom-view/#dom-mediaquerylist-addlistener) with the standard EventListener.

Old:

```webidl
interface MediaQueryList {
  readonly attribute DOMString media;
  readonly attribute boolean matches;
  void addListener(MediaQueryListListener listener);
  void removeListener(MediaQueryListListener listener);
};

callback MediaQueryListListener = void (MediaQueryList list);
```

Latest:

```
interface MediaQueryList : EventTarget {
  readonly attribute DOMString media;
  readonly attribute boolean matches;
  void addListener(EventListener? listener);
  void removeListener(EventListener? listener);
           attribute EventHandler onchange;
};
```

saschanaz:
So my suggestion: remove the type entirely.

irrationalRock:
Thanks for the feedback. 
I removed `MediaQueryListListener` and made `MediaQueryList` extend `EventTarget`.

inputfiles/removedTypes.json
```diff
@@ -24,6 +24,7 @@
     },
     "interfaces": {
         "interface": {
+            "MediaQueryListListener": null,
```
saschanaz: Can you move this line to below HTMLHRElement and over MSStreamReader?

PR #480, opened 2018-06-04
Remove redundant iterator methods from HTMLCollectionOf

saschanaz:
...and only keep `[Symbol.iterator](): IterableIterator<T>`.

### DIFF
--- a/src/emitter.ts
+++ b/src/emitter.ts
@@ -817,6 +817,7 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
                     if (indexer) {
                         printer.printLine(`[${indexer.name}: ${convertDomTypeToTsType(indexer)}]: ${convertDomTypeToTsType({
                             type: m.signature[0].type,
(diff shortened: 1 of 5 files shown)


src/emitter.ts
```diff
@@ -1052,11 +1054,14 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
                 }
                 return i.iterator.type.map(convertDomTypeToTsType);
             }
-            else if (i.name !== "Window" && i.properties) {
+            else if (i.name !== "Window") {
                 const iterableGetter = findIterableGetter();
-                const lengthProperty = mapToArray(i.properties.property).find(findLengthProperty);
+                const lengthProperty = findLengthProperty(i) || findLengthProperty(allInterfacesMap[i.extends]);
```
mhegazy: should this be a recursive function to repeat on every extended type?

saschanaz: I'm not sure, as this really is only for `HTMLCollectionOf`. All other interfaces are fine without inheritance check. Should recursion be preferred?

PR #483, opened 2018-06-05
Add `willChange` to `CSSStyleDeclaration`

mDibyo:
This is a Candidate Recommendation currently.

MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/will-change#Specification


mDibyo:
Looks like I make the changes the right way. Will open a new PR.

PR #487, opened 2018-06-07
Add Screen Orientation types

saschanaz:
Fixes Microsoft/TypeScript#18480

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -13760,6 +13760,7 @@ interface Screen extends EventTarget {
     readonly logicalYDPI: number;
     readonly msOrientation: string;
     onmsorientationchange: ((this: Screen, ev: Event) => any) | null;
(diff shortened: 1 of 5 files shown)


baselines/dom.generated.d.ts
```diff
@@ -17225,6 +17247,8 @@ type NavigationReason = "up" | "down" | "left" | "right";
 type NavigationType = "navigate" | "reload" | "back_forward" | "prerender";
 type NotificationDirection = "auto" | "ltr" | "rtl";
 type NotificationPermission = "default" | "denied" | "granted";
+type OrientationLockType = "any" | "natural" | "landscape" | "portrait" | "portrait-primary" | "portrait-secondary" | "landscape-primary" | "landscape-secondary";
```
mhegazy: looks like we have another definition  of `OrientationLockType` in Addedtypes.json

PR #489, opened 2018-06-08
Add exposure module

saschanaz:
Replaces related lines from #459 

Now non-Worker exposed types are not processed as `"object"` but 1) are omitted if optional 2) throws if not.

This removes some Worker-incompatible non-standard types.


baselines/webworker.generated.d.ts
```diff
@@ -2458,7 +2472,7 @@ interface XMLHttpRequest extends XMLHttpRequestEventTarget {
      * Initiates the request. The optional argument provides the request body. The argument is ignored if request method is GET or HEAD.
      * Throws an "InvalidStateError" DOMException if either state is not opened or the send() flag is set.
      */
-    send(body?: object | BodyInit | null): void;
```
mhegazy: why is the `| null` removed?

saschanaz: Oops, fixed now.

PR #490, opened 2018-06-09
Add HTML semantics types

saschanaz:
Removes `HTMLLinkElement.import`. It was for HTML Imports but the spec hasn't been able to gather enough interest from vendors.

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -235,23 +235,6 @@
                     }
                 }
             },
-            "HTMLLinkElement": {
-                "name": "HTMLLinkElement",
-                "properties": {
-                    "property": {
-                        "disabled":{
-                            "deprecated": 0,
-                            "name": "disabled",
-                            "type": "boolean"
-                        },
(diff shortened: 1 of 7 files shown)


inputfiles/overridingTypes.json
```diff
@@ -235,23 +235,6 @@
                     }
                 }
             },
-            "HTMLLinkElement": {
```
mhegazy: we have explicitly added these in https://github.com/Microsoft/TSJS-lib-generator/pull/488 and https://github.com/Microsoft/TSJS-lib-generator/pull/425 can we keep them for now.

saschanaz: `relList` is in the spec so it's not really removed. `disable` is not, the spec does not mention it, even not as a deprecated attribute...

saschanaz: Okay, it was once in the spec, [has been removed](https://github.com/whatwg/html/issues/1081) not because it's needless but it's in inappropriate document. I'll add it again.

PR #494, opened 2018-06-10
Add HTML text level semantics types

saschanaz:
Removes nonstandard properties from HTMLAnchorElement.


inputfiles/idl/HTML - Links.commentmap.json
```diff
@@ -0,0 +1,13 @@
+{
+    "hyperlink-href": "Returns the hyperlink's URL.\nCan be set, to change the URL.",
```
mhegazy: i do not see these comments showing up

mhegazy:
Some merge conflicts need attention

inputfiles/idl/HTML - Links.commentmap.json
saschanaz: Not sure what should be done there, it seems should be `htmlhyperlinkelementutils-href` instead of `hyperlink-href` to add it to the lib file. Should it be manually mapped?

mhegazy:
looks like my manual merge did not work.

saschanaz:
Fixed!

mhegazy:
thanks!

mhegazy:
@saschanaz do you have a list of all the remaining widl files? we should probably have this on an issue to make it easier for others who wanna help to jump in and pick a few.

saschanaz:
I don't have one right now, but I hope I can get one after I add full HTML and DOM types. Currently I just try adding those types and getting their dependencies by build errors.

PR #495, opened 2018-06-11
Add form/imagemap/table/input types

saschanaz:
Removes `HTMLAreasCollection`, etc.


inputfiles/idl/HTML - Tabular data.commentmap.json
```diff
@@ -0,0 +1,25 @@
+{
+    "table-caption": "Returns the table's caption element.\nCan be set, to replace the caption element.",
```
mhegazy: comments do not seem to be showing here either.

saschanaz: Same here, it's not `htmltableelement-caption` so it doesn't show up.

Issue #501, opened 2018-06-13
Missing overload for document.createElementNS() function


stevedorries:
The createElementNS function has an optional third parameter, options, that can be either a string(for backwards compatibility) or an object. The document.createElement() function already has an equivalent function overload being generated, createElement(tagName: string, options?: ElementCreationOptions): HTMLElement;. 
There should be an overload signature like this:
createElementNS(namespaceURI: string, qualifiedName: string, options?: ElementCreationOptions): Element; 

There are existing overloads for createElementNS located in the AddedTypes JSON file, but the overloads for createElement are in the emitter.ts file, this inconsistency raises the question, in my mind, where should the additional overload for createElementNS be located?

mhegazy:
>  where should the additional overload for createElementNS be located?

addedTypes.json

PR #502, opened 2018-06-13
Proposed fix for issue 501

stevedorries:
Adds proposed function signature overload.


mhegazy:
You need to also include the generated baseline

stevedorries:
Generated baseline is now included.

mhegazy:
Sorry i merged that too fast. i think there is one more change needed.. instead of using `"additional-signatures"` use `"override-signatures"` to avoid having the extra signature at the bottom.

HolgerJeromin:
Ref #501

PR #503, opened 2018-06-14
Fixed #242025 - Added nonce

irrationalRock:
This PR addresses the following issue: https://github.com/Microsoft/TypeScript/issues/24205.

It adds `nonce` to both `HTMLScriptElement` and `HTMLStyleElement`.


HolgerJeromin:
Shouldn't this change go to addingTypes.json?

mhegazy:
@saschanaz should we add https://html.spec.whatwg.org/multipage/scripting.html#the-script-element instead?

saschanaz:
@mhegazy #496 already did that. `nonce` is part of https://html.spec.whatwg.org/multipage/dom.html, but we probably can't add this right now because `GlobalEventHandlers` mixin should be processed first. Otherwise we'll lose lots of event handlers.

saschanaz:
Also, `nonce` should be added to HTMLElement instead of HTMLScriptElement.

irrationalRock:
Thanks, I appreciate the responses.
I moved `nonce` to `HTMLOrSVGElement` and `FocusOptions`.

saschanaz:
HTMLOrSVGElement should also be added to `implements` list of HTMLElement.

FocusOptions should rather be added by https://html.spec.whatwg.org/multipage/interaction.html and also `focus()` signature should be overridden.

irrationalRock:
Thanks for the feedback, I added `HTMLOrSVGElement` to `HTMLElement`'s list.
However, this causes HTMLElementMap to lose it's extends of `ElementEventMap`.
Also, I overridden the focus() signature.
@saschanaz I'm wondering if you can give more information on what you mean by FocusOptions should be added by https://html.spec.whatwg.org/multipage/interaction.html.

saschanaz:
@irrationalRock `extends` shouldn't be overridden here, doing so will break EventMap generation which uses the field.

For FocusOptions, you can add this to inputfiles/idlSources.json:

```json
{
  "url": "https://html.spec.whatwg.org/multipage/interaction.html",
  "title": "HTML - User interaction"
}
```

...and then run `npm run fetch && npm run build && npm run baseline-accept`. It will automatically import FocusOptions from the spec document.

irrationalRock:
I added `FocusOptions` with `inputfiles/idlSources.json`.

If `extends` shouldn't be used here, what would be the best place for it?

saschanaz:
`implements` should be overridden but I'm afraid it won't work currently, as the merge function doesn't support arrays very well. Perhaps we may add `additional-implements` like the current `additional-signatures`.

saschanaz:
Oops, I was wrong. Just adding implements: ["HTMLOrSVGElement"] will append to the current list.

inputfiles/overridingTypes.json
```diff
@@ -897,6 +897,18 @@
                     }
                 }
             },
+            "HTMLElement": {
+                "methods": {
+                    "method": {
+                        "focus": {
+                            "name": "focus",
+                            "override-signatures": [
+                                "focus(options?: FocusOptions): void"
```
mhegazy: so why is this needed?

saschanaz: For https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus. (Separate from nonce, though.)

PR #504, opened 2018-06-16
Use getExposureTypes for Window

saschanaz:
It turns out that the browser json file has several unreferenced MS-prefixed types, so this PR removes them.

~~I think TSJS ultimately should remove Worker-only types (which includes ServiceWorker-only types), but this PR doesn't do it yet.~~ Oops, the master branch already does this 😁


inputfiles/knownTypes.json
```diff
@@ -0,0 +1,111 @@
+{
+    "Window": [
+        "AAGUID",
```
mhegazy: not sure i understand why we need these

saschanaz: I'm not sure too, it's just migrated from the old F#-age list. Let's just remove them.

inputfiles/knownTypes.json
```diff
@@ -0,0 +1,92 @@
+{
+    "Window": [
```
mhegazy: i actually meant the new `window` types.. i understand the the worker, but not sure i understand the new window types.

saschanaz: It's mainly because we cannot automatically track `override-signatures` and `override-type`s. Additionally there are some types that are not directly referenced in IDL by other types but still used by specs, e.g. subdictionaries of RTCStats.

mhegazy: Can not we compute this by walking the types after we merge in addedTypes?

saschanaz: We already do after addedTypes but the two `override` types are not in IDL structures but in TypeScript string format. Maybe we still can extract type names there, but I'm not sure it'll worth the effort as they ultimately will be removed...

mhegazy: fair enough. can you resolve the merge conflicts.

PR #506, opened 2018-06-20
Update Mailmap

nrcoley:
Hi - I'd really appreciate it if this PR could be merged in to update my email address in the mailmap. I'd feel more comfortable if this information was less easily available. Thanks!


mhegazy:
This file is generated from the git commit history. i am happy to remove your name altogether, but it does not change the git history.

nrcoley:
Hi @mhegazy I understand that it does not change the commit history - I'm just hoping to reduce how easy it is to access. If you could remove my name altogether that would be great, currently it comes up pretty quickly with a google search of that email (which leads to this mailmap, as well as the one for the Typescript repo that I was about to make a similar PR for) which is what I'm hoping to reduce. Thanks!

mhegazy:
can you remove the entry altogether then.

nrcoley:
Yes, sorry, thanks, just updated the PR.

PR #509, opened 2018-06-22
Add HTML canvas types

### LINKED ISSUE #473: addPath method missing from Path2D declaration
The `addPath` method is missing from `Path2D`. To fix this, it should be added to `Path2D`, [following the standard](https://html.spec.whatwg.org/multipage/canvas.html#dom-path2d-addpath)
```ts
addPath(path: Path2D, transform?: SVGMatrix)
```

saschanaz:
without OffscreenCanvas which has not been implemented yet in any stable browsers.

Closes #473


mhegazy:
> generated/webworker.generated.d.ts(434,26): error TS2304: Cannot find name 'SVGMatrix'.

saschanaz:
Done 😁

PR #510, opened 2018-06-23
Fixed #24084 - Changed HTMLSelectElement Index Signature

irrationalRock:
This PR addresses https://github.com/Microsoft/TypeScript/issues/24084.

I changed the `HTMLSelectElement` index to `[name: string]: HTMLOptionElement | HTMLOptGroupElement`.


saschanaz:
String indexers generally don't work, unless the target interface has no other members.

mhegazy:
A string indexer type should be of a type that covers every property on the type.

irrationalRock:
Thanks for the comments.

I changed the signature to `[name: number]: HTMLOptionElement | HTMLOptGroupElement`

PR #519, opened 2018-06-27
Remove unused file

krzysztof-grzybek:
Correct me if I'm wrong, but I think that this file is not used any more and it was only temporary useful on [this](https://github.com/Microsoft/TSJS-lib-generator/pull/383) branch.


mhegazy:
I still need that one. i use it to remove some MS-only declarations from input file we get from the edge team. once we have fully generated the .d.ts files from the webidl files, we can remove it.

krzysztof-grzybek:
Ok, got it. Thank You for reply and sorry for bothering.

PR #524, opened 2018-07-04
Add HTML webappapi types

saschanaz:
Almost all event handlers are moved into GlobalEventHandlers mixin.


baselines/dom.generated.d.ts
```diff
@@ -16132,27 +15918,23 @@ interface WindowLocalStorage {
...
-    setInterval(handler: any, timeout?: any, ...args: any[]): number;
-    setTimeout(handler: (...args: any[]) => void, timeout: number): number;
-    setTimeout(handler: any, timeout?: any, ...args: any[]): number;
-}
-
-interface WindowTimersExtension {
-    clearImmediate(handle: number): void;
-    setImmediate(handler: (...args: any[]) => void): number;
-    setImmediate(handler: any, ...args: any[]): number;
+interface WindowTimers {
```
saschanaz: I know this is weird, this will be removed in a following PR.

mhegazy:
@saschanaz can you remove `atob` and `btoa` duplicate declarations as well.

PR #525, opened 2018-07-06
Add HTML DOM types

saschanaz:
Diffs are mainly because of HTMLElement not including event handlers anymore.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -10941,7 +10941,7 @@ declare var ProgressEvent: {
 };
 
 interface PromiseRejectionEvent extends Event {
-    readonly promise: PromiseLike<any>;
+    readonly promise: Promise<any>;
     readonly reason: any;
 }
 
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3799,7 +3799,7 @@ interface DhKeyGenParams extends Algorithm {
     prime: Uint8Array;
 }
 
-interface DocumentEventMap extends GlobalEventHandlersEventMap {
+interface DocumentEventMap extends GlobalEventHandlersEventMap, DocumentAndElementEventHandlersEventMap {
     "abort": UIEvent;
     "activate": Event;
     "beforeactivate": Event;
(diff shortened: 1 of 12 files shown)


baselines/dom.generated.d.ts
```diff
@@ -5267,9 +5298,13 @@ interface HTMLAnchorElement extends HTMLElement, HTMLHyperlinkElementUtils {
      */
     text: string;
     type: string;
-    addEventListener<K extends keyof HTMLElementEventMap>(type: K, listener: (this: HTMLAnchorElement, ev: HTMLElementEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
+    addEventListener<K extends keyof DocumentAndElementEventHandlersEventMap>(type: K, listener: (this: HTMLAnchorElement, ev: DocumentAndElementEventHandlersEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
+    addEventListener<K extends keyof ElementEventMap>(type: K, listener: (this: HTMLAnchorElement, ev: ElementEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
```
mhegazy: `HTMLElementEventMap` already extends `ElementEventMap`

mhegazy: also i think the change here should be on `HTMLElementEventMap` to extend `DocumentAndElementEventHandlersEventMap` and `GlobalEventHandlersEventMap` instead of on every element.

saschanaz:
Yes, but that map is now empty and not generated because HTMLElement does not directly include any event handlers anymore.

Sent from my Windows 10 phone

baselines/dom.generated.d.ts
saschanaz: >HTMLElementEventMap already extends ElementEventMap

Yes, but that map is now empty and not generated because HTMLElement does not directly include any event handlers anymore.

mhegazy: we should generate the map always instead of generating overloads.. the map can still be empty, but just extends other maps.

mhegazy:
thanks!

PR #532, opened 2018-07-11
Inherit const members in constructors

### LINKED ISSUE #531: Support constructor inheritance for constants
For example, `AnimationEvent` extends from `Event` which includes `Event.AT_TARGET`, so `AnimationEvent` should also include `AnimationEvent.AT_TARGET`.

saschanaz:
Fixes #531  

This doesn't look very ideal as this basically copies everything, but not sure what should be done for those inline type literals.

Maybe we can introduce new `(InterfaceName)Constants` interfaces and use them:

```ts
declare var MyClass: {
  prototype: MyClass;
  new(): MyClass;
} & MyClassConstants;
```

What do you think? @mhegazy

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1711,6 +1711,10 @@ interface AnimationEvent extends Event {
 declare var AnimationEvent: {
     prototype: AnimationEvent;
     new(type: string, animationEventInitDict?: AnimationEventInit): AnimationEvent;
+    readonly AT_TARGET: number;
+    readonly BUBBLING_PHASE: number;
+    readonly CAPTURING_PHASE: number;
+    readonly NONE: number;
 };
 
 interface AnimationPlaybackEvent extends Event {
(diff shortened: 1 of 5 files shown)


mhegazy:
that looks like a lot of duplication really. Adding a new interface per variable does not seem much better either with the global scope pollution; the intersection does not look attractive either.

I would say the issue here is that these all should be defined as classes instead of variable+interface pairs, this way you get the static inheritance for free.

I suggest we shelve this until we have updated the DOM to use the webidl as a source first, then revisit the class generation question.

saschanaz:
>I suggest we shelve this until we have updated the DOM to use the webidl as a source first

Actually this becomes a blocker when WebGL has moved every constant from `interface WebGLRenderingContext` to [`interface mixin WebGLRenderingContextBase`](https://www.khronos.org/registry/webgl/specs/latest/1.0/). This way the constructor loses all constants when without constant inheritance.

Should we just add a special check for WebGLRenderingContext case as a temporary solution?

saschanaz:
This PR now only covers mixin cases.

PR #533, opened 2018-07-12
Add WebGL 1 types

saschanaz:
WebGL 2 types will follow later.


inputfiles/idlSources.json
```diff
@@ -230,6 +230,10 @@
         "url": "https://www.w3.org/TR/webdriver1/",
         "title": "WebDriver"
     },
+    {
+        "url": "https://www.khronos.org/registry/webgl/specs/latest/1.0/webgl.idl",
```
mhegazy: We need a licence/terms of use for this one. the other ones all came from w3.org or whatwg, and both have licences/terms of use docs on their specs, we have included the licencees in the TS Third Party Notice. See https://github.com/Microsoft/TypeScript/blob/master/ThirdPartyNoticeText.txt#L82-L167

Can you file a bug for the Khronos group to add a licence text/terms of use documents to the spec so we can safely use it.

saschanaz: Can we use this: https://www.khronos.org/legal/terms-of-use/ or do we need something else?

mhegazy: From the copyright notice:

> The contents of this site are subject to protection under U.S. and foreign copyright laws. You may not copy or distribute any portion of this site, except as necessary to view this site.

I think they need to have explicit terms of use on the documents themselves. for example: this is what W3C has, and they have it linked from all their specs: https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document

WhatWG uses Creative Commons Attribution 4.0 International License, and again, they have it linked in all their specs.

saschanaz: We've got the license: https://github.com/KhronosGroup/WebGL/pull/267

mhegazy:
Thanks!

mhegazy:
https://github.com/Microsoft/TypeScript/pull/26102

PR #534, opened 2018-07-13
Introduce Messageable and Transferable for postMessage

nikeee:
This PR addresses https://github.com/Microsoft/TypeScript/issues/25176

Relevant MDN pages:
https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel/postMessage
https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
https://developer.mozilla.org/en-US/docs/Web/API/Client/postMessage
https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage
https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/postMessage
https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope/postMessage

Map and Set do not seem to be available in `dom.generated.d.ts`. How would one implement the `Messageable` type?

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -274,7 +274,7 @@
                         "postMessage": {
                             "name": "postMessage",
                             "override-signatures": [
-                                "postMessage(message: Messageable, targetOrigin: string, transfer?: Transferable[]): void"
+                                "postMessage(message: any, targetOrigin: string, transfer?: Transferable[]): void"
                             ]
                         }
                     }
@@ -1909,7 +1909,7 @@
                         "postMessage": {
                             "name": "postMessage",
                             "override-signatures": [
-                                "postMessage(message: Messageable, transfer?: Transferable[]): void"
+                                "postMessage(message: any, transfer?: Transferable[]): void"
                             ]
                         }
                     }
@@ -2255,7 +2255,7 @@
                         "postMessage": {
                             "deprecated": 0,
                             "override-signatures": [
-                                "postMessage(message: Messageable, transferList?: Transferable[]): void"
+                                "postMessage(message: any, transfer?: Transferable[]): void"
                             ]
                         }
                     }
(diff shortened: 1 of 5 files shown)


mhegazy:
`Map` and `Set` are defined in the ES6 lib file. you will need to split the declaration into two overloads, and have the `Map`/`Set` ones in `dom.iterable.generated.d.ts`.

mhegazy:
@saschanaz can you please review.

nikeee:
> Map and Set are defined in the ES6 lib file. you will need to split the declaration into two overloads, and have the Map/Set ones in dom.iterable.generated.d.ts

How do I add types that sould only be included in `dom.iterable.generated.d.ts`.

mhegazy:
see https://github.com/Microsoft/TSJS-lib-generator/blob/master/src/index.ts#L24

nikeee:
I couldn't find any way to include different type defs without changing the emitter code. I removed the `Messageable` type def for now.

I added OffscreenCanvas to `Transferable` since it's also Transferable [according to the spec IDL](https://html.spec.whatwg.org/multipage/canvas.html#the-offscreencanvas-interface)

inputfiles/overridingTypes.json
```diff
@@ -2318,6 +2337,58 @@
...
+                }
+            },
+            "MessagePort": {
+                "name": "MessagePort",
+                "methods": {
+                    "method": {
+                        "postMessage": {
+                            "name": "postMessage",
+                            "override-signatures": [
+                                "postMessage(message: any, transferList?: Transferable[]): void"
```
saschanaz: Why `transferList` instead of `transfer` here?

nikeee: I named it `transferList` because it is landed so in the MDN docs.

https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage

I can change it to `transfer` if it's preferred.

nikeee: I changed it to `transfer`.

PR #535, opened 2018-07-17
Correct use of --project parameter

nikeee:
`tsc --help` outputs (as of Version 2.9.2):

```
...
-p FILE OR DIRECTORY, --project FILE OR DIRECTORY
...
```

In the `package.json` scripts, `--p` is used, which is wrong accorging to the help docs. This may also be the reason the appveyor build fails.

This PR changes the compiler invocation to `--project`.

Edit: The appveyor build still fails. It may be that an old `tsc` version is used that does not support `--project`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,10 @@
   "name": "tsjs-lib-generator",
   "private": true,
   "scripts": {
-    "build": "tsc --p ./tsconfig.json && node ./lib/index.js",
-    "fetch": "tsc --p ./tsconfig.json && node ./lib/fetcher.js",
+    "build": "tsc --project ./tsconfig.json && node ./lib/index.js",
+    "fetch": "tsc --project ./tsconfig.json && node ./lib/fetcher.js",
     "baseline-accept": "cpx \"generated\\*\" baselines\\",
-    "test": "tsc --p ./tsconfig.json && node ./lib/index.js && node ./lib/test.js"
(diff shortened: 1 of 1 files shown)


mhegazy:
> This may also be the reason the appveyor build fails.

I do not think this causes failures. the compiler allows both `--` and `-` in argument names. Do you have references for failures?

nikeee:
> Do you have references for failures?

I was looking at the latest build on AppVeyor: https://ci.appveyor.com/project/zhengbli/tsjs-lib-generator/build/1.0.451-rmxtjfde

It fails with the error message: `error TS5023: Unknown compiler option 'p'.`

I was suspecting that this is due to the fact that `--help` only states `-p` and `--project` are valid and it magically fails on Windows, but works on Linux. This is also supported by the unix conventions for command line arguments (`--` denoting words, `-` their respective abbreviations).

However, the build for this PR also fails. Now I'm suspecting that the used compiler version does not know `-p` because it may be too old.

From the build log I do see that it clones the latest TypeScript master and runs `npm install` in it. Maybe it's just not using the right `tsc`. I'd like to fix the AppVeyor build, but the entire build details are apparently private and there is no `appveyor.yml` (any more).

mhegazy:
I have disabled appvoyer a while back. not sure where is it getting the configurations from...

PR #541, opened 2018-07-23
Add Streams

### LINKED ISSUE #521: Add Streams types
[The Streams spec](https://streams.spec.whatwg.org/) does not provide any Web IDL so it has to be added through the manual way.

cc: @Surma

MattiasBuelens:
This PR updates the type definitions for streams to the [latest version of the spec](https://streams.spec.whatwg.org/) ([snapshot](https://streams.spec.whatwg.org/commit-snapshots/51227372cc84846bdcf68312724c4cac6a4b9e58/)). This also introduces generic type parameters for the `chunk` argument, which enables type-safe reading, writing, transforming and piping of streams.

This is largely based on [`@types/whatwg-streams`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a2a659c351b611398fd2f3f577fd545dab335306/types/whatwg-streams/index.d.ts). Important differences:
* Default values for the type parameters are set to `any` (instead of `ArrayBufferView`). This ensures that developers can still use `ReadableStream` (and others) as before and read chunks as `any`. In places where we know the precise chunk type (e.g. `BodyInit` and `Body.body`), I've added an explicit type parameter.
* Callback interfaces are used instead of callback types. This is mostly a side effect from the WebIDL processing of `callback` definitions.

Since the streams spec does not provide an official WebIDL (see whatwg/streams#45), the input WebIDL file was hand-written just for type generation purposes. The generated types are then further adjusted to add generics.

Fixes #521 
Fixes Microsoft/TypeScript#25277


MattiasBuelens:
Mainly looking for some feedback.

Is there a nicer way to introduce a type parameter on a `callback-function`? Right now, to generate this:
```typescript
interface ReadableStreamDefaultControllerCallback<R> {
    (controller: ReadableStreamDefaultController<R>): any;
}
```
I have to do add this to `overridingTypes.json`:
```json
"callback-functions": {
    "callback-function": {
        "ReadableStreamDefaultControllerCallback": {
            "name": "ReadableStreamDefaultControllerCallback<R>",
            "override-signatures": [
                "(controller: ReadableStreamDefaultController<R>): any"
            ]
        }
    }
}
```
but then I also have to add a line in `knownTypes.json`:
```json
{
    "Window": [
        "ReadableStreamDefaultControllerCallback<R>"
    ]
}
```
I tried using `"type-parameters"` on the `"callback-function"`, but that doesn't seem to work.

mhegazy:

saschanaz:
>I tried using "type-parameters" on the "callback-function", but that doesn't seem to work.

You may fix `emitCallBackFunction` in emitter.ts to make it work. You'll want to take a look at `processInterfaceType`.

inputfiles/overridingTypes.json
```diff
@@ -2400,6 +2433,274 @@
                         }
                     }
                 }
+            },
+            "ReadableStream": {
+                "name": "ReadableStream",
+                "override-exposed": "Window Worker",
```
saschanaz: `name` and `override-exposed` are redundant here. Yes, `name`s in e.g. Worker are also redundant, but no need to fix them in this PR.

inputfiles/overridingTypes.json
```diff
@@ -2400,6 +2433,274 @@
...
+                            "override-signatures": [
+                                "tee(): [ReadableStream<R>, ReadableStream<R>]"
+                            ]
+                        }
+                    }
+                }
+            },
+            "ReadableStreamReader": {
+                "name": "ReadableStreamReader",
+                "override-exposed": "Window Worker",
```
saschanaz: And here.

inputfiles/overridingTypes.json
```diff
@@ -2400,6 +2433,274 @@
...
+                            "name": "read",
+                            "override-signatures": [
+                                "read(): Promise<ReadableStreamReadResult<R>>"
+                            ]
+                        }
+                    }
+                }
+            },
+            "ReadableStreamReadResult": {
+                "name": "ReadableStreamReadResult",
```
saschanaz: Here too.

inputfiles/overridingTypes.json
```diff
@@ -2400,6 +2433,274 @@
...
+                            "name": "value",
+                            "override-type": "T"
+                        }
+                    }
+                },
+                "no-interface-object": "1"
+            },
+            "ReadableStreamBYOBReader": {
+                "name": "ReadableStreamBYOBReader",
+                "override-exposed": "Window Worker",
```
saschanaz: Also here, because the IDL has `[Exposed=(Window,Worker)]`.

inputfiles/idlSources.json
```diff
@@ -269,5 +269,9 @@
     {
         "url": "https://html.spec.whatwg.org/multipage/web-messaging.html",
         "title": "Web Messaging"
+    },
+    {
+        "url": "https://streams.spec.whatwg.org/",
+        "title": "Streams"
```
saschanaz: Can you move this above Storage? Also, `npm run fetch` will fail because the spec page does not have any IDL. We may add `"local": true` and fix fetcher.ts so that it won't fetch IDL when marked as `local`.

MattiasBuelens: I believe you mean *below* storage, since `"Storage"` < `"Streams"`? Sure, I can move "Streams" into alphabetical order.

inputfiles/overridingTypes.json
```diff
@@ -2400,6 +2455,243 @@
...
+                        "enqueue": {
+                            "override-signatures": [
+                                "enqueue(chunk: R): void"
+                            ]
+                        }
+                    }
+                }
+            },
+            "ReadableWritableStreamPair": {
+                "name": "ReadableWritableStreamPair<R extends ReadableStream, W extends WritableStream>",
```
MattiasBuelens: Right now, `TypeParameter` only supports `name` and `default`. Therefore, I simply patched in the type parameters into directly the interface's name. Ideally, this would go into the `"type-parameters"` property of the interface.

I tried to fix this by adding an `extends` property, but that immediately caused the emitter to throw a `"Maximum call stack size exceeded"` error in `getParentsWithEventHandler` and `getParentEventHandler`. I'll look into this again later, to see why this would cause infinite recursion.

MattiasBuelens: Found it! The overriding interface still needs a `name` property for this to work, which I had removed while testing. Putting the `name` back fixes it.

inputfiles/overridingTypes.json
```diff
@@ -2400,6 +2455,253 @@
...
+                            "name": "readable",
+                            "override-type": "R"
+                        },
+                        "writable": {
+                            "name": "writable",
+                            "override-type": "W"
+                        }
+                    }
+                },
+                "no-interface-object": "1"
```
saschanaz: 1. A dictionary should be in dictionary part of overridingTypes.json.
2. A dictionary doesn't need `override-exposed`.

MattiasBuelens: This is probably a leftover from my initial hack to make `<R extends ReadableStream>` work. I'll look into it.

baselines/dom.generated.d.ts
```diff
@@ -1507,11 +1524,26 @@ interface UIEventInit extends EventInit {
...
+    close?: WritableStreamDefaultControllerCallback | null;
+    start?: WritableStreamDefaultControllerCallback | null;
+    write?: WritableStreamDefaultControllerWriteCallback<W> | null;
+}
+
+interface UnderlyingSource<R = any> {
+    cancel?: ReadableStreamErrorCallback | null;
+    pull?: ReadableStreamDefaultControllerCallback<R> | null;
+    start?: ReadableStreamDefaultControllerCallback<R> | null;
+    type?: undefined;
```
MattiasBuelens: This looks a bit weird, but it is intentional. [The spec says](https://streams.spec.whatwg.org/#rs-constructor):

> 6. If _typeString_ is `"bytes"`,
> 7. Otherwise, if _type_ is `undefined`,
> 8. Otherwise, throw a `RangeError` exception.

With this type definitions, we can catch errors like this:
```
var readable = new ReadableStream({
    type: 'bytez', // oops!
    start(c) {
        c.enqueue(new Uint8Array([1]));
    }
});
```

baselines/dom.generated.d.ts
```diff
@@ -11742,20 +11774,90 @@ declare var Range: {
...
+    releaseLock(): void;
+}
+
+declare var ReadableStreamDefaultReader: {
+    prototype: ReadableStreamDefaultReader;
+    new<R = any>(stream: ReadableStream<R>): ReadableStreamDefaultReader<R>;
 };
 
-interface ReadableStreamReader {
+interface ReadableStreamReadResult<T> {
```
MattiasBuelens: A more precise type for this would be:
```typescript
type ReadableStreamReadResult<T> = {
    done: false;
    value: T;
} | {
    done: true;
    value: undefined;
};
```
However, I also want this to be compatible with `IteratorResult<T>`, which is [still defined as a regular interface](https://github.com/Microsoft/TypeScript/blob/v3.0-rc/lib/lib.es2015.iterable.d.ts#L31):
```typescript
interface IteratorResult<T> {
    done: boolean;
    value: T;
}
```
So yeah... ¯\\\_(ツ)_/¯

baselines/dom.generated.d.ts
```diff
@@ -14745,6 +14847,26 @@ declare var TrackEvent: {
...
+}
+
+declare var TransformStream: {
+    prototype: TransformStream;
+    new<R = any, W = any>(transformer?: Transformer<R, W>, writableStrategy?: QueuingStrategy<W>, readableStrategy?: QueuingStrategy<R>): TransformStream<R, W>;
+};
+
+interface TransformStreamDefaultController<R = any> {
+    readonly desiredSize: number | null;
+    enqueue(chunk: any): void;
```
MattiasBuelens: Ah dangit, I missed one.

MattiasBuelens:
@saschanaz If you find some time for another review, that'd be much appreciated! 😁 (If not, that's okay as well, I can wait.)

MattiasBuelens:
> Not sure this is maintainable, as updating things should be completely manual.

Understandable. It'd be nicer if the streams spec had a WebIDL definition, but there's still the whole "JSIDL vs WebIDL" discussion going on there. Even better would be if Microsoft Edge updated its implementation to the latest version of the spec so we could use its generated types, but it doesn't seem like that's going to happen any time soon.

I can help keep these types up-to-date. Feel free to @ me if you see any streams-related issues popping up. 😉

MattiasBuelens:
Any idea when this might land? I can wait, but it'd be nice to know when to expect this in a future TypeScript release.

I noticed that the last time a PR got merged was on August 2, and all newer PRs are still open. Are the maintainers taking a (well deserved) break, or are you working on something else?

saschanaz:
@mhegazy is (probably) taking a break 😁

image

zenmumbler:
Oh man, why did I not check this before I started my own version of adding all the types to TSJS, I only checked another issue in the main TS repo Aargh!… Anyway…, since the maintainer is MIA and this is still open for edits, can I make some? Mostly for @MattiasBuelens 

* Please change the order of the R and W parameters for `Transformer` and `TransformStream`. Reason is that it then reads as `TransformStream<InType, OutType>`. The signature of the `TransformStream` constructor is also setup this way with the writable strategy first.
* Ah, I see you changed the type from the deftyped version that I don't have to do something like`TransformStream<WritableStream<string>, ReadableStream<Uint8Array>>` but can just do the natural `TransformStream<string, Uint8Array>` Again, if the order of these params would be `from, to` instead of `to, from`

Great work, I will also gladly help out with updating these if necessary.

zenmumbler:
BTW, this will also close https://github.com/Microsoft/TypeScript/issues/17247

MattiasBuelens:
@zenmumbler Ah dangit, I didn't see that issue, otherwise I would have definitely put a link to this PR. Hope you didn't spend too much time on your own PR. 😅 

Anyway, I definitely see the benefit of swapping the type arguments on `TransformStream`. I think `TransformStream<I, O>` would be most appropriate. Originally I wanted to keep this PR somewhat close to the DefinitelyTyped type definitions, but it's probably more important to "get it right" the first time for the TypeScript DOM types.

MattiasBuelens:
Rebased on `master`, and swapped the order of the type parameters.

@zenmumbler Could you have a look? 😃

zenmumbler:
Hi @MattiasBuelens, something I found quickly, will do another round of checks in a bit:

The following types all need `"no-interface-object": 1` added so that they don't define a global constructor:

```
ReadableByteStreamController
ReadableStreamBYOBReader
ReadableStreamBYOBRequest
ReadableStreamDefaultReader
TransformStreamDefaultController
WritableStreamDefaultWriter
```

This is because these objects, according to the spec, cannot be directly constructed and have hidden/private constructor types. The spec even adds checks in the constructors themselves to throw if invoked from outside the Streams internals if the user acquired a constructor function through sneaky means.

zenmumbler:
Regarding several locations where `| null` is specified on an optional field in a dictionary: except for the `body` field in the `Body`/`Response` type and the various `desiredSize` fields, where the spec explicitly sets these fields to `null` in specific situations, these are all, if not incorrect, at least confusing.

E.g. the `preventClose`, etc fields in `QueuingStrategy` These are optional fields that will be interpreted as booleans. While the spec does a reinterpret-as-boolean operation on the field, it does not specifically mention `null` to have special meaning. In that case since it's either truthy or falsy, the type could as well be `any`.

I would remove these `| null` annotations as their explicit addition implies some kind of meaning, and that isn't there. Again, keep the ones for `desiredSize` as that gets explicitly set to `null` by the spec.

It concerns all fields in the following interfaces that have an `| null` type union.
```
PipeOptions
QueuingStrategy
Transformer
UnderlyingByteSource
UnderlyingSink
UnderlyingSource
```

sandersn:
Thanks @MattiasBuelens @zenmumbler and @saschanaz for all your work on this one. I have taken over this repo since @mhegazy has moved on, so I'll merge it when you all judge that it's done.

zenmumbler:
@MattiasBuelens Do you have time to look into this this week, I'm not actually sure how I would help out if you don't have time, would I clone your branch? Anyway, 3.2 is slated for November and I'd _really_ like to get this in for that release to avoid having to do more workarounds for my streams code. Thanks!

MattiasBuelens:
@zenmumbler Thanks for your extensive review! I'll go over it today. 😃

MattiasBuelens:
> Sorry, probably made a typo there, my reason to comment was to indicate that the `type`, `readableType` and `writableType` fields are all sort of future field placeholders and in the current version only `type` is defined and not the other two.

Ah, I see. I think we can do the `type?: undefined` trick to enforce forward-compatibility.

baselines/dom.generated.d.ts
zenmumbler: In addition, the IteratorResult type is only defined in es2015.lib as I found out the hard way.

baselines/dom.generated.d.ts
```diff
@@ -11834,20 +11864,70 @@ declare var Range: {
...
+}
+
+interface ReadableStream<R = any> {
     readonly locked: boolean;
-    cancel(): Promise<void>;
-    getReader(): ReadableStreamReader;
+    cancel(reason?: any): Promise<void>;
+    getReader(options: { mode: "byob" }): ReadableStreamBYOBReader;
+    getReader(): ReadableStreamDefaultReader<R>;
+    pipeThrough<T extends ReadableStream>({ writable, readable }: { writable: WritableStream<R>, readable: T }, options?: PipeOptions): T;
```
zenmumbler: My only niggle is that now that the `WritableReadablePair` interface is gone and with it the generic types that extend _both_ R and W streams, that this is now inconsistent with the rest of the interfaces. It is now the _only_ location that allows for a theoretical type T that extends ReadableStream, versus being a ReadableStream parameterised by a new chunk generic type (O). Like so:

```suggestion
    pipeThrough<O>({ writable, readable }: { writable: WritableStream<R>, readable: ReadableStream<O> }, options?: PipeOptions): ReadableStream<O>;
```
I realise that `T` was also used to not have to have type RS<O> twice, but personally I find this more explicit (and thus clear) and also more consistent with the other interfaces. You can choose any name for type `O` here, I used it to be consistent with output chunk types from TransformStream, but it does differ from `R` but maybe that's fine. This function is one of those instance methods that should have been a free function, but I digress.

MattiasBuelens: Fair enough, I guess it's a bit overkill to preserve the subclass type when it's unlikely that people will encounter a subclass of `ReadableStream` in practice.

zenmumbler:
Hi @sandersn, thanks for merging. This is the first time I've contributed to the lib generator and I was wondering how often these definitions are merged back in the main TS repo. Is it on the `next` track or only updated with beta/rc releases? Thank you.

MattiasBuelens:
@zenmumbler From [the history of `src/lib/dom.generated.d.ts`](https://github.com/Microsoft/TypeScript/commits/master/src/lib/dom.generated.d.ts), the team seems to pull in updates around 'big' releases, so not for every nightly release. I would think they'll do an update somewhere during the preparation for a 3.2 release?

zenmumbler:
Thanks @MattiasBuelens, I kind of need to use this today so I just resorted to copying over the contents of the TSJS generated file into my local TS installs… Not exactly scalable, but workable until 3.2(rc) comes out 😅. Only thing I can't do is publish updates to my packages until then but I have pnpm and verdaccio to take care of that.

MattiasBuelens:
@zenmumbler Microsoft/TypeScript#28343 has landed, so the new stream types are now available on `typescript@next`! 😄

PR #544, opened 2018-07-31
Changed returned type of Element.getElementsByClassName to HTMLCollectionOf<Element>

### LINKED ISSUE #543: Element.getElementsByClassName should return HTMLCollectionOf<Element>
Mdn: https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByClassName
Spec: https://dom.spec.whatwg.org/#dom-element-getelementsbyclassname

timocov:
Fixes #543


mhegazy:

PR #547, opened 2018-08-01
Add HTMLDialogElement DOM Typing

AbhiPrasad:
First time doing this, so not sure if everything is correct. Would appreciate a point in the right direction if I'm doing this wrong!

Fixes https://github.com/Microsoft/TypeScript/issues/26081


AbhiPrasad:
@mhegazy Any idea why the tests are failing? Super confused.

saschanaz:
The PR itself looks okay, it seems some change in TS nightly build caused the error.

PR #560, opened 2018-08-02
Fix lib breaks

mhegazy:
Fixes a set of issues that were identified by our real world code test suits.


saschanaz:
Can we mark the deprecated argument of createTreeWalker as deprecated? ALso `caretRangeFromPoint`.

mhegazy:
> Can we mark the deprecated argument of createTreeWalker as deprecated? Also caretRangeFromPoint and almost all other additions here.

sounds like a good idea.

PR #562, opened 2018-08-04
Add partial CSSStyleDeclaration

saschanaz:
... as [CSSOM spec says](https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface):

> For each CSS property *property* that is a supported CSS property, the following partial interface applies where *camel-cased attribute* is obtained by running the CSS property to IDL attribute algorithm for property.
> ```webidl
> partial interface CSSStyleDeclaration {
>   [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _camel_cased_attribute;
> };
> ```


src/fetcher.ts
```diff
@@ -45,12 +58,29 @@ async function fetchIDL(source: IDLSource) {
             }
             return !previous.classList.contains("atrisk") && !previous.textContent!.includes("IDL Index");
         });
-    if (!elements.length) {
-        throw new Error(`Found no IDL code from ${source.url}`);
+    return elements.map(element => trimCommonIndentation(element.textContent!).trim()).join('\n\n');
+}
+
+function extractCSSDefinitions(dom: DocumentFragment) {
+    const properties =  Array.from(dom.querySelectorAll("dfn.css[data-dfn-type=property],.propdef dfn"))
```
sandersn: double-space here

sandersn: You should probably create cssSelector to match idlSelector instead of having a literal here.

src/fetcher.ts
```diff
@@ -45,12 +58,29 @@ async function fetchIDL(source: IDLSource) {
...
+        .map(element => element.textContent!.trim());
+
+    if (!properties.length) {
+        return "";
     }
-    const idl = elements.map(element => trimCommonIndentation(element.textContent!).trim()).join('\n\n');
-    const comments = processComments(dom);
-    return { idl, comments };
+
+    return `partial interface CSSStyleDeclaration {${
```
sandersn: is using code to augment in the inputfile's IDL normal? I thought that additions to the json files were the ways to do additions. 

Please excuse my lack of knowledge of the system, since I haven't worked with it much yet.

saschanaz: Adding CSS properties [implicitly augments `CSSStyleDeclaration`](https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface):

>For each CSS property property that is a supported CSS property, the following partial interface applies where camel-cased attribute is obtained by running the CSS property to IDL attribute algorithm for property.
>```webidl
>partial interface CSSStyleDeclaration {
>  [CEReactions] attribute [TreatNullAs=EmptyString] CSSOMString _camel_cased_attribute;
>};
>```

So I think it's natural to augment IDL.

Issue #558, opened 2018-08-02
document.body should not be nullable, just HTMLElement


sandersn:
Making it nullable is way too pedantic. It privileges the corner case where it could be null over the many, many uses where it could not possibly be null.

mhegazy:
Should be fixed by https://github.com/Microsoft/TSJS-lib-generator/pull/560

PR #565, opened 2018-08-06
Make Document#documentElement non-null

andy-ms:
Fixes an issue similar to #558.
Discovered in `puppeteer` on DefinitelyTyped.


vkrol:
What do you think about document.head and document.location nullability?

vkrol:
@andy-ms thank you!

PR #567, opened 2018-08-17
Update Web Animations keyframe definitions

birtles:
This is an initial attempt at fixing https://github.com/Microsoft/TypeScript/issues/26073

It should possibly be split up into several patches/PRs but I wanted to get some initial feedback first.

A few notes:
* I've included the update to the `CompositeOperationOrAuto` type which was [recent spec change](https://github.com/w3c/csswg-drafts/commit/ced6b5aac02e0e021001f10e4d63e03456686dad), simply because it makes the index type for `PropertyIndexedKeyframes` a little more sane.
* I have no idea how to represent this sort of open dictionary with a few typed keys in TypeScript so this approach might be completely wrong. Bear in mind that with the arrival of animatable custom properties the set of properties that can be specified on a keyframe is essentially open-ended.
* Allowing `number` for the property values is actually ok since WebIDL will perform `ToString` on passed-in numbers so it is actually quite valid (and common) to specify a keyframe as `{ opacity: 0 }` and let WebIDL turn the zero into a string.
* I've made the members of `ComputedKeyframe` required since these dictionary objects are only returned from `getKeyframes()` where they are guaranteed to be filled-in.

Edit: Updated the example about stringification from `{ offset: 0 }` to `{ opacity: 0 }` (since that's what I originally meant to say, particularly since `offset` is actually supposed to be a number).

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -201,7 +201,7 @@ interface ComputedEffectTiming extends EffectTiming {
 }
 
 interface ComputedKeyframe extends Record<keyof CSSStyleDeclaration, string> {
-    composite?: CompositeOperation | null;
+    composite?: CompositeOperationOrAuto;
     computedOffset?: number;
     easing?: string;
     offset?: number | null;
@@ -567,7 +567,7 @@ interface KeyboardEventInit extends EventModifierInit {
 }
 
 interface Keyframe extends Record<keyof CSSStyleDeclaration, string> {
-    composite?: CompositeOperation | null;
+    composite?: CompositeOperationOrAuto;
     easing?: string;
     offset?: number | null;
 }
@@ -922,7 +922,7 @@ interface PromiseRejectionEventInit extends EventInit {
 }
 
 interface PropertyIndexedKeyframes extends Record<keyof CSSStyleDeclaration, string | string[]> {
-    composite?: CompositeOperation | (CompositeOperation | null)[];
+    composite?: CompositeOperationOrAuto | CompositeOperationOrAuto[];
     easing?: string | string[];
     offset?: number | (number | null)[];
 }
@@ -17648,6 +17648,7 @@ type ChannelCountMode = "max" | "clamped-max" | "explicit";
 type ChannelInterpretation = "speakers" | "discrete";
 type ClientTypes = "window" | "worker" | "sharedworker" | "all";
 type CompositeOperation = "replace" | "add" | "accumulate";
+type CompositeOperationOrAuto = "replace" | "add" | "accumulate" | "auto";
 type DirectionSetting = "" | "rl" | "lr";
 type DisplayCaptureSurfaceType = "monitor" | "window" | "application" | "browser";
 type DistanceModelType = "linear" | "inverse" | "exponential";
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -200,11 +200,12 @@ interface ComputedEffectTiming extends EffectTiming {
     progress?: number | null;
 }
 
-interface ComputedKeyframe extends Record<keyof CSSStyleDeclaration, string> {
-    composite?: CompositeOperationOrAuto;
-    computedOffset?: number;
-    easing?: string;
-    offset?: number | null;
+interface ComputedKeyframe {
+    composite: CompositeOperationOrAuto;
+    computedOffset: number;
+    easing: string;
+    offset: number | null;
+    [property: string]: string | number | null | undefined;
 }
 
 interface ConfirmSiteSpecificExceptionsInformation extends ExceptionInformation {
@@ -566,10 +567,11 @@ interface KeyboardEventInit extends EventModifierInit {
     repeat?: boolean;
 }
 
-interface Keyframe extends Record<keyof CSSStyleDeclaration, string> {
+interface Keyframe {
     composite?: CompositeOperationOrAuto;
     easing?: string;
     offset?: number | null;
+    [property: string]: string | number | null | undefined;
 }
 
 interface KeyframeAnimationOptions extends KeyframeEffectOptions {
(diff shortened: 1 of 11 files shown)


saschanaz:
Using `npm run fetch` is preferred over manually adding types in `addedTypes.json`.

Expected steps:

1. `npm run fetch`
2. Cancel WebGL changes, as it currently causes breakage.
3. `npm run build && npm run baseline-accept`

birtles:
Oh, nice! I'll give that a try.

birtles:
I've updated and split into two commits: One to run `npm fetch` and update the definitions (except WebGL and Web Speech API since they're broken), and then one to make the changes to Web Animations keyframe definitions.

baselines/dom.generated.d.ts
```diff
@@ -200,11 +200,12 @@ interface ComputedEffectTiming extends EffectTiming {
...
-    composite?: CompositeOperation | null;
-    computedOffset?: number;
-    easing?: string;
-    offset?: number | null;
+interface ComputedKeyframe {
+    composite: CompositeOperationOrAuto;
+    computedOffset: number;
+    easing: string;
+    offset: number | null;
+    [property: string]: string | number | null | undefined;
```
saschanaz: >I have no idea how to represent this sort of open dictionary with a few typed keys

How about using `extends Partial<Record<keyof CSSStyleDeclaration, string | number | null | undefined>>`, while it's somewhat long?

birtles: Does that work for `{ '--variable': '5px' }` though?

saschanaz: Ah, right, custom properties. 😭

baselines/dom.generated.d.ts
```diff
@@ -921,10 +923,11 @@ interface PromiseRejectionEventInit extends EventInit {
     reason?: any;
 }
 
-interface PropertyIndexedKeyframes extends Record<keyof CSSStyleDeclaration, string | string[]> {
-    composite?: CompositeOperation | (CompositeOperation | null)[];
+interface PropertyIndexedKeyframes {
+    composite?: CompositeOperationOrAuto | CompositeOperationOrAuto[];
     easing?: string | string[];
     offset?: number | (number | null)[];
+    [property: string]: string | string[] | number | null | (number | null)[] | undefined;
```
saschanaz: Could you explain the `(number | null)[]` thing? Do we also need `(string | null)[]`?

saschanaz: > the type of indexed properties can't conflict with the type of other properties specified on the interface

Ah, that's right. Thanks!

inputfiles/overridingTypes.json
```diff
@@ -2527,15 +2527,41 @@
...
+                ]
             },
             "BaseComputedKeyframe": {
                 "name": "ComputedKeyframe",
-                "extends": "Record<keyof CSSStyleDeclaration, string>"
+                "members": {
+                    "member": {
+                        "offset": {
+                            "name": "offset",
+                            "required": 1
```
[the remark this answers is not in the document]
saschanaz: Makes sense!

You don't need the `name` field anyway, it's only for adding new members.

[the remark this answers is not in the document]
birtles: Great, thanks. Should be fixed now.

birtles:
> LGTM with the name field nits. Probably @mhegazy will review and merge this later, but unfortunately that may take some time as it seems he is not active here these days.

Thanks for the super quick review!

That's fine, it will probably take my employer a few days to approve the CLA anyway.

birtles:
I got approval to sign the CLA so this should be all good to go now. I'd recommend NOT squashing and merging since I've split into two commits: (1) to just do the `npm run fetch` and (2) to do the Web Animations-specific changes. Both of them are atomic in that the tests pass after each step.

birtles:
@mhegazy review ping?

This is blocking anyone using Web Animations from updating to TS3.

saschanaz:
Or maybe @RyanCavanaugh, as I think @mhegazy is on vacation.

birtles:
@mhegazy? @RyanCavanaugh? ping?

birtles:
Review ping? It's been ~6 weeks or so.

/@andy-ms Do you know anyone who might be able to review/merge this?

jonchardy:
@mhegazy @DanielRosenwasser 

Someone please review and merge this. Two months later and it's still blocking all of us using the Web Animations API from updating to TS3.

inputfiles/idl/Web Animations.widl
```diff
@@ -93,22 +93,22 @@ interface KeyframeEffect : AnimationEffect {
...
 
 dictionary BaseComputedKeyframe {
-     double?             offset = null;
-     double              computedOffset;
-     DOMString           easing = "linear";
-     CompositeOperation? composite = null;
+     double?                  offset = null;
+     double                   computedOffset;
+     DOMString                easing = "linear";
+     CompositeOperationOrAuto composite = "auto";
```
DanielRosenwasser: Are these `.widl` files supposed to be hand-edited? Next time anyone runs the `fetcher` script, they'll get overridden, so I would imagine you want to use `overridingTypes.json` to achieve what you want.

birtles: Hi Daniel, thanks for looking at this PR. Please see the comment of this commit (the first of the two). That is, this is literally the result of running the fetcher script so this will not get overridden next time someone runs it.

birtles:
Hi @DanielRosenwasser, I responded to your comment but then marked the conversation as resolved so it's possible you may miss it.

I've set this PR up as two commits (which should ideally be merged without squashing):

1. One which updates the definitions using the fetcher script. The exact steps used are described in the extended commit message.
2. One which updates the other files.

DanielRosenwasser:
Ah, gotcha! Sorry, it's not totally clear to me and I'm not as familiar with the mechanics of reviewing this repo. Thanks for being patient with me though.

DanielRosenwasser:
Just as a heads up, Mohamed is unfortunately no longer our dev lead (:sad:). That's partially why you've seen more delays. Try pinging me in the future though.

birtles:
Thanks Daniel! I hope someone can merge this soon.

birtles:
@DanielRosenwasser Any chance you could merge this PR? Thanks!

DanielRosenwasser:
Thank you for all your patience!

PR #568, opened 2018-08-24
Allow function as resolver in document.evaluate

KnisterPeter:
Add function reference to `document.evaluate` signature.

Closes Microsoft/TypeScript#26437


KnisterPeter:
Already closed by #566

PR #569, opened 2018-09-03
Add kid to JsonWebKey definition

funkybob:
This is to address https://github.com/Microsoft/TypeScript/issues/26854

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -546,6 +546,7 @@ interface JsonWebKey {
     ext?: boolean;
     k?: string;
     key_ops?: string[];
+    kid?: string;
     kty?: string;
     n?: string;
     oth?: RsaOtherPrimesInfo[];
--- a/baselines/webworker.generated.d.ts
+++ b/baselines/webworker.generated.d.ts
@@ -216,6 +216,7 @@ interface JsonWebKey {
     ext?: boolean;
     k?: string;
     key_ops?: string[];
+    kid?: string;
     kty?: string;
     n?: string;
     oth?: RsaOtherPrimesInfo[];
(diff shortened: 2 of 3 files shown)


sandersn:
@funkybob you need to commit the generated code in addition to the widl change.

HolgerJeromin:
Thanks for jumping in.
This is not in the widl from upstream.
So the [correct place](https://github.com/Microsoft/TSJS-lib-generator#input-files) would be the addTypes.json

funkybob:
@sandersn I honestly don't know how to make the "generated code"... I ran into this problem when trying to migrate a project to TypeScript, figured I'd trying to solve it, and got this far before I ran out of time to contribute, and worked around it locally.

sandersn:
It's `npm run build` and then you diff `baseline` and `generated`. If it looks right, then you run `npm run baseline-accept`.

saschanaz:
Looks like this is not in the WebCrypto spec. [Filing an issue there](https://github.com/w3c/webcrypto/issues) may help. (Reposting from https://github.com/Microsoft/TypeScript/issues/26854#issuecomment-430450762)

sandersn:
Sounds like we should close this here then.

pranjaldaffodil:
Can't believe nobody bothered to report this to W3C in 8 years! I have opened https://github.com/w3c/webcrypto/issues/427 now

pranjaldaffodil:
@sandersn what's the process once WebCrypto updates it on their side?

saschanaz:
It will be automatically updated as the crawled spec data is served by [`@webref/idl`](https://www.npmjs.com/package/@webref/idl)

pranjaldaffodil:
> It will be automatically updated as the crawled spec data is served by [`@webref/idl`](https://www.npmjs.com/package/@webref/idl)

What about adding an interface for JWK Set? It should be relatively simple to create, but will that also require changes in WebCrypto spec?

I noticed that not just `kid` but there were even more missing parameters/attributes and have added them to the above mentioned issue.

saschanaz:
Not being in the spec implies not being in browser implementations, so yes

PR #572, opened 2018-10-03
Removed duplicate "the" from documentation

PunKeel:
https://github.com/Microsoft/TypeScript/issues/27528

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -7135,7 +7135,7 @@ interface HTMLMediaElement extends HTMLElement {
      */
     readonly seekable: TimeRanges;
     /**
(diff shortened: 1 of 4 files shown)


DanielRosenwasser:
Thanks!

saschanaz:
`*.commentmap.json` must not be hand-written as it will then be overwritten by `npm run fetch`. This specific case has been fixed in the upstream (whatwg/fetch#822), though.

DanielRosenwasser:
Good call, we should really have `.generated` or `.external` in the name then.

saschanaz:
Actually anything in `inputfiles/idl` shouldn't be hand-written 😀

I filed #602 to prevent this in CI level, but it seems the readme should also be updated as currently it only says about `browser.webidl.json`.

PR #576, opened 2018-10-08
Add SVGAnimationElemet type

nissy-dev:
Related Microsoft/TypeScript#24001

I added SVGAnimationElemet type for SVGAnimateElement, SVGAnimateMotionElement, SVGAnimateTransformElement.
If there are any problems, please tell me 🙇

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -12062,6 +12062,22 @@ declare var SVGAnimatedTransformList: {
     new(): SVGAnimatedTransformList;
 };
 
+interface SVGAnimationElement extends SVGElement {
+    readonly targetElement: SVGElement;
+    getCurrentTime(): number;
+    getSimpleDuration(): number;
+    getStartTime(): number;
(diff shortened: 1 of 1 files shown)


inputfiles/idl/SVG - Basic Data Types and Interfaces.widl
```diff
@@ -180,6 +180,14 @@ interface SVGAnimatedLengthList {
   [SameObject] readonly attribute SVGLengthList animVal;
 };
 
+[Exposed=Window]
+interface SVGAnimationElement : SVGElement {
+  readonly attribute SVGElement targetElement;
+  float getStartTime();
+  float getCurrentTime();
+  float getSimpleDuration();
+};
```
HolgerJeromin: Thanks for jumping in.
I suspect no, so the [correct place](https://github.com/Microsoft/TSJS-lib-generator#input-files) would be the addTypes.json

nd-02110114:
@HolgerJeromin 
I fixed!!

PR #578, opened 2018-10-08
Override XPathNSResolver and document's evaluate method

Fabianopb:
According to the issue discussed in https://github.com/Microsoft/TypeScript/issues/26437 `XPathNSResolver`'s property `lookupNamespaceURI` should also return `null` and `Document`'s `evaluate` method should accept  _"a function which returns a String and takes a String parameter instead of the resolver parameter."_ (and also return `null` if needed).

I had originally created the PR in https://github.com/Microsoft/TypeScript/pull/27592, my bad, this is the correct place to change.


DanielRosenwasser:
Thank you!

PR #579, opened 2018-10-13
Remove wheelDelta fields from WheelEvent

lgoenner:
Microsoft/TypeScript#27886

These fields originated from the deprecated [MouseWheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseWheelEvent) type, which is currently defined as `type MouseWheelEvent = WheelEvent;`.

I couldn't find any references to this type so it might make sense to remove it, since it already is incorrect and will be broken even more through this fix.
`onmousewheel: ((this: Window, ev: Event) => any) | null;`
The corresponding listeners simply use `Event`.

If there is any reason to keep MouseEventWheel, I think creating an extra interface for it would be the best solution. Feedback is welcome.

### DIFF
--- a/inputfiles/removedTypes.json
+++ b/inputfiles/removedTypes.json
@@ -193,6 +193,15 @@
                         "ontouchstart": null
                     }
                 }
+            },
+			"WheelEvent": {
+                "properties": {
+                    "property": {
+                        "wheelDelta": null,
+                        "wheelDeltaX": null,
+                        "wheelDeltaY": null
+                    }
+                }
             }
         }
     },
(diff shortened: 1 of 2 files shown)


inputfiles/removedTypes.json
```diff
@@ -193,6 +193,15 @@
                         "ontouchstart": null
                     }
                 }
+            },
+			"WheelEvent": {
```
saschanaz: whitespace! 😆

lgoenner: whoops, fixed

PR #178, opened 2016-12-22
Add "| boolean" to pixelStorei method parameter

VStas:
For issue (https://github.com/Microsoft/TypeScript/issues/13093)

PR #583, opened 2018-10-17
Allow GLBoolean in pixelStorei

AVGP:
related to https://github.com/microsoft/typescript/issues/27542

### DIFF
--- a/inputfiles/idl/WebGL 1.widl
+++ b/inputfiles/idl/WebGL 1.widl
@@ -653,7 +653,7 @@ interface mixin WebGLRenderingContextBase
     void lineWidth(GLfloat width);
     void linkProgram(WebGLProgram program);
     void pixelStorei(GLenum pname, GLint param);
-    void pixelStorei(GLenum pname, GLBoolean param);
+    void pixelStorei(GLenum pname, GLboolean param);
     void polygonOffset(GLfloat factor, GLfloat units);
 
     void readPixels(GLint x, GLint y, GLsizei width, GLsizei height,
--- a/inputfiles/idl/WebGL 1.widl
+++ b/inputfiles/idl/WebGL 1.widl
@@ -653,6 +653,7 @@ interface mixin WebGLRenderingContextBase
     void lineWidth(GLfloat width);
     void linkProgram(WebGLProgram program);
     void pixelStorei(GLenum pname, GLint param);
+    void pixelStorei(GLenum pname, GLBoolean param);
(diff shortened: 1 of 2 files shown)


AVGP:
The reason for this change is that while the GL spec specifies the type for `param` to be `GLint`, the table in https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glPixelStore.xhtml demonstrates that `GLboolean` is a valid type, depending on the parameter to be set.

fatcerberus:
Is it possible to use overloads to enforce the correct type based on the specific parameter being set?  That would be ideal if it's possible to do so.

fatcerberus:
You’d have to use literal types, but this requires the value of the GLenum constants to be known at compile time - if they aren’t compile-time constants (e.g. members of an `enum`, or known numeric values) then you can’t do it.

Just to illustrate the concept if you aren’t familiar with it:
```ts
func(control: 0, value: number);
func(control: 1, value: string);
```
So for the above function, if you pass 0 as the first arg, the second must be a number, if the first arg is 1 you have to pass a string.

I have a feeling the `gl.CONSTANT_NAME` values might not be static constants and/or are browser-specific though (so you couldn’t put them into an `enum`). :(

sandersn:
@fatcerberus I looked at WebGl 1.widl and it defines GLenum as `unsigned long`. Then there are lots of related constants in WebGLRenderingContextBase, but the generated d.ts doesn't include their values. It should. But this requires a change to the generator code, which I'm not up to speed on yet.

I would also need to find out if widl supports overloads discriminated by integer literal types. If not, then the change would have to go in overridingTypes.json (I think).

@AVGP It turns out that this change WAS in dom.generated.d.ts, but it got lost when we switched away from IE's idl files to standard ones. Can you just copy the change from #178 instead of adding an overload? I think it's better for the simple fix.

inputfiles/idl/WebGL 1.widl
```diff
@@ -653,6 +653,7 @@ interface mixin WebGLRenderingContextBase
     void lineWidth(GLfloat width);
     void linkProgram(WebGLProgram program);
     void pixelStorei(GLenum pname, GLint param);
+    void pixelStorei(GLenum pname, GLboolean param);
```
saschanaz: IDL files shouldn't be manually changed, instead overridingTypes.json can.

orta:
This is deprecated by #638

PR #585, opened 2018-10-24
add referrerPolicy to HTMLIFrameElement

emraa:
Hi, first time contributor, so beware.

Trying to add the missing "referrerPolicy" attribute to iframe:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-referrerpolicy

Would like to check a few things:

- Can someone confirm referrerPolicy should be "readonly"? I'm not 100% sure how to determine this, following the "sandbox" example.
- string or ReferrerPolicy type? (In other use-cases, referrerPolicy is mostly defined as "string" whereas there is one example usage of type ReferrerPolicy. The type has the correct values though)
- Should I open a bug on main TS repository referencing this, or is this fine as such?

Thanks.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -6504,6 +6504,7 @@ interface HTMLIFrameElement extends HTMLElement, GetSVGDocument {
      * Sets or retrieves the frame name.
      */
     name: string;
+    readonly referrerPolicy: ReferrerPolicy;
     readonly sandbox: DOMTokenList;
     /**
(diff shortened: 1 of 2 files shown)


sandersn:
Hi @emraa, I just took over this repo, so I'm not sure my answers are correct. Nevertheless, here they are:

* I tried to read the specs for sandbox and policy and I'm still not sure; following the example of sandbox seems reasonable.
* ReferrerPolicy is the better type.
* No need for a separate bug, I think.

PR #592, opened 2018-10-29
WebGLRenderingContext.uniform[1234][fv]() accepts ArrayLike<number>

Istar-Eldritch:
Related to:
https://github.com/Microsoft/TypeScript/issues/28198

Validated with MDN:
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/uniform


saschanaz:
I'm not sure MDN says ArrayLike is allowed. [Float32List is defined as](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14):

```webidl
typedef ([AllowShared] Float32Array or sequence<GLfloat>) Float32List;
```

It does allow sequences, but [sequences require `Symbol.iterator` support](https://heycam.github.io/webidl/#es-sequence). `ArrayLike` in TypeScript is invalid because it does not include `Symbol.iterator`.

orta:
Thanks for the PR! This looks like it was superseded #699

PR #593, opened 2018-10-31
Add FullscreenOptions, etc.

saschanaz:
This adds `FullscreenOptions` for `requestFullscreen()` and some comment changes.


saschanaz:
It would be great for reviewers if CI can alert if the IDL change is just hand-written and not from `npm run fetch`. I have no idea how it could be done.

HolgerJeromin:
This is a good idea. 
But when the upstream source changes the next (unrelated) pr would fail... :-(

PR #595, opened 2018-10-31
Change `Event.srcElement` type to `EventTarget` and expose it to Worker

FrogTheFrog:
This PR changes `Event.srcElement` type to `EventTarget` and exposes it to Worker. Accepting this PR should close https://github.com/Microsoft/TypeScript/issues/28252, however I would like to hear from @mhegazy first as to why `Element` type was chosen. Maybe there's a specific reason I don't know and now I'm ruining everything :).

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -2227,8 +2227,7 @@
                 "properties": {
                     "property": {
                         "srcElement": {
-                            "exposed": "Window",
-                            "type": "Element",
+                            "exposed": "Window Worker",
                             "nullable": 1,
                             "deprecated": 1
                         }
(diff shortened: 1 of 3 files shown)


inputfiles/overridingTypes.json
```diff
@@ -2227,8 +2227,7 @@
                 "properties": {
                     "property": {
                         "srcElement": {
-                            "exposed": "Window",
-                            "type": "Element",
+                            "exposed": "Window Worker",
```
saschanaz: You can also remove `exposed` as it's inherited by default.

FrogTheFrog:
@saschanaz Could you merge this if there are no other problems with this?

saschanaz:
Pinging @sandersn @DanielRosenwasser for this, as I have no write access.

Issue #551, opened 2018-08-02
document.fullscreenElement gone


andy-ms:
`interface Document` used to define a property `readonly fullscreenElement: Element | null;`. It looks like as of Microsoft/TypeScript#25944 this property is missing. It doesn't seem to be deprecated so maybe it should be added back?
Detected in `vast-client` on DefinitelyTyped.

andy-ms:
`document.onfullscreenchange` and `document.documentElement.requestFullscreen` are gone too, although browser compatibility for those isn't as good. Detected in `w3c-screen-orientation` on DefinitelyTyped.

andy-ms:
This doesn't seem to be fixed in `typescript@3.1.0-dev.20180804` -- `document.fullscreenElement` is still a compile error.

magcius:
The PR added the events and methods back but not the property.

magcius:
PR attached.

saschanaz:
@github-actions close

PR #597, opened 2018-11-01
Fix DocumentOrShadowRoot.fullscreenElement

magcius:
DocumentOrShadowRoot needs to be a mixin, not an interface,
to allow us to extend it without addedTypes.json overwriting it.

Fixes issue #551

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -19,6 +19,58 @@
                     ]
                 }
             },
+            "DocumentOrShadowRoot": {
+                "name": "DocumentOrShadowRoot",
+                "methods": {
+                    "method": {
+                        "getSelection": {
+                            "name": "getSelection",
+                            "override-signatures": [
+                                "getSelection(): Selection | null"
+                            ]
+                        },
+                        "elementFromPoint": {
+                            "name": "elementFromPoint",
+                            "override-signatures": [
+                                "elementFromPoint(x: number, y: number): Element | null"
+                            ]
+                        },
+                        "elementsFromPoint": {
+                            "name": "elementsFromPoint",
+                            "override-signatures": [
+                                "elementsFromPoint(x: number, y: number): Element[]"
+                            ]
+                        },
+                        "caretRangeFromPoint": {
+                            "name": "caretRangeFromPoint",
+                            "deprecated": 1,
+                            "override-signatures": [
+                                "caretRangeFromPoint(x: number, y: number): Range"
+                            ]
+                        },
+                        "caretPositionFromPoint": {
+                            "name": "caretPositionFromPoint",
+                            "override-signatures": [
+                                "caretPositionFromPoint(x: number, y: number): CaretPosition | null"
+                            ]
+                        }
+                    }
+                },
+                "properties": {
+                    "property": {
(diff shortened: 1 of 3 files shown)


inputfiles/idl/Fullscreen.widl
```diff
@@ -15,6 +15,6 @@ partial interface Document {
   attribute EventHandler onfullscreenerror;
 };
 
-partial interface DocumentOrShadowRoot {
+partial interface mixin DocumentOrShadowRoot {
```
saschanaz: IDL files must not be hand-written. Currently the code takes care of partial interfaces incorrectly extends interface mixins, so IMO no need to change this manually.

inputfiles/addedTypes.json
```diff
@@ -1175,7 +1175,10 @@
                         }
                     }
                 },
-                "no-interface-object": "1"
+                "no-interface-object": "1",
+                "includes": [
+                    "DocumentOrShadowRoot"
```
saschanaz: We don't need this manual `includes` field as the DOM spec already takes care of this.

magcius:
Out of curiosity, do you know where the upstream Fullscreen webidl is? DOM.webidl defines DocumentOrShadowRoot as an `interface mixin`, but Fullscreen.webidl just uses `partial interface`. You mentioned it wasn't hand-written but I couldn't find any other source.

saschanaz:
The IDL files are fetched from upstream web specifications. I just filed https://github.com/whatwg/fullscreen/pull/141 to fix it..

magcius:
Ah, I didn't realize the IDL was embedded in the .bs file then :)

inputfiles/addedTypes.json
```diff
@@ -19,6 +19,60 @@
...
+                            "override-type": "Element | null"
+                        },
+                        "styleSheets": {
+                            "name": "styleSheets",
+                            "read-only": 1,
+                            "override-type": "StyleSheetList"
+                        }
+                    }
+                },
+                "no-interface-object": "1"
```
saschanaz: We still have `no-interface-object` 😁

magcius: This was copied from the other location. I don't know what this does.

magcius: Looks like it's already set internally for all mixin interfaces. I'll remove it.

saschanaz: It tells the code that no global object declarations (`declare var DocumentOrShadowRoot`) should be generated. Interfaces may need it because by default they have global constructor objects, whereas mixins do not.

inputfiles/addedTypes.json
```diff
@@ -19,6 +19,59 @@
                     ]
                 }
             },
+            "DocumentOrShadowRoot": {
+                "name": "DocumentOrShadowRoot",
+                "exposed": "Window",
```
saschanaz: Probably the last request, but `exposed` is also redundant here for mixins, as by default it follows the host interface (which are `Document` and `ShadowRoot` in this case).

magcius:
Quick ping to have this merged?

Syam1388 (approved):

PR #602, opened 2018-11-08
Prevent hand-written changes from PRs

saschanaz:
Sometimes contributors just modifiy IDL files instead of calling `npm run fetch`. This PR prevents that by comparing with upstream sources in Travis.


DanielRosenwasser:
Two thoughts:

1. Doesn't this mean running the script will make CI fail?
2. I wonder if we could use https://github.com/danger/danger-js to make this cleaner.

saschanaz:
1. No, as the CI will pass if the changed IDLs are identical with the upstream. Hand-written changes will also pass if it correctly reflects the upstream, anyway...
2. It looks great, `danger.git.modified_files` is so simple 👍

saschanaz:
I tried using dangerjs, but it looks like it requires a bot? Do we want a bot to do this?

HolgerJeromin:
> Doesn't this mean running the script will make CI fail?

This is probably not so nice from the user perspective, but the correct thing to do. Because we wont accept changes which violates the rules.

saschanaz:
Contributions are rare here, so maybe just can be manually checked.

PR #609, opened 2018-11-09
Add missing HTML tags

saschanaz:
Fixes Microsoft/TypeScript#17828

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -17176,6 +17176,7 @@ interface HTMLElementTagNameMap {
     "b": HTMLElement;
     "base": HTMLBaseElement;
     "basefont": HTMLBaseFontElement;
+    "bdi": HTMLElement;
     "bdo": HTMLElement;
     "blockquote": HTMLQuoteElement;
     "body": HTMLBodyElement;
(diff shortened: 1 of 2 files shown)


falsandtru:
@sandersn Merge this plz. I need this patch.

inad9300:
It's not *too* bad to merge this... since it is an improvement. But it is not yet close to fixing https://github.com/Microsoft/TypeScript/issues/17828:
- The HTML elements `details` and `dialog` haven't been included.
- The new HTML interfaces added by @arjunyel (`HTMLMainElement`, `HTMLDetailsElement`, etc.), as mentioned in the comments, have been ignored.
- The SVG elements mentioned in the comments (`a`, `script`, `style`, and `title`) haven't been considered.

In my opinion, these changes should be reverted until every point is tackled, or else it will be extra work to verify what is and what isn't done.

saschanaz:
Hi @inad9300, those HTML tags are covered by other PRs and included in master branch: https://raw.githubusercontent.com/Microsoft/TSJS-lib-generator/master/baselines/dom.generated.d.ts

The SVG elements is related with https://github.com/Microsoft/TSJS-lib-generator/issues/582#issuecomment-430437141. We currently can't add them because of the blocker.

inad9300:
Hi @saschanaz. Thanks for following up this issue and for finally resolving the problem with `ElementTagNameMap`. Nevertheless, please do use the more specific interfaces for the newly added elements, such as `HTMLMainElement` and the others.

saschanaz:
> please do use the more specific interfaces for the newly added elements, such as HTMLMainElement and the others.

There is no HTMLMainElement in the standard. The interface in current lib.d.ts should be removed instead.

inad9300:
Great then! Sorry for the inconvenience caused.

PR #612, opened 2018-11-12
Warn that `inputfiles/idl/` should not be touched manually.

saschanaz:
cc: @DanielRosenwasser


HolgerJeromin:
We could add a template, too:

https://help.github.com/articles/about-issue-and-pull-request-templates/

MattiasBuelens:
Indeed, everything in that directory is generated... except for `Streams.widl`. 🙈 That file was manually authored in #541 and is explicitly skipped with a `local` flag in `idlSources.json`, see cc52c698311659f481fa9fe47ca2e5f14e204b4f. Hopefully the Streams spec will switch to Web IDL "soon", then we can get rid of that nasty hack.

However, I still agree that we should recommend not touching anything in that directory. 🙂

PR #617, opened 2018-11-14
Unwrap Promise from param types

saschanaz:
Fixes Microsoft/TypeScript#17804

Because the Web IDL spec says [JS->IDL conversion internally uses `Promise.resolve()`](https://heycam.github.io/webidl/#es-promise).

>An ECMAScript value V is converted to an IDL PromiseT value as follows:
>1. Let promise be ! Call(%Promise_resolve%, %Promise%, «V»).
>2. Return the IDL promise type value that is a reference to the same object as promise.

### DIFF
--- a/src/emitter.ts
+++ b/src/emitter.ts
@@ -501,7 +501,7 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
                 p = { name: p.name, type: [p.subtype!, p] }
             }
             const isOptional = !p.variadic && p.optional;
-            const pType = convertDomTypeToNullableTsType(p);
+            const pType = isOptional ? convertDomTypeToTsType(p) : convertDomTypeToNullableTsType(p);
             const variadicParams = p.variadic && pType.indexOf('|') !== -1;
             return (p.variadic ? "..." : "") +
                 adjustParamName(p.name) +


src/emitter.ts
```diff
@@ -497,8 +497,11 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
     /// Generate the parameters string for function signatures
     function paramsToString(ps: Browser.Param[]) {
         function paramToString(p: Browser.Param) {
+            if (p.type === "Promise" && !Array.isArray(p.subtype)) {
```
sandersn: what does it mean when `Array.isArray(p.subtype)` and why don't we do anything in that case?

src/emitter.ts
```diff
@@ -497,8 +497,11 @@ export function emitWebIDl(webidl: Browser.WebIdl, flavor: Flavor) {
...
+                p = { name: p.name, type: [p.subtype!, p] }
+            }
             const isOptional = !p.variadic && p.optional;
-            const pType = isOptional ? convertDomTypeToTsType(p) : convertDomTypeToNullableTsType(p);
```
sandersn: I don't understand what change this is expected to have on *.generated.d.ts. Is it the change of `Promise<any>` to `any`? If so, why?

If there is no change in generated files, then is this just extra cleanup?

src/emitter.ts
saschanaz: `p.subtype` being an array means it's `Promise<TypeA, TypeB>`, which will never happen. This check is mainly to not treat `p.subtype` as a union in this block.

src/emitter.ts
saschanaz: Yes, this is intended to be an extra cleanup, but now I think it's better to just remove `convertDomTypeToNullableTsType` because `convertDomTypeToTsType` now also can process nullable types.

saschanaz: Reverted this line and moved it to #641.

PR #622, opened 2018-11-24
Make Response json generic

### LINKED ISSUE #458: Add a generic type argument to `fetch()` function and `Body` interface
### **Expected behavior (example):**
**Generated definitions:**
```ts
interface Body<T = any> {
    readonly bodyUsed: boolean;
    arrayBuffer(): Promise<ArrayBuffer>;
    blob(): Promise<Blob>;
    formData(): Promise<FormData>;
    json(): Promise<T>;
    text(): Promise<string>;
}

interface Request<T = any> extends Body<T> {...}
interface Response<T = any> extends Body<T> {...}
...

interface GlobalFetch {
    fetch<T>(input?: Request | string, init?: RequestInit): Promise<Response<T>>;
}
```

**Usage with type:**
```ts
type MyFetchResponse = { a: string, b: number };
fetch<MyFetchResponse>('https://someserver.com/get?a=string&b=number').then(response => {
  response.json().then(data => {
    // data is MyFetchResponse

    // Both of these are autocompleted at this point
    data.a
    data.b

   data.c //Error: Property 'c' does not exist on type MyFetchResponse
  })
})
```

**Usage without types (same as current):**
```ts
fetch('https://someserver.com/get?a=string&b=number').then(response => {
  response.json().then(data => {
    // data is any
    data.a
    data.b

   data.c //No error here
  })
})
```

### **Actual behavior:**
```ts
fetch('https://someserver.com/get?a=string&b=number').then(response => {
  response.json().then(data => {
    // data is any
    data.a
    data.b

   data.c //No error here
  })
})
```

###**Motivation:**

The proposed change would make the usage of fetch much more convenient in typed code without the need for unnecessary castings and overrides.

This change will not brake any other definitions relying on `Body`, `Request` and `Response` interfaces due to the use of **optional generic arguments**. Therefore, in current definitions, `declare var a: Response` will automatically evaluate to `declare var a: Response<any>` thus making the change backward-compatible.

iamandrewluca:
Closes #458


sandersn:
@iamandrewluca Sorry for the delay on PRs for this repo. 

Can you give some examples of the new usage? I am by default sceptical of new type parameters that only show up in the return type, since it's not a very good inference position. 

I suspect that the new type parameter is mostly used by explicitly specifying it. In that case I recommend using a cast, since that makes it explicit that you're manually introducing a type.

iamandrewluca:
```typescript
interface DataResponse {
  info: string[]
}

// Example 1
fetch<DataResponse>('/endpoint')
  .then(res => res.json())
  .then(data => data.info)

// Example 2
fetch('/endpoint')
  .then(res => res.json<DataResponse>())
  .then(data => data.info)
```

I understand what you mean, we can do it this way
```typescript
fetch('/endpoint')
  .then(res => res.json())
  .then((data: DataResponse) => data.info)
```

ViceIce:
I think example 1 looks much cleaner to me than the cast.

sandersn:
I think the current state is reasonably convenient and straightforward. Looks like you won't even need a cast most of the time.

@ViceIce Since Typescript doesn't have type information at runtime, an explicit type parameter is basically the same as a cast. You're not really providing more information with an explicit type parameter compared to a cast. The annotation is also more local: the last example @iamandrewluca gives is annotated at exactly the point noImplicitAny would provide an error. If somebody tries to remove the annotation later, the error will appear exactly at that location and they will understand why the annotation was added.

I think we can close this PR. @iamandrewluca do you agree?

iamandrewluca:
@sandersn Yes.

PR #623, opened 2018-12-01
Change AbortSignal's event type to Event

FrogTheFrog:
For some reason AbortSignal has event type of `ProgressEvent`, however I could not find anywhere, where it specifies that it must have `ProgressEvent` interface (might have been a copy-paste typo in https://github.com/Microsoft/TSJS-lib-generator/pull/523/commits/260bdbdcd279c2e89dc7d9be938ef03abcfa28c2). Therefore it it should have `Event` interface for `abort` event.

`ProgressEvent` could lead to errors: [proof (kinda)](http://jsfiddle.net/y5gacx3o/1/).

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1634,7 +1634,7 @@ declare var AbortController: {
 };
 
 interface AbortSignalEventMap {
-    "abort": ProgressEvent;
+    "abort": Event;
 }
 
 interface AbortSignal extends EventTarget {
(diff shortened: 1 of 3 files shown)


mysticatea:
I confirmed that it should be `Event` rather than `ProgressEvent`.

- IDL is here: https://dom.spec.whatwg.org/#interface-AbortSignal  
  The `EventHandler` type is `Event'`s handler.
- "signal abort" algorithm is here: https://dom.spec.whatwg.org/#abortsignal-signal-abort  
  It says "Fire an event named abort at signal."  
  In the "Fire an event" algorithm, the default constructor is `Event`: https://dom.spec.whatwg.org/#concept-event-fire

PR #625, opened 2018-12-02
Fetch web api descriptions from MDN

jwbay:
Ref: https://github.com/Microsoft/TypeScript/issues/26404

Add a script to fetch summary descriptions of interfaces from [MDN](https://developer.mozilla.org/en-US/docs/Web/API), and add the resulting file as an input for emit.


src/mdnfetcher.ts
```diff
@@ -0,0 +1,44 @@
...
+fetchInterfaceDescriptions();
+
+async function fetchInterfaceDescriptions() {
+    const webIdl = require("../inputfiles/browser.webidl.preprocessed.json");
+    const interfaceNames = Object.keys(webIdl.interfaces.interface).sort();
+    const descriptions: Record<string, string> = {};
+
+    await interfaceNames.reduce(async (previousRequest, name) => {
+        // Issuing too many requests in parallel causes 504 gateway errors, so chain
+        await previousRequest;
```
sandersn: why not just make this whole thing synchronous then? (Probably because node-fetch is 500x more popular than the nearest synchronous package)

PR #630, opened 2018-12-19
Remove inappropriate IDBKeyRange usages from lib.dom.d.ts.

mikelehen:
I hit a compile error because IDBCursor.primaryKey was incorrectly typed as
`IDBValidKey | IDBKeyRange` instead of just `IDBValidKey`. I fixed it and
manually audited the rest of the instances of `IDBValidKey | IDBKeyRange` and
fixed them.  If there's a better way to fix these other than manual audit based
on my understanding of the IndexedDB API, let me know.

Fixes https://github.com/Microsoft/TypeScript/issues/29070


mikelehen:
Thanks @saschanaz.  Who would actually merge this? 😁

mikelehen:
Thanks!

PR #632, opened 2018-12-26
Accept null in addition to string in createDataChannel

TomasHubelbauer:
`null` is treated as if an empty string was passed in.

https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel#Parameters

This is my second attempt at Microsoft/TypeScript#29156.

### DIFF
--- a/inputfiles/idl/WebRTC.widl
+++ b/inputfiles/idl/WebRTC.widl
@@ -486,7 +486,7 @@ dictionary RTCTrackEventInit : EventInit {
 
 partial interface RTCPeerConnection {
     readonly        attribute RTCSctpTransport? sctp;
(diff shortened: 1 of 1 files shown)


cpplearner:
It seems that the spec dropped the special treatment to `null` in https://github.com/w3c/webrtc-pc/commit/f30426466cba472463285d385e445d2a617964ec

TomasHubelbauer:
@cpplearner Thanks! I have issued an update to the [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel) instead, hopefully it gets accepted.

PR #633, opened 2018-12-28
Replace ErrorEventHandler with OnErrorEventHandler

### LINKED ISSUE #631: GlobalEventHandlers missing `null` type for `onerror`
GlobalEventHandlers entries all look like

    ondurationchange: ((this: GlobalEventHandlers, ev: Event) => any) | null;

except for `onerror`, which is just `ErrorEventHandler`, without allowing for `null`.  I believe it is supposed to be legal to un-register the error handler by setting `window.onerror = null`.  I'm guessing somebody updated all the other events to be nullable but missed `onerror` because it specifies a callable interface instead of a written-out function signature literal.

ETA: forgot to reference #261 .

saschanaz:
Fixes #631 

`ErrorEventHandler` is a nonstandard name of [`OnErrorEventHandler`](https://html.spec.whatwg.org/multipage/webappapis.html#onerroreventhandler) so use the standard one instead.

cc: @thw0rted @DanielRosenwasser @sandersn


baselines/dom.generated.d.ts
```diff
@@ -5455,7 +5455,7 @@ interface GlobalEventHandlers {
      * Fires when an error occurs during object loading.
      * @param ev The event.
      */
-    onerror: ErrorEventHandler;
+    onerror: OnErrorEventHandler;
```
thw0rted: Shouldn't this be `OnErrorEventHandler | null`?

thw0rted: Ah, sorry, was just looking at the patch and didn't notice the different names.  Does the lib generator have test cases somewhere? Would be nice to be able to add a null assignment somewhere to be sure it compiles.

saschanaz: Unfortunately we currently don't have one, but definitely great to have.

PR #641, opened 2019-01-25
refactor: remove convertDomTypeToNullableTsType() 

saschanaz:
Because `convertDomTypeToTsType` now already internally calls `makeNullable()`.


saschanaz:
Removing paket is okay as it is an F#-era thing.

PR #643, opened 2019-01-26
fix: redefine ElementTagNameMap to fix SVGElementTagNameMap

saschanaz:
cc: @falsandtru and @inad9300


inad9300:
I was expecting this to land in 3.4.x. How come that didn't happen if this feature is merged already?

saschanaz:
@inad9300 This is landed to the master branch via https://github.com/Microsoft/TypeScript/pull/30754, so probably 3.5.

PR #648, opened 2019-01-29
Add WebGL 2 types

saschanaz:
As both Firefox and Chrome have shipped WebGL 2.


saschanaz:
Delaying this until https://github.com/KhronosGroup/WebGL/issues/2216 gets solved.

PR #654, opened 2019-02-01
Narrow declared type of document.body to HTMLBodyElement

Andarist:


saschanaz:
This is wrong as [`document.body` can also be `HTMLFrameSetElement`](https://html.spec.whatwg.org/multipage/dom.html#dom-document-body) (although it would be rare).

Andarist:
@saschanaz thanks for double checking, i've fixed the issue by adding requested type to that declaration

PR #656, opened 2019-02-04
Fixes Microsoft/TypeScript#29620

p85:
Fixes https://github.com/Microsoft/TypeScript/issues/29620


saschanaz:
Updating `idlSources.json` is now preferred over `addedTypes.json`. Would you add the following instead:

```
    {
        "url": "https://www.w3.org/TR/css-will-change/",
        "title": "CSS Will Change"
    },
```

PR #659, opened 2019-02-08
Verbosity replacers

mihailik:
Using a short array of variative `RegExp`s, to match `(the|a)` and `(object|event|interface)` and `(is|represents)`.

* In a simple case, a whole intro lead is just removed.
* In a more complex case, a long intro is replaced with a shorter one.

Replacers are tried in array order, until one succeeds. More tailored replacers can go first, and if those fail -- less clever ones can have a go.

In either case, after replace the first letter of description is re-capitalised again.

It's only 6 replacers, hope easy to review/maintain.


mihailik:
OK, I've scanned through all the changes -- all wording fit and proper. Please review @sandersn @jwbay

saschanaz:
Looks good but I think the function should be in idlfetcher.ts (where `processComments()` exists).

mihailik:
That would not be a good idea, @saschanaz 

Processing string replacement AFTER the docs imported in JSON leaves a useful trace of text before and after string replacement.

While well-tuned to fit future MDN text, the string replacers are not fool-proof. One day they may produce weird results, and having such trace make diagnostics and tweaking process smoother.

----

As I've worked on this PR, it took me a few iterations to get string replacers right. Being able to see before/after states is key to productivity.

There's a lot of text to process, hundreds of replacement matches. With this current implementation, in case of doubt you consult original MDN text sitting in JSON. Otherwise you would have to consult MDN website instead, which can't really scale for 100s of replacers.

saschanaz:
Makes sense. Thanks for clarification.

mihailik:
Ping @sandersn do you agree?

src/index.ts
```diff
@@ -30,6 +30,17 @@ function emitDom() {
     const inputFolder = path.join(__SOURCE_DIRECTORY__, "../", "inputfiles");
     const outputFolder = path.join(__SOURCE_DIRECTORY__, "../", "generated");
 
+    // ${name} will be substituted with the name of an interface
+    const removeVerboseIntroductions: (RegExp | [RegExp, string])[] = [
+        [/^(The|A) ${name} interface of (the\s*)*([a-z\s]+ API)(\\\'s)?/, 'An interface of the $3 '],
+        /^(The|A) ${name} (interface|event|object) (is|represents|represent|describes|defines)?/,
+        /^An object implementing the ${name} interface (is|represents|represent|describes|defines)/,
+        /^The ${name} is an interface representing/,
+        /^This type (is|represents|represent|describes|defines)?/,
```
sandersn: what does `represent` match?

src/index.ts
```diff
@@ -30,6 +30,17 @@ function emitDom() {
     const inputFolder = path.join(__SOURCE_DIRECTORY__, "../", "inputfiles");
     const outputFolder = path.join(__SOURCE_DIRECTORY__, "../", "generated");
 
+    // ${name} will be substituted with the name of an interface
+    const removeVerboseIntroductions: (RegExp | [RegExp, string])[] = [
```
sandersn: I would personally make this `: [RegExp, string][]`; the extra punctuation in the definition seems worth the simplification in transformVerbosity.

I would think differently if (1) this were coming from a file or (2) there were dozens of rules. But I don't think either of those should happen.

src/index.ts
```diff
@@ -70,13 +81,35 @@ function emitDom() {
...
+            if (Array.isArray(regTemplate)) {
+                template = regTemplate[0].source;
+                replace = regTemplate[1];
+            }
+            else {
+                template = regTemplate.source;
+                replace = '';
+            }
+
+            const reg = new RegExp(template.replace(/\$\{name\}/g, name).replace(/\s+/, '\\s*') + '\\s*', 'i');
```
sandersn: Hm. Does the `.replace(/\s+/, '\\s*')` turn out to be necessary? That's a little strange.

src/index.ts
sandersn: Looks like XMLDocument and DOMImplementation have typoes in the source.

src/index.ts
mihailik: Over-cautious against double-spaces. I'll remove and recheck.

src/index.ts
saschanaz: Just fixed those typos on upstream, maybe the json should be updated soon.

mihailik:
Good, applied your changes @sandersn and everything tidy and works. Thanks!

src/index.ts
saschanaz: @mihailik Can `represent` be removed now? XMLDocument and DOMImplementation have no such typo anymore.

src/index.ts
```diff
@@ -30,6 +30,16 @@ function emitDom() {
     const inputFolder = path.join(__SOURCE_DIRECTORY__, "../", "inputfiles");
     const outputFolder = path.join(__SOURCE_DIRECTORY__, "../", "generated");
 
+    // ${name} will be substituted with the name of an interface
+    const removeVerboseIntroductions: [RegExp, string][] = [
+        [/^(The|A) ${name} interface of (the\s*)*([a-z\s]+ API)(\\\'s)?/, 'An interface of the $3 '],
+        [/^(The|A) ${name} (interface|event|object) (is|represents|represent|describes|defines)?/, ''],
+        [/^An object implementing the ${name} interface (is|represents|represent|describes|defines)/, ''],
+        [/^The ${name} is an interface representing/, ''],
+        [/^This type (is|represents|represent|describes|defines)?/, ''],
```
saschanaz: Looks like `represent` is not needed anymore for the latest MDN descriptions.

saschanaz:
@mihailik Can you try running `npm run fetch-mdn` and rebuild? Looks like the latest descriptions need some need additional rules.

mihailik:
Updated:
* removing cautious 'represent' handling of a typo (as fixed in the upstream),
* re-running `fetch-mdn` from upstream.

Works good, and spotted this quirk coming from the original upstream text:
**inputfiles/mdn/apiDescriptions.json**
```JSON
  "IntersectionObserver": "provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport.",
```

The actual MDN page for **<a href="https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API">IntersectionObserver</a>** doesn't have that quirky text, must be an issue with the import?

image

Note: this is not a result of my changes. Just as I referred above, the MDN fetch process and verbosity string replacers are two separate steps. The mangling happened before verbosity, somewhere in scraping the docs from MDN (or in the way MDN exports the docs maybe?)

baselines/dom.generated.d.ts
```diff
@@ -9394,7 +9395,7 @@ declare var ImageData: {
     new(array: Uint8ClampedArray, width: number, height: number): ImageData;
 };
 
-/** An interface of the Intersection Observer API provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport. */
+/** provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport. */
```
mihailik: Beginning of the sentence is lost, because `fetch-mdn` produced mangled line.

inputfiles/mdn/apiDescriptions.json
```diff
@@ -183,7 +184,7 @@
...
   "IIRFilterNode": "The IIRFilterNode interface of the Web Audio API is a AudioNode processor which implements a general infinite impulse response (IIR)  filter; this type of filter can be used to implement tone control devices and graphic equalizers as well. It lets the parameters of the filter response be specified, so that it can be tuned as needed.",
   "ImageData": "The ImageData interface represents the underlying pixel data of an area of a <canvas> element. It is created using the ImageData() constructor or creator methods on the CanvasRenderingContext2D object associated with a canvas: createImageData() and getImageData(). It can also be used to set a part of the canvas by using putImageData().",
-  "IntersectionObserver": "The IntersectionObserver interface of the Intersection Observer API provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport.",
+  "IntersectionObserver": "provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport.",
```
mihailik: Here the results coming from `mdn-fetch` are mangled, the very beginning of the sentence is lost.

See <a href="https://github.com/Microsoft/TSJS-lib-generator/pull/659/commits/d12ae599a8ba35075690c177094e4bb9466ac65d#r257345851">a comment above</a>

mihailik:
@saschanaz the examples you've mentioned seem fine. Text is meaningful, and fluent enough.

However `IntersectionObserver` is mangled by `fetch-mdn` it seems.

src/index.ts
saschanaz: @mihailik We still have one `represent`, is it still needed?

saschanaz:
>the examples you've mentioned seem fine.

CryptoKey currently says:

>The CryptoKey dictionary of the Web Crypto API represents a cryptographic key.

I thought you would be prefer `A cryptographic key.` instead to remove redundant things.

mihailik:
Wiped one more missed `'represent'` (thanks @sandersn), but decided against updating the rules to handle `CryptoKey` (sorry @saschanaz).

`CryptoKey` is a good catch, I haven't noticed it earlier. However it should be left as is, here's why. The 'API' replacer looks like this:

`[/^(The|A) ${name} interface of (the\s*)*([a-z\s]+ API)(\\\'s)?/, 'An interface of the $3 ']`

If `CryptoKey` were referred to as an *interface*, the conversion would have been:

`The CryptoKey interface of the Web Crypto API represents a cryptographic key.` -->
`An interface of the Web Crypto API represents a cryptographic key.`

As you can see, it's not meant to convert into `A cryptographic key.` anyway, I do want to keep the name of the API too. The issue with `CryptoKey` being **a** dictionary not **an** interface is the article. The same replacer can't deal with both 'interface' and 'dictionary' in a kind of polymorphic way, because it need to coordinate articles.

We can introduce another replacer, but it would only target 2 cases: `CryptoKey` and `CryptoKeyPair`. This to me looks to much fine-tuning for now. I want to lay some groundwork, do replacements for some broad rules, and enable fine-tuning for further PRs, if needed.

saschanaz:
So we had some mistakes there:

1. CryptoKey is an interface, not a dictionary. Looks like someone in MDN made a mistake, but I just fixed it.
2. CryptoKeyPair is indeed a dictionary but TSJS thinks it's an interface. We should fix it.

Thanks for clarification 😀

src/index.ts
HolgerJeromin: I would keep it in this code. This can slip in at any time again in MDN.

saschanaz: We can fix typos on MDN in any time. Better to fix on MDN rather than to workaround on TSJS.

sandersn:
Note that I merged from master but did *not* re-run fetch-mdn. The results looked funny so I left it alone.

mihailik:
Thank you, thank you @sandersn!

Sorry I left it ferment for long, hope it brewed into a hearty dish :-)

And thanks for making TypeScript such a useful tool. Good job folks!

sandersn:
No problem. I've only been merging PRs here at the beginning of each release, so it's mostly my fault.

PR #660, opened 2019-02-08
Remove ShadowRoot#innerHTML

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/29820

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -14608,7 +14608,6 @@ declare var ServiceWorkerRegistration: {
 
 interface ShadowRoot extends DocumentFragment, DocumentOrShadowRoot {
     readonly host: Element;
-    innerHTML: string;
     readonly mode: ShadowRootMode;
 }
 
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -1067,17 +1067,6 @@
                     }
                 }
             },
-            "ShadowRoot": {
-                "name": "ShadowRoot",
-                "properties": {
-                    "property": {
-                        "innerHTML": {
-                            "name": "innerHTML",
-                            "type": "DOMString"
-                        }
-                    }
-                }
-            },
             "SpeechRecognition": {
                 "events": {
                     "event": [


saschanaz:
I intentionally kept it when I did #603 because [spec authors wants to properly spec it](https://github.com/w3c/DOM-Parsing/issues/21#issuecomment-436381180).

falsandtru:
I think undocumented proposed APIs shouldn't be added but I'm not sure they should be kept when they are already added.

saschanaz:
@falsandtru IMO undocumented APIs can stay if they are widely implemented. Recently web folks tend to spec those widespread APIs rather than remove it from browsers.

falsandtru:
But they may be removed.

saschanaz:
@falsandtru True. When you are not sure, try opening an issue on the relevant web spec. Then the spec authors will discuss what should be done. If they decide to remove it, IMO that's when TSJS-lib-generator removes the API.

falsandtru:
Just to clarify, I mean "added" means "added to declaration files of TypeScript" at https://github.com/Microsoft/TSJS-lib-generator/pull/660#issuecomment-462034292.

saschanaz:
I understood correctly then 😃

My opinion:

1. It's already in lib.d.ts
2. It's widely implemented
3. The spec authors agree to add it to the specification. That means even if we remove it, we'll have to add it again.

So I think it should stay.

saschanaz:
Hi @falsandtru, the spec just added the official `ShadowRoot.innerHTML` definition! https://github.com/w3c/DOM-Parsing/pull/42 🎉

Please run `npm run fetch-idl -- "DOM Parsing and Serialization"` to refresh the IDL 👍

falsandtru:
Why you didn't update http://w3c.github.io/webcomponents/spec/shadow/?

saschanaz:
@falsandtru Read the warning at the top of the spec, that's why I didn't 😆

`innerHTML` is defined in DOM Parsing and Serialization and so naturally `ShadowRoot.innerHTML` should go there too.

falsandtru:
I saw such extensional definition style for the first time. Confirmed that spec is used to both Element and ShadowRoot.

saschanaz:
Would you mind opening a new PR to remove the addedTypes.json item and replace it by the official IDL?

falsandtru:
I've no plan. Your PR wouldn't conflict.

PR #664, opened 2019-02-23
Added TextEncoder and TextDecoder to window

Mischala:


Mischala:
The types exist, yes. 
However, they are not applied to Window.

As I said in the PR description, calling `new TextEncoder()` works, however, `new window.TextEncoder()` does not. 
This is because they type exists in scope, but has not been applied to the Window object.

PR #667, opened 2019-03-10
Add interface SVGFEDropShadowElement

Fabianopb:
Added interface `SVGFEDropShadowElement` according to [spec](https://developer.mozilla.org/en-US/docs/Web/API/SVGFEDropShadowElement) as requested in https://github.com/Microsoft/TypeScript/issues/29966.

Apparently the interface would have to extend also `SVGFilterPrimitiveStandardAttributes`, but I noticed that the spec of the latter doesn't provide any method, so `setStdDeviation` was manually added.

How does it look? Any thoughts?


inputfiles/addedTypes.json
```diff
@@ -245,6 +245,50 @@
...
+                    "method": {
+                        "setStdDeviation": {
+                            "name": "setStdDeviation",
+                            "override-signatures": [
+                                "setStdDeviation(stdDeviationX: number, stdDeviationY: number): void"
+                            ]
+                        }
+                    }
+                },
+                "extends": "SVGElement"
```
saschanaz:
```
"implements": ["SVGFilterPrimitiveStandardAttributes"]
```

is also needed.

Fabianopb: Thanks, just did it and pushed the new commit!

Fabianopb:
@RyanCavanaugh just noticed you were tagged in the issue (https://github.com/Microsoft/TypeScript/issues/29966), but not here! Could you please check this one, it should be ready to merge IMO.

PR #668, opened 2019-03-14
Remove BroadcastChannelEventMap

goodmind:


goodmind:
There is already one generated from IDL

saschanaz:
Covered and fixed by #647.

PR #672, opened 2019-04-03
Make postMessage's 2nd param, targetOrigin, optional

sandersn:
Fixes Microsoft/TypeScript#30042.


saschanaz:
This is wrong, it is a must.

https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
https://html.spec.whatwg.org/multipage/window-object.html#the-window-object:dom-window-postmessage

sandersn:
Thanks for pointing that out @saschanaz. I got confused by the part of the spec linked from Microsoft/TypeScript#30042 that references the options object, where targetOrigin *is* optional.

sandersn:
Reverted.

PR #676, opened 2019-04-05
Add SVG - Basic Shapes section

sandersn:
Adds detail to the inheritance hierarchy, so that it fixes Microsoft/TypeScript#29573


sandersn:
@saschanaz This is the first widl I've added. Mind taking a look to see if it's right?

saschanaz:
The following should be added for each new element:

```json
                "element": [
                    {
                        "namespace": "SVG",
                        "name": "circle"
                    }
                ],
```

This information is not in IDL so we should manually added it.

sandersn:
OK, done.

PR #680, opened 2019-04-06
Update comment maps

saschanaz:
Applied several fixes:

* Adds comments also for constants and dictionary fields.
* Imports nested comments. (Previously nested ones are just inserted as a raw text.)
* Tracks unmatched comments. (Some specs tend to use their own arbitrary identifier rather than autogenerated one so this PR tracks them.)
* Keeps indentations from inline codes


inputfiles/idl/XMLHttpRequest.commentmap.json
```diff
@@ -1,15 +1,15 @@
...
+    "xmlhttprequest-withcredentials": "True when credentials are to be included in a cross-origin request. False when they are to be excluded in a cross-origin request and when cookies are to be ignored in its response. Initially false.\nWhen set: throws an \"InvalidStateError\" DOMException if state is not unsent or opened, or if the send() flag is set.",
+    "xmlhttprequest-upload": "Returns the associated XMLHttpRequestUpload object. It can be used to gather transmission information when data is transferred to a server.",
+    "xmlhttprequest-send": "Initiates the request. The body argument provides the request body, if any, and is ignored if the request method is GET or HEAD.\nThrows an \"InvalidStateError\" DOMException if either state is not opened or the send() flag is set.",
     "xmlhttprequest-abort": "Cancels any network activity.",
-    "xmlhttprequest-overridemimetype": "Acts as if the `Content-Type` header value for response is mime.\n(It does not actually change the header though.)\nThrows an \"InvalidStateError\" DOMException if state is loading or done.",
```
sandersn: The newlines make reading documentation a lot easier for long comments. I tested VS, VS Code and emacs.

VS and emacs treat newline as newline.
VS Code treats two-spaces-before-newline as newline -- that is, it's interpreting the comment as markdown.

I think the right thing to do is to convert the original newlines into '  \n' so that both approaches will look good.

saschanaz: I did the `innerText` way because the original newlines were for spec authors who edit raw HTML, which tends to look arbitrary when extracted as comments. In the worst case a full line only included a single word, which was perfectly reasonable in raw HTML because it was wrapped by an anchor element.

That said, I fixed a bug that `<p>` didn't insert double newlines. If it still doesn't look good, then I can skip and defer the newline formatting part here.

sandersn: Hmm. I guess it's OK, but maybe we need to have our own wrapping algorithm eventually? The editors I tried don't wrap on their own [1], which becomes pretty hard to read.

[1] VS wraps to the width of the screen, emacs wraps to the width of the window. These are often nearly the same.

saschanaz: >but maybe we need to have our own wrapping algorithm eventually?

Hmm, I'll try it.

PR #683, opened 2019-04-09
Revert "Narrow declared type of document.body to HTMLBodyElement | HT…

sandersn:
…MLFrameSetElement"

This reverts commit 1f7b275407e9a92ea37674b13451d36cf449ba90.

That commit made `document.body` a union type, which in turns makes methods like `addEventListener` unions of overloaded signatures, which do not resolve correctly for the following example code, which I believe is very common:

```ts
document.body.addEventListener('click', (event) => {
    cx = event.pageX;
    cy = event.pageY;
});
```

With the union type, `event: Event`, which does not have `pageX` or `pageY` properties. After reverting this commit, `event: MouseEvent`, which does.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4218,7 +4218,7 @@ interface Document extends Node, NonElementParentNode, DocumentOrShadowRoot, Par
     /**
      * Specifies the beginning and end of the document body.
      */
-    body: HTMLBodyElement | HTMLFrameSetElement;
+    body: HTMLElement;
     /**
      * Returns document's encoding.
      */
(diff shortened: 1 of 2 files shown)


HolgerJeromin:
Why does a click handler on a `HTMLFrameSetElement` does not get a `MouseEvent`?

sandersn:
It does, and so one on a HTMLBodyElement. But 
`document.body : HTMLBodyElement | HTMLFrameSetElement`, 
and `(HTMLBodyElement | HTMLFrameSetElement)['addEventListener']` 
isn't an overloaded signature like `HTMLFrameSetElement['addEventListener']` is. It just has one signature with a callback that passes `event: Event`. So `document.body.addEventListener`'s callback never gets anything but `event: Event`.

PR #684, opened 2019-04-12
General type updates

saschanaz:
Adds `WebGLRenderingContextBase.canvas` and `self.requestAnimationFrame()` for worker, and exposes `External` global constructor.

Had to modify the code to process the special same-named properties `canvas`. Not super pretty but should be good enough for now.


inputfiles/addedTypes.json
```diff
@@ -1075,15 +1075,13 @@
             },
             "Performance": {
                 "name": "Performance",
-                "properties": {
```
sandersn: this is just a bugfix right?

src/widlprocess.ts
```diff
@@ -130,8 +130,19 @@ function convertInterfaceCommon(i: webidl2.InterfaceType | webidl2.InterfaceMixi
...
         }
         else if (member.type === "attribute") {
-            result.properties!.property[member.name] = convertAttribute(member, result.exposed);
-            addComments(result.properties!.property[member.name], commentMap, i.name, member.name);
+            const { properties } = result;
+            const prop = convertAttribute(member, result.exposed);
+            addComments(prop, commentMap, i.name, member.name);
+
+            if (member.name in properties!.namesakes!) {
+                properties!.namesakes![member.name].push(prop);
```
sandersn: it looks like this can happen more than once, but the code in mergeNamesakes assumes there's only one entry in each `namesakes` in line 20: `for (const [prop] of ...`. Is that correct?

saschanaz: Yup, because in line 20 the `namesakes` value is already filtered by exposure (`[Exposed]` in IDL).

saschanaz:
>Questions about the namesakes code. Also, would a name like duplicateProperties or aliasedProperties be more precise?

It's not precisely "duplicated" or "aliased" because those properties are mutually exclusive (only one property should be alive and others will be filtered out, for each lib target):

```webidl
interface X {
  [Exposed=Window] attribute Foo x; // This one will be used in dom.generated.d.ts
  [Exposed=Worker] attribute Bar x; // This one will be used in worker.generated.d.ts
};
```

PR #696, opened 2019-04-18
Add focusin/focusout to GlobalEventHandlers

Fabianopb:
According to https://github.com/Microsoft/TypeScript/issues/30716 focusin/focusout are missing from `GlobalEventHandlersEventMap`.

I have added `onfocusin` and `onfocusout` to `GlobalEventHandlers` and the events `focusin` and `focusout`, which were supposed to be mapped, but they haven't shown up in `GlobalEventHandlersEventMap`, so I could use a small help!

@saschanaz, I've searched in the history and I noticed that in https://github.com/Microsoft/TSJS-lib-generator/pull/674 you've managed to add `deviceorientationabsolute` to `WindowEventMap`. How did you achieve that? I think I'm missing some detail here...

Also, I noticed that `onfocus` has a comment explaining its purpose, would it make sense to include comments also for `onfocusin` and `onfocusout`?

Fixes https://github.com/Microsoft/TypeScript/issues/30716

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5567,6 +5567,8 @@ interface GlobalEventHandlersEventMap {
     "ended": Event;
     "error": ErrorEvent;
     "focus": FocusEvent;
+    "focusin": FocusEvent;
+    "focusout": FocusEvent;
     "gotpointercapture": PointerEvent;
     "input": Event;
     "invalid": Event;
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5728,8 +5728,6 @@ interface GlobalEventHandlers {
      * @param ev The event.
      */
     onfocus: ((this: GlobalEventHandlers, ev: FocusEvent) => any) | null;
-    onfocusin: ((this: GlobalEventHandlers, ev: FocusEvent) => any) | null;
-    onfocusout: ((this: GlobalEventHandlers, ev: FocusEvent) => any) | null;
     ongotpointercapture: ((this: GlobalEventHandlers, ev: PointerEvent) => any) | null;
     oninput: ((this: GlobalEventHandlers, ev: Event) => any) | null;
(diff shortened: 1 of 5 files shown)


saschanaz:
`onfocusin` and `onfocusout` shouldn't be added as those IDL attributes are not implemented in any browsers nor included in any spec.

So those events are without IDL attributes, and currently we only supports events handlers with corresponding attributes, so this goes tricky...

We have to fix the code so that we can add attribute-less events including `DOMContentLoaded`.

Fabianopb:
Good to know @saschanaz , I was not sure, the references around `onfocusin` and `onfocusout` are a bit uncertain, even though [MDN points them as event handler properties](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusin_event), you can't really find any spec.

Thanks for the clarification, so apparently the work is around [these lines of the emitter](https://github.com/Microsoft/TSJS-lib-generator/blob/master/src/emitter.ts#L882). I'll revert the added properties and try to work something out in the emitter.

Fabianopb:
@saschanaz I have now updated the PR to allow adding `"attributeless-events"` to interfaces, which are mapped to the `-EventMap`s, but we have to still figure out the IDL checks (reason why tests are failing). Any tips for that? Also, why do tests pass locally but break in the CI?

saschanaz:
The master branch is currently failing (probably by TypeScript nightly behavior change). Try fixing line 200 of `widlprocess.ts` to `name: operation.name || undefined`.

Fabianopb:
Thanks @saschanaz, Ci fixed! I'll remove the "WIP" from the title, this PR is good to be reviewed!

Fabianopb:
Good point @saschanaz, and thanks for the approval! Ping @RyanCavanaugh as this is probably ready to be merged.

Fabianopb:
Keeping the PR alive, ping perhaps @sandersn ? :)

PR #698, opened 2019-04-24
Add WebGL 2 types

saschanaz:
For https://github.com/Microsoft/TypeScript/issues/30998.

Depends on #684 because of canvas problem.


saschanaz:
Did a quick search and found this: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webgl2/index.d.ts

PR #699, opened 2019-04-28
Emit iterable arguments for ES2015+

saschanaz:
Closes https://github.com/Microsoft/TypeScript/issues/28198.

### DIFF
--- a/src/emitter.ts
+++ b/src/emitter.ts
@@ -689,7 +689,7 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
 
     /// Emit the properties and methods of a given interface
     function emitMembers(prefix: string, emitScope: EmitScope, i: Browser.Interface) {
-        const conflictedMembers = extendConflictsBaseTypes[i.name] ? extendConflictsBaseTypes[i.name].memberNames : new Set();
+        const conflictedMembers = extendConflictsBaseTypes[i.name] ? extendConflictsBaseTypes[i.name].memberNames : new Set<string>();
         emitProperties(prefix, emitScope, i);
         const methodPrefix = prefix.startsWith("declare var") ? "declare function " : "";
         emitMethods(methodPrefix, emitScope, i, conflictedMembers);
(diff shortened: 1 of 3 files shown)


src/emitter.ts
```diff
@@ -1207,17 +1210,57 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
...
+            function typeIncludesSequence(type: string | Browser.Typed[]): boolean {
+                if (Array.isArray(type)) {
+                    return type.some(t => typeIncludesSequence(t.type))
+                }
+                return type === "sequence" || !!sequenceTypedefMap[type];
+            }
+            return !!s.param && s.param.some(p => typeIncludesSequence(p.type));
+        }
+
+        function filterSignaturesWithSequence(signature: Browser.Signature[]): Browser.Signature[] {
```
sandersn: nit: can you name the function `replaceSignaturesWithSequence` and the parameter `signatures`?

src/emitter.ts
```diff
@@ -1207,17 +1210,57 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
...
+            return signature.filter(hasSequenceArgument).map(s => {
+                const newParams = s.param!.map(p => {
```
sandersn: similarly, `params` would be an easier name to read, although that would add a couple dozen lines of code changes.

src/emitter.ts
```diff
@@ -1207,17 +1210,57 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
...
+                    const typedef = typeof p.type === "string" ? sequenceTypedefMap[p.type]: undefined;
```
sandersn: you could drop lines 1227-1229 if you use `p.type` instead of `undefined`. It's less efficient but that's not a concern here.

```suggestion
                    const typedef = typeof p.type === "string" ? sequenceTypedefMap[p.type] : p.type;
```

saschanaz: Doing this will break code as `sequenceTypedefMap[p.type]` also can be `undefined`.

sandersn: That makes sense.

PR #700, opened 2019-05-12
Update Navigator type

jeffryang24:
**Changes**

- Add "HTML - System state and capabilities" spec from https://html.spec.whatwg.org/multipage/system-state.html.
- Update some baseline files.
- Add new generated webidl files.
- Fix https://github.com/microsoft/TypeScript/issues/31358


jeffryang24:
> There are a few MS-specific interfaces that are removed in this PR. Can you add them back manually unless you are sure that nobody is using them?

I will try to add these specific interfaces back to the files. Sorry for the late response and thanks for notifying it... :+1:

jeffryang24:
hmm... why the test still failed while my local success...? :thinking:

sandersn:
It's because dom.generated.d.ts is generated from various widl files plus some overriding jsons. You need to add `MSFileSaver` and MSNavigatorDoNotTrack` to `inputfiles/addedTypes.json` and then add them to `Navigator` in `inputfiles/overridingTypes.json`.

Here's an example of the first:

```json
    "interfaces": {
        "interface": {
            "MSFileSaver": {
                "name": "MSFileSaver",
                "methods": {
                    "method": {
                        "msSaveBlob": {
                            "name": "msSaveBlob",
                            "override-signatures": [
                                "msSaveBlob(blob: any, defaultName?: string): boolean"
                            ]
                        }
// more methods here ...
                    }
                },
                "no-interface-object": 1
            },
```

Here's what I've got in overridingTypes.json for Navigator, but it's not working properly yet:

```json
        "interface": {
            "Navigator": {
                "name": "Navigator",
                "extends": ["MSFileSaver", "MSNavigatorDoNotTrack"]
            },
```

I'll keep looking at the code and update when I figure it out. @saschanaz any suggestions?

saschanaz:
You don't need to manually rewrite `MSFileSaver`, just do `implements: ["MSFileSaver", "MSNavigatorDoNotTrack"]` and it should automatically augment to the `extends` list and reuse existing types.

inputfiles/idlSources.json
```diff
@@ -514,5 +514,9 @@
     {
         "url": "https://html.spec.whatwg.org/multipage/web-messaging.html",
         "title": "Web Messaging"
+    },
+    {
+        "url": "https://html.spec.whatwg.org/multipage/system-state.html",
+        "title": "HTML - System state and capabilities"
```
saschanaz: Could you sort this in alphabet order?

jeffryang24: Ok, will do...

jeffryang24:
> You don't need to manually rewrite `MSFileSaver`, just do `implements: ["MSFileSaver", "MSNavigatorDoNotTrack"]` and it should automatically augment to the `extends` list and reuse existing types.

Hmm... what's different between `extends` and `implements`? For example below section: 
```json
"SVGFEDropShadowElement": {
                "name": "SVGFEDropShadowElement",
                "exposed": "Window",
                "properties": {
                    "property": {
                        "in1": {
                            "name": "in1",
                            "read-only": 1,
                            "override-type": "SVGAnimatedString"
                        },
                        "dx": {
                            "name": "dx",
                            "read-only": 1,
                            "override-type": "SVGAnimatedNumber"
                        },
                        "dy": {
                            "name": "dy",
                            "read-only": 1,
                            "override-type": "SVGAnimatedNumber"
                        },
                        "stdDeviationX": {
                            "name": "stdDeviationX",
                            "read-only": 1,
                            "override-type": "SVGAnimatedNumber"
                        },
                        "stdDeviationY": {
                            "name": "stdDeviationY",
                            "read-only": 1,
                            "override-type": "SVGAnimatedNumber"
                        }
                    }
                },
                "methods": {
                    "method": {
                        "setStdDeviation": {
                            "name": "setStdDeviation",
                            "override-signatures": [
                                "setStdDeviation(stdDeviationX: number, stdDeviationY: number): void"
                            ]
                        }
                    }
                },
                "extends": "SVGElement",
                "implements": [
                    "SVGFilterPrimitiveStandardAttributes"
                ]
            },
```

The output is: `interface SVGFEDropShadowElement extends SVGElement, SVGFilterPrimitiveStandardAttributes`.

saschanaz:
The `extends` in WebIDL means it "inherits" from the target interface: `SVGFEDropShadowElement.__proto__.prototype === SVGElement.prototype`, while `implements` means it just includes their members. It's really a WebIDL thing rather than TypeScript.

jeffryang24:
Hmm... My local test already success, but why it failed again..? :thinking:

saschanaz:
Hmm, it seems you should do rebase or merge from master branch.

jeffryang24:
> Hmm, it seems you should do rebase or merge from master branch.

Ok, will merge from master first...

jeffryang24:
Currently already up to date with master.

inputfiles/addedTypes.json
```diff
@@ -2207,6 +2207,150 @@
                         }
                     }
                 }
+            },
+            "MSFileSaver": {
```
saschanaz: All changes here in addedTypes.json looks redundant. `implements` in `overridingTypes.json` should do the things automatically.

jeffryang24: All the added interfaces in addedTypes? such as MSFileSaver, MSNavigatorDoNotTrack, ExceptionInformation, etc.?

saschanaz: Yup. They were once removed from the result because `implements` does not included them (so the code thought they were needless), but as you added them again via `overridingTypes.json`, now those should be kept intact without these manual typings.

jeffryang24: But, some interfaces like `MSLaunchUriCallback`, `ExceptionInformation`, `ConfirmSiteSpecificExceptionsInformation`, `StoreExceptionsInformation`, and `StoreSiteSpecificExceptionsInformation` have been deleted, maybe I need to add it back manually through `addedTypes.json`.

saschanaz: That's because we have a type tracker that auto-deletes unreferenced/unexposed dictionaries and interfaces. Try removing them building again, and you'll see those types won't be removed again 😁

saschanaz:
The test failure seems to be from an ordering problem... What environment are you on? OS and node.js version, etc.

inputfiles/addedTypes.json
```diff
@@ -2207,6 +2207,20 @@
                         }
                     }
                 }
+            },
+            "MSLaunchUriCallback": {
+                "name": "MSLaunchUriCallback",
```
saschanaz: Ah, this is a problem. Could you remove this and add `MSLaunchUriCallback` to "Window" field of knownTypes.json?

jeffryang24:
> The test failure seems to be from an ordering problem... What environment are you on? OS and node.js version, etc.

OS: Office Notebook => Ubuntu 18.04 LTS (amd64), Personal Notebook (current): Manjaro Illyria 18.0.4 (amd64)
NodeJS: Office Notebook => lts/carbon (v8.15.1), Personal Notebook (current): lts/dubnium (v10.16.0)

inputfiles/addedTypes.json
jeffryang24: Ok, sure.. will mve it to knownTypes.json...

saschanaz:
I'm not sure what's the core problem yet but confirmed that we have a compatibility problem with Node.js 8/10. Updating to Node.js 12 fixes the problem.

jeffryang24:
> I'm not sure what's the core problem yet but confirmed that we have a compatibility problem with Node.js 8/10. Updating to Node.js 12 fixes the problem.

I see... will switch node to v12... :+1:

baselines/dom.generated.d.ts
```diff
@@ -10805,30 +10817,21 @@ declare var NavigationPreloadManager: {
 };
 
 /** The state and the identity of the user agent. It allows scripts to query it and to register themselves to carry on some activities. */
-interface Navigator extends NavigatorID, NavigatorOnLine, NavigatorContentUtils, NavigatorStorageUtils, MSNavigatorDoNotTrack, MSFileSaver, NavigatorBeacon, NavigatorConcurrentHardware, NavigatorUserMedia, NavigatorLanguage, NavigatorStorage, NavigatorAutomationInformation {
-    readonly activeVRDisplays: ReadonlyArray<VRDisplay>;
```
saschanaz: We should restore `activeVRDisplays`/`getVRDisplays()` and `NavigatorBeacon`. Looks good otherwise.

PR #701, opened 2019-05-21
Add null types to optional IDB query parameters

philipwalton:
Fixes https://github.com/microsoft/TypeScript/issues/31490

This PR adds `null` as a valid type to all IndexedDB API methods that accept an optional query param followed by another argument.

It also standardizes on the name "query" everywhere (previously it was sometimes called "query" and sometimes called "range"), as "query" is what's used in [Web IDL in the spec](https://w3c.github.io/IndexedDB/#object-store-interface).

*__Note__: I was getting an error when running the build step that appears to be unrelated to my changes (it fails on master as well). I'm not sure what (if anything) I should do about that.*


sandersn:
You should be able to merge from master to fix this build error.

PR #704, opened 2019-06-03
Restore FileReaderProgressEvent

sandersn:
This type doesn't exist, but it did before about a year ago, and is useful until we parametrise ProgressEvent with subtype properties.

Also fixes variable names in `merge`, which had source and target backward, making for some confusing reading.

Fixes microsoft/Typescript#25510

### DIFF
--- a/src/emitter.ts
+++ b/src/emitter.ts
@@ -153,7 +153,7 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
     /// Interface name to its related eventhandler name list map
     /// Note:
     /// In the xml file, each event handler has
-    /// 1. eventhanlder name: "onready", "onabort" etc.
+    /// 1. eventhandler name: "onready", "onabort" etc.
     /// 2. the event name that it handles: "ready", "SVGAbort" etc.
     /// And they don't just differ by an "on" prefix!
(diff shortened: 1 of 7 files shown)


sandersn:
@saschanaz You might be interested in this one.

sandersn:
@andrewbranch @rbuckton I'm adding you as reviewers so that somebody else on the team is aware of TSJS-lib-generator changes.

saschanaz:
Well, why not just parameterize it?

src/emitter.ts
```diff
@@ -153,7 +153,7 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
     /// Interface name to its related eventhandler name list map
     /// Note:
     /// In the xml file, each event handler has
-    /// 1. eventhanlder name: "onready", "onabort" etc.
+    /// 1. eventhandler name: "onready", "onabort" etc.
```
andrewbranch: 💯

sandersn:
@saschanaz This is easier for now until I understand the compatibility concerns of parametrising ProgressEvent. This definitely works the way people expect because that's the way it used to work.

fatcerberus:
What is meant by "parametrise ProgressEvent with subtype properties"?  Making it into a discriminated union or something...?

sandersn:
@fatcerberus See #707 for @saschanaz' implementation.

PR #707, opened 2019-06-05
Parameterize ProgressEvent

saschanaz:
@sandersn, how about this?

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -2174,21 +2174,6 @@
                         }
                     }
                 }
-            },
-            "FileReaderProgressEvent": {
-                "name": "FileReaderProgressEvent",
-                "extends": "ProgressEvent",
-                "no-interface-object": 1,
-                "properties": {
-                    "property": {
-                        "target": {
-                            "name": "target",
-                            "read-only": 1,
-                            "nullable": 1,
-                            "type": "FileReader"
-                        }
-                    }
-                }
             }
         }
     },
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -599,8 +599,9 @@
                     }
                 }
             },
-            "NodeListOf<TNode extends Node>": {
-                "name": "NodeListOf<TNode extends Node>",
+            "NodeListOf": {
+                "name": "NodeListOf",
+                "generic": "TNode extends Node",
                 "exposed": "Window",
                 "extends": "NodeList",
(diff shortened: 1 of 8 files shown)


sandersn:
Is this useful anywhere else? I expected that lots of other event types would want to have custom targets?

saschanaz:
Might be better to start rather than doing all in once. IMO this will at least prevent a future breaking change "Where is my FileReaderProgressEvent" as we already saw one.

sandersn:
If this is a start that will have some follow-up, I want to understand what the follow-up will be. Are there other subtypes of ProgressEvent that will provide different arguments to `T`?

If the only subtype of ProgressEvent is FileReader, then I think it's too expensive and complex to create a generic type for just one case.

saschanaz:
>Are there other subtypes of ProgressEvent

Current users of ProgressEvent in lib.d.ts include ApplicationCache, Document, GlobalEventHandlers, and XMLHttpRequest,

sandersn:
How would ApplicationCache customise ProgressEvent.target? Would `T=ApplicationCache` ?

saschanaz:
Yep, and `T=XMLHttpRequest`.

This can be generalized for nearly all Event sub-interfaces, so we even could do some automation in the future rather than manual `generic` field.

sandersn:
OK, that makes sense. Can you make the change for ApplicationCache, Document, GlobalEventHandlers, and XMLHttpRequest in this PR?

src/emitter.ts
```diff
@@ -257,7 +268,7 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
                 return event.type;
```
sandersn: why doesn't this one need a call to `getGenericEventType`?

saschanaz: They are not affected because they are currently all using FocusEvent. Might be useful in the future though.

inputfiles/addedTypes.json
```diff
@@ -634,8 +635,9 @@
                     ]
                 }
             },
-            "HTMLCollectionOf<T extends Element>": {
-                "name": "HTMLCollectionOf<T extends Element>",
+            "HTMLCollectionOf": {
+                "name": "HTMLCollectionOf",
+                "generic": "T extends Element",
```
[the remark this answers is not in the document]
saschanaz: Just found that we already have `type-parameters` so I refactored this PR to use it.

PR #714, opened 2019-06-11
Add CSS Overflow types

saschanaz:
Adds `webkitLineClamp` support which is in Chrome and Edge.


sandersn:
I fixed a couple of warnings (from the new redundancy checking code) in the latest commit.

PR #715, opened 2019-06-12
Use globalThis for .window/.self

saschanaz:
Should fix some DefinitelyTyped errors in https://github.com/microsoft/TSJS-lib-generator/pull/452#issuecomment-501029800.

Closes https://github.com/microsoft/TypeScript/issues/19816

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -18478,11 +18478,6 @@ interface WindowEventMap extends GlobalEventHandlersEventMap, WindowEventHandler
 
 /** A window containing a DOM document; the document property points to the DOM document loaded in that window. */
 interface Window extends EventTarget, WindowTimers, WindowSessionStorage, WindowLocalStorage, WindowConsole, GlobalEventHandlers, IDBEnvironment, WindowBase64, AnimationFrameProvider, WindowOrWorkerGlobalScope, WindowEventHandlers {
-    Blob: typeof Blob;
-    TextDecoder: typeof TextDecoder;
-    TextEncoder: typeof TextEncoder;
-    URL: typeof URL;
-    URLSearchParams: typeof URLSearchParams;
     readonly applicationCache: ApplicationCache;
     readonly caches: CacheStorage;
     readonly clientInformation: Navigator;
@@ -18562,14 +18557,14 @@ interface Window extends EventTarget, WindowTimers, WindowSessionStorage, Window
     readonly scrollX: number;
     readonly scrollY: number;
     readonly scrollbars: BarProp;
-    readonly self: Window;
+    readonly self: Window & typeof globalThis;
     readonly speechSynthesis: SpeechSynthesis;
     status: string;
     readonly statusbar: BarProp;
     readonly styleMedia: StyleMedia;
     readonly toolbar: BarProp;
     readonly top: Window;
-    readonly window: Window;
+    readonly window: Window & typeof globalThis;
     alert(message?: any): void;
     blur(): void;
     /** @deprecated */
@@ -19472,11 +19467,6 @@ declare var Image: {
 declare var Option: {
     new(text?: string, value?: string, defaultSelected?: boolean, selected?: boolean): HTMLOptionElement;
 };
-declare var Blob: typeof Blob;
-declare var TextDecoder: typeof TextDecoder;
-declare var TextEncoder: typeof TextEncoder;
-declare var URL: typeof URL;
-declare var URLSearchParams: typeof URLSearchParams;
 declare var applicationCache: ApplicationCache;
 declare var caches: CacheStorage;
 declare var clientInformation: Navigator;
@@ -19556,14 +19546,14 @@ declare var screenY: number;
 declare var scrollX: number;
 declare var scrollY: number;
 declare var scrollbars: BarProp;
-declare var self: Window;
+declare var self: Window & typeof globalThis;
 declare var speechSynthesis: SpeechSynthesis;
 declare var status: string;
 declare var statusbar: BarProp;
 declare var styleMedia: StyleMedia;
 declare var toolbar: BarProp;
 declare var top: Window;
-declare var window: Window;
+declare var window: Window & typeof globalThis;
 declare function alert(message?: any): void;
 declare function blur(): void;
 /** @deprecated */
--- a/baselines/webworker.generated.d.ts
+++ b/baselines/webworker.generated.d.ts
@@ -5361,7 +5361,7 @@ interface WorkerGlobalScope extends EventTarget, WorkerUtils, WindowConsole, Win
     readonly location: WorkerLocation;
     onerror: ((this: WorkerGlobalScope, ev: ErrorEvent) => any) | null;
     readonly performance: Performance;
-    readonly self: WorkerGlobalScope;
+    readonly self: WorkerGlobalScope & typeof globalThis;
     msWriteProfilerMark(profilerMarkName: string): void;
     addEventListener<K extends keyof WorkerGlobalScopeEventMap>(type: K, listener: (this: WorkerGlobalScope, ev: WorkerGlobalScopeEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
     addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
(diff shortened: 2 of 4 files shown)


sandersn:
I'll need to test this on our tests and on Definitely Typed as well, but it's a good change.

Let's hold #452 until this PR is in.

sandersn:
Errors from Typescript tests: without #452, cancelAnimationFrame, setTimeout and fetch are duplicated. We'll need to make sure that this is not the case after both PRs are in.

I'm running user tests right now.

sandersn:
User tests look pretty good; there are lots of diffs in chrome-devtools, but they seem to be good. That project uses lots of `window.x` references to refer to globals.

saschanaz:
>Errors from Typescript tests: without #452, cancelAnimationFrame, setTimeout and fetch are duplicated.

Only those three? Currently we expose all members of `Window` to global (to allow both `window.performance` and `performance`) so `Window & typeof globalThis` can have duplicated members, but this way other members should get the same errors too.

saschanaz:
Turns out we don't need `Window &` part because we already add all the members to `globalThis`.

sandersn:
That should take care of the duplicates. Here are the errors from definitely typed, although I'll re-run with your new changes.

I'm not sure why dojo runs out of memory. It doesn't when I just run `tsc` so I guess it's a problem with dtslint-runner.

```
Error in apollo-upload-client
Error: Errors in typescript@local for external dependencies:
node_modules/apollo-link-http-common/lib/index.d.ts(50,13): error TS2304: Cannot find name 'GlobalFetch'.

    at /home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:190:19
    at Generator.next (<anonymous>)
    at fulfilled (/home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:5:58)


Error in dom-inputevent
Error: Errors in typescript@local for external dependencies:
../../../ts/built/local/lib.dom.d.ts(581,5): error TS2687: All declarations of 'isComposing' must have identical modifiers.
../../../ts/built/local/lib.dom.d.ts(9656,11): error TS2300: Duplicate identifier 'InputEvent'.
../../../ts/built/local/lib.dom.d.ts(9662,13): error TS2300: Duplicate identifier 'InputEvent'.

    at /home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:190:19
    at Generator.next (<anonymous>)
    at fulfilled (/home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:5:58)


Error in dojo
Out of memory


Error in jquery
Error: /home/nathansa/DefinitelyTyped/types/jquery/jquery-tests.ts:195:13
ERROR: 195:13  expect  Expected type to be:
  JQuery<Window>
got:
  JQuery<Window & typeof globalThis>

/home/nathansa/DefinitelyTyped/types/jquery/test/jquery-slim-window-module-tests.ts:5:1
ERROR: 5:1     expect  Expected type to be:
  JQuery<Window>
got:
  JQuery<Window & typeof globalThis>

/home/nathansa/DefinitelyTyped/types/jquery/test/jquery-window-module-tests.ts:5:1
ERROR: 5:1     expect  Expected type to be:
  JQuery<Window>
got:
  JQuery<Window & typeof globalThis>

    at /home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:190:19
    at Generator.next (<anonymous>)
    at fulfilled (/home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:5:58)


Error in redux-storage
Error: /home/nathansa/DefinitelyTyped/types/redux-storage/redux-storage-tests.ts:26:63
ERROR: 26:63  expect  TypeScript@local compile error: 
Argument of type 'Reducer<TestState>' is not assignable to parameter of type 'Reducer<unknown>'.
  Types of parameters 'state' and 'state' are incompatible.
    Type 'unknown' is not assignable to type 'TestState'.
ERROR: 36:85  expect  TypeScript@local compile error: 
Argument of type 'Reducer<TestState>' is not assignable to parameter of type 'Reducer<unknown>'.

    at /home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:190:19
    at Generator.next (<anonymous>)
    at fulfilled (/home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:5:58)


Error in webappsec-credential-management
Error: Errors in typescript@local for external dependencies:
../../../ts/built/local/lib.dom.d.ts(3622,11): error TS2300: Duplicate identifier 'Credential'.
../../../ts/built/local/lib.dom.d.ts(3627,13): error TS2300: Duplicate identifier 'Credential'.
../../../ts/built/local/lib.dom.d.ts(10725,14): error TS2687: All declarations of 'credentials' must have identical modifiers.
../../../ts/built/local/lib.dom.d.ts(19853,6): error TS2300: Duplicate identifier 'CredentialMediationRequirement'.

    at /home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:190:19
    at Generator.next (<anonymous>)
    at fulfilled (/home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:5:58)


Error in webgl2
Error: /home/nathansa/DefinitelyTyped/types/webgl2/index.d.ts:582:13
ERROR: 582:13  expect  TypeScript@local compile error: 
Subsequent variable declarations must have the same type.  Variable 'WebGL2RenderingContext' must be of type '{ new (): WebGL2RenderingContext; prototype: WebGL2RenderingContext; readonly ACTIVE_ATTRIBUTES: number; readonly ACTIVE_TEXTURE: number; readonly ACTIVE_UNIFORMS: number; readonly ALIASED_LINE_WIDTH_RANGE: number; ... 554 more ...; readonly WAIT_FAILED: number; }', but here has type '{ new (): WebGL2RenderingContext; prototype: WebGL2RenderingContext; readonly ACTIVE_ATTRIBUTES: number; readonly ACTIVE_TEXTURE: number; readonly ACTIVE_UNIFORMS: number; readonly ALIASED_LINE_WIDTH_RANGE: number; ... 555 more ...; readonly MAX_CLIENT_WAIT_TIMEOUT_WEBGL: number; }'.

    at /home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:190:19
    at Generator.next (<anonymous>)
    at fulfilled (/home/nathansa/dtslint-runner/node_modules/dtslint/bin/index.js:5:58)
The following packages had errors: apollo-upload-client, dom-inputevent, dojo, jquery, redux-storage, webappsec-credential-management, webgl2
FAILED
```

Note that we haven't updated DT for all the DOM changes of the last two months, so it's possible that these failures are not even related to `window` (webgl2 especially)

sandersn:
Quite a few errors after the new commit. I haven't looked at the reason yet.
[errors.txt](https://github.com/microsoft/TSJS-lib-generator/files/3287539/errors.txt)

saschanaz:
Eh. Removing `Window &` part will break augmentation to Window interface:

```ts
// in cordova-plugin-x-socialsharing
interface Window {
    plugins: CordovaPlugins;
}
```

Looks like we should restore `Window &` (and just deal with the duplicates), thoughts?

(Only jQuery error look related with `window` in https://github.com/microsoft/TSJS-lib-generator/pull/715#issuecomment-501790865)

sandersn:
Hm. I think Window augmentation is a well-established pattern, unfortunately. We should address the duplication instead.

Where does the duplication come from, globals that are also members of Window? I assume that `fetch`, `performance` et al can be removed from `Window` if so.

saschanaz:
Things like `var x: Window["performance"]` will fail then, but probably worth trying.

saschanaz:
Okay, I see two options to solve the duplicated function overload problem (that is, only functions are problematic):

* `window: Omit<Window, keyof typeof globalThis> & typeof globalThis`. The type becomes dirty mainly because of `keyof`.
* Empty `Window`: `interface Window {};`. Will break `Window["fetch"]`.

sandersn:
The signature of fetch is generated from the same signature in the widl, isn't it? That makes me think there are two more options:

* keep the duplication, since the signatures will be identical
* change the emitter to not emit signatures in Window that would otherwise be duplicated.

I'll check to see how the checker handles duplicate identical signatures.

saschanaz:
>The signature of fetch is generated from the same signature in the widl

Yes, it is, so it shouldn't be a trouble.

sandersn:
@weswigham does the checker have special handling for multiple signatures arising from an intersection? I thought there was, but I can't find it. As-is, it seems like an intersection with duplicate signatures will take the overload code path everywhere, but should behave like a single intersection in relationship checking, resolveCall, etc.

sandersn:
Update: the checker now has special handling for multiple signatures in an intersection; they are deduped if they are identical. In this case they should be, so I think duplication is the best option.

sandersn:
* lodash, async, debug: One or two bad errors during feature detection are now gone.
* chrome-devtools: weird name resolution pattern needs type annotation now:

```js
var o = self
for (const n of names) {
  o = o[n]
}
```

chrome-devtools uses window property accesses to access globals, so this PR improves it quitea 
 bit.

DT run looks clean. A few ExpectType comments will need to change.

PR #716, opened 2019-06-12
Add CSSOM/Conditional Rules types

saschanaz:
~~*Note: This includes types that should be excluded because of implementation problem. You must first check if there are multiple stable implementations when you open a PR.*~~

~~Anyone who wants to boost merging CSS types might do some cherrypicking from here.~~

Now ready to get a review.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2614,13 +2614,6 @@ declare var CDATASection: {
     new(): CDATASection;
 };
 
-/** Holds useful CSS-related methods. No object with this interface are implemented: it contains only static methods and therefore is a utilitarian interface. */
-interface CSS {
-    escape(value: string): string;
-    supports(property: string, value?: string): boolean;
-}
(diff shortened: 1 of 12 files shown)


orta:
We're gonna run DT on this to make sure that the CSS interface -> namespace change doesn't break things 👍

sandersn:
I opened a PR at microsoft/Typescript#37256. There were 4 failures in DT:

* react-shadow-dom-retarget-events: TODO
* recharts: TODO
* skatejs: TODO
* w3c-css-typed-object-model-level-1: the conflict with `CSS` is expected since this is a polyfill library. We'll have to fix this up, but it's probably OK to just delete it.

(I'll edit this after I investigate the first 3)

sandersn:
In summary, the DT breaks seem expected and fixable except for the construct signature of CustomElementConstructor becoming a call signature. But I don't know why that's related this PR.

sandersn:
Thanks! I'll fix up the DT breaks in the next day or two.

PR #730, opened 2019-06-15
Sort members on dom.iterable.generated.d.ts

saschanaz:
This fixes a compatibility issue with Node.js 8/10 where `.sort()` behaves differently with Node.js 12, by doing proper sort by name.

Thanks @jeffryang24 and sorry for the inconvenience! (https://github.com/microsoft/TSJS-lib-generator/pull/700#issuecomment-502342018)


orta:
Thanks for this, my `npm run test` was failing

PR #737, opened 2019-07-10
Update @types/node and typescript dependency

sandersn:
The new version of @types/node avoids the new generator types introduced recently in typescript@next.


sandersn:
@saschanaz #715 needs this to build. I'll merge from master on that branch after merging this PR.

PR #740, opened 2019-07-17
Fix incorrect caches.match() type definition

philipwalton:
The `CacheStorage.match()` method is correctly defined in the IDL doc, but then the overrides file sets the type to the incorrect value.

See here that the spec uses `MultiCacheQueryOptions` for the CacheStorage interface:
https://w3c.github.io/ServiceWorker/#cachestorage-interface


philipwalton:
Note, some tests are failing, but those tests were failing before I made my changes, and no additional tests were failing after I made my changes.

sandersn:
webworker baselines weren't updated so I updated them.

PR #743, opened 2019-07-24
Duplicate word in TSdoc

massic80:


saschanaz:
That file should not be hand-written, instead it should be generated via `npm run fetch-idl DOM`.

That said, the sentence is from https://dom.spec.whatwg.org/#interface-eventtarget:

>Dispatches a synthetic event _event_ to _target_ and returns true if either _event_’s `cancelable` attribute value is false or its `preventDefault()` method was not invoked, and false otherwise.

Maybe we should do some markdown do support the styling.

massic80:
Aw! Sorry, I knew that there was some automation but I didn't know what was copied and from where :)
Thanks for your explanation.
I'll prune this branch

PR #744, opened 2019-08-01
Override Node previousSibling to be type of ChildNode.

drivron:
To fix / finish off the pull request #619 .

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -691,6 +691,9 @@
                         },
                         "nextSibling": {
                             "override-type": "ChildNode | null"
+                        },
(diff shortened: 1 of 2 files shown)


inputfiles/overridingTypes.json
```diff
@@ -691,6 +691,9 @@
                         },
                         "nextSibling": {
                             "override-type": "ChildNode | null"
+                        },
+                        "previousSibling": {
+									"override-type": "ChildNode | null"
```
HolgerJeromin: Probably the indention should be fixed here.

drivron: Thanks. Fixed.

PR #749, opened 2019-08-31
Makes the `ontouch*` events on window optional.

MicahZoltu:
Browsers without touch support (e.g., Firefox/Chrome on desktop) will not have `ontouch*` events defined on `window`, so the current type definition is not correct.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -6096,10 +6096,10 @@ interface GlobalEventHandlers {
      */
     ontimeupdate: ((this: GlobalEventHandlers, ev: Event) => any) | null;
     ontoggle: ((this: GlobalEventHandlers, ev: Event) => any) | null;
-    ontouchcancel: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
-    ontouchend: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
-    ontouchmove: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
-    ontouchstart: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
+    ontouchcancel?: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
+    ontouchend?: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
+    ontouchmove?: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
+    ontouchstart?: ((this: GlobalEventHandlers, ev: TouchEvent) => any) | null;
     ontransitioncancel: ((this: GlobalEventHandlers, ev: TransitionEvent) => any) | null;
     ontransitionend: ((this: GlobalEventHandlers, ev: TransitionEvent) => any) | null;
     ontransitionrun: ((this: GlobalEventHandlers, ev: TransitionEvent) => any) | null;
@@ -19886,10 +19886,10 @@ declare var onsuspend: ((this: Window, ev: Event) => any) | null;
  */
 declare var ontimeupdate: ((this: Window, ev: Event) => any) | null;
 declare var ontoggle: ((this: Window, ev: Event) => any) | null;
-declare var ontouchcancel: ((this: Window, ev: TouchEvent) => any) | null;
-declare var ontouchend: ((this: Window, ev: TouchEvent) => any) | null;
-declare var ontouchmove: ((this: Window, ev: TouchEvent) => any) | null;
-declare var ontouchstart: ((this: Window, ev: TouchEvent) => any) | null;
+declare var ontouchcancel: ((this: Window, ev: TouchEvent) => any) | null | undefined;
+declare var ontouchend: ((this: Window, ev: TouchEvent) => any) | null | undefined;
+declare var ontouchmove: ((this: Window, ev: TouchEvent) => any) | null | undefined;
+declare var ontouchstart: ((this: Window, ev: TouchEvent) => any) | null | undefined;
 declare var ontransitioncancel: ((this: Window, ev: TransitionEvent) => any) | null;
 declare var ontransitionend: ((this: Window, ev: TransitionEvent) => any) | null;
 declare var ontransitionrun: ((this: Window, ev: TransitionEvent) => any) | null;
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -129,6 +129,18 @@
                     "property": {
                         "onerror": {
                             "override-type": "OnErrorEventHandler"
+                        },
+                        "ontouchcancel": {
+                            "required": 0
+                        },
+                        "ontouchend": {
+                            "required": 0
+                        },
+                        "ontouchmove": {
+                            "required": 0
+                        },
+                        "ontouchstart": {
+                            "required": 0
                         }
                     }
                 }
--- a/src/emitter.ts
+++ b/src/emitter.ts
@@ -617,8 +617,12 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
             else {
                 pType = convertDomTypeToTsType(p);
             }
-            const requiredModifier = p.required === undefined || p.required === 1 ? "" : "?";
+            const required = p.required === undefined || p.required === 1;
+            const requiredModifier = required || prefix ? "" : "?";
             pType = p.nullable ? makeNullable(pType) : pType;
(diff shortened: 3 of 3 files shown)


MicahZoltu:
Fixes #5708

orta:
This makes sense, want to update the snapshots and this can be merged?

MicahZoltu:
How does one "update the snapshots"?  Is that a fancy way of saying "rebase on master"?

MicahZoltu:
Ah, you are probably referring to the instructions here?  https://github.com/microsoft/TSJS-lib-generator#contribution-guidelines

It makes me sad that those aren't done automatically as part of CI/CD.  😢 It means I'll have to actually clone this project to make this PR rather than just doing it in the GitHub UI.

MicahZoltu:
Hmm, digging in on this it seems I need to add an `overridingTypes.json` entry, presumably on the `window` interface.  Is there a way to say, "keep the auto-generated signature, but make it optional" or do I have to provide a copy of the signature with `| undefined` on it?

MicahZoltu:
I have rebased the PR and applied the necessary change to `overridingTypes.json` in what I _think_ is the proper way.  I couldn't figure out how to make something optional other than by adding `| undefined` to the type signature.  It is also worth noting that the global variable `ontouchcancel` will not be defined (different from undefined because JavaScript), yet the TS definition currently has it as `declare var ontouchcancel: ... | null | undefined`, which means if you do `if (ontouchstart === undefined)` TS will compile but you'll get a runtime exception I believe.  I don't know how to resolve this.

inputfiles/overridingTypes.json
```diff
@@ -129,6 +129,18 @@
                     "property": {
                         "onerror": {
                             "override-type": "OnErrorEventHandler"
+                        },
+                        "ontouchcancel": {
+                            "override-type": "((this: Window, ev: TouchEvent) => any) | null | undefined"
```
saschanaz: Try replacing this line with `nullable: 1`.

MicahZoltu: Of course, why didn't I think that `nullable: 1` would change a `nullable` type into a `nullable | undefined` type...

MicahZoltu:
Rebased again and updated to use `"nullable": 1` instead of the manual override.

saschanaz:
You also need to run `npm run build && npm run baseline-accept` 😁

MicahZoltu:
Bleck.  I cannot run `npm baseline-accept` because it doesn't work on Windows.  😢  If a maintainer wants to run that, feel free.  Or just crib this PR and make a new one.  I just want the code fixed, don't care how it happens.

MicahZoltu:
Ah, easy enough to replicate the `baseline-accept` step by hand.  Fixed.

saschanaz:
Could you try `npm i cpx2` and run it again? Because I accidentally broke it 🙄

MicahZoltu:
`npm install cpx2` does appear to resolve the error.  I already have matching baseline, so I'm just going to assume the copy did something.  😉

saschanaz:
This PR now unexpectedly is no-op 🤯

I think what we want to do here is marking it as `ontouchstart?: (whatever)`. We should probably create a new way to do it because AFAIK there is no way to do it.

inputfiles/overridingTypes.json
```diff
@@ -129,6 +129,18 @@
                     "property": {
                         "onerror": {
                             "override-type": "OnErrorEventHandler"
+                        },
+                        "ontouchcancel": {
+                            "nullable": 1
```
saschanaz: I got an ugly working hack: Try replacing this line with `"name": "ontouchcancel?"` and so on.

MicahZoltu: Hmm, I could have sworn that it was not a no-op when I committed...

MicahZoltu: Ewe.... that is dirty... but it appears to work.  Pushed.

baselines/dom.generated.d.ts
```diff
@@ -19886,10 +19886,10 @@ declare var onsuspend: ((this: Window, ev: Event) => any) | null;
  */
 declare var ontimeupdate: ((this: Window, ev: Event) => any) | null;
 declare var ontoggle: ((this: Window, ev: Event) => any) | null;
-declare var ontouchcancel: ((this: Window, ev: TouchEvent) => any) | null;
-declare var ontouchend: ((this: Window, ev: TouchEvent) => any) | null;
-declare var ontouchmove: ((this: Window, ev: TouchEvent) => any) | null;
-declare var ontouchstart: ((this: Window, ev: TouchEvent) => any) | null;
+declare var ontouchcancel?: ((this: Window, ev: TouchEvent) => any) | null;
```
saschanaz: Ah, no, this won't work.... Sorry to make you break code 🙄

MicahZoltu: Should I let this PR lapse until such work is complete, or return to the original solution that will break if the auto-generated types ever change?

saschanaz: Okay, I got a working way. Please fix `src/emitter.ts` line 620 as:

```diff
-            const requiredModifier = p.required === undefined || p.required === 1 ? "" : "?";
+            const requiredModifier = p.required === undefined || p.required === 1 || prefix ? "" : "?";
```

And use `required: 0` on `overridingTypes.json`. I tried it and it works this time!

MicahZoltu: That change doesn't appear to update the global variables:
```typescript
declare var ontouchcancel: ((this: Window, ev: TouchEvent) => any) | null;
declare var ontouchend: ((this: Window, ev: TouchEvent) => any) | null;
declare var ontouchmove: ((this: Window, ev: TouchEvent) => any) | null;
declare var ontouchstart: ((this: Window, ev: TouchEvent) => any) | null;
```
Those need to be `|undefined` as well I believe?
in Firefox if you do `>> ontouchcancel` you get
> ReferenceError: ontouchcancel is not defined

MicahZoltu: Updated PR with your suggested changes in case that is desired behavior.

saschanaz: Ah, you're right. We can modify it further:

```diff
             }
-            const requiredModifier = p.required === undefined || p.required === 1 ? "" : "?";
+            const required = p.required === undefined || p.required === 1;
+            const requiredModifier = required || prefix ? "" : "?";
             pType = p.nullable ? makeNullable(pType) : pType;
+            if (!required && prefix) {
+                pType += " | undefined"
+            }
```

MicahZoltu: Updated.  I look forward to the day when someone git-blames me for this and I have no idea what any of it does.  😛

sandersn:
Do we still need this PR if feature detection code like `(('ontouchstart' in window) || (window.navigator))` works in 3.6 and later?

MicahZoltu:
I'm not familiar with the feature detection code you are referring to, but the current definition of `Window` in `dom` TypeScript lib is _incorrect_.  So unless the feature detection thing you are referring to fixes the bug in the `Window` definition, this PR is still necessary.

saschanaz:
I'm not sure the current types can break things given that the `in` type guard problem is now fixed.

```ts
if (window.ontouchstart === undefined) {
    window.ontouchstart // currently this becomes never
}
```

This is the only weird thing I could find now but it shouldn't break anything.

MicahZoltu:
The current problem is that `window.ontouchstart` type definition isn't a union with `undefined`.  This means that the following code will not give a compile time error, but it should:
```typescript
window.ontouchstart(...) // should error at compile time with calling possibly undefined
```

saschanaz:
> This means that the following code will not give a compile time error

It will, because it's a union with `null` and `null` is not callable.

MicahZoltu:
Ah, good point.  However, in a desktop browser with no touch interface, `window.ontouchcancel` is `undefined`, not null.  In TypeScript 3.7.0-beta, the following will compile (just tested) but I _believe_ it will fail at runtime since `myWindow.ontouchcancel` is `undefined`, not `null`:
```typescript
declare const touchEvent: TouchEvent
declare const myWindow: Window
if (myWindow.ontouchcancel !== null) {
	myWindow.ontouchcancel(touchEvent)
}
```

saschanaz:
Hmmm technically yes... but why would anyone directly call a event listener?

MicahZoltu:
🤷‍♂
Regardless of whether it is _useful_, I think that it should be _correct_.

sandersn:
Seem reasonable to make it correct. If there are unforeseen problems that people run into, we have enough time to revert this change before 3.8.

PR #751, opened 2019-09-05
Map WebIDL `object` to TypeScript `object`

### LINKED ISSUE #290: Emit `object` for IDL `object` type on arguments
For example, `document.scrollIntoView` should receive `object | boolean` while `PaymentResponse.prototype.details` should still be `any` to allow arbitrary property access.

RReverser:
This type on the TS side is relatively new, which is probably why it wasn't used in the first place, but gives more precise definitions and, so, better type checking.

As a side-effect, fixes #290 (although in a different way).

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -130,7 +130,7 @@ interface AudioWorkletNodeOptions extends AudioNodeOptions {
     numberOfOutputs?: number;
     outputChannelCount?: number[];
     parameterData?: Record<string, number>;
-    processorOptions?: any;
+    processorOptions?: object | null;
 }
 
 interface AuthenticationExtensionsClientInputs {
@@ -948,7 +948,7 @@ interface PaymentDetailsInit extends PaymentDetailsBase {
 
 interface PaymentDetailsModifier {
     additionalDisplayItems?: PaymentItem[];
-    data?: any;
+    data?: object;
     supportedMethods: string | string[];
     total?: PaymentItem;
 }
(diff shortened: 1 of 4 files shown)


orta:
@saschanaz - do you have a general opinion about tightening the types like this?

If these are new APIs, I'm OK with tightening it I think

RReverser:
@orta These are not new APIs, but it is a relatively new type in TypeScript, and I think it makes sense to use it as it's more precise than `any` and more correct for these cases (that is, it won't break any valid code, but will catch more mistakes).

RReverser:
I'm starting to think that, instead of an `object`, it would be better to use `Record<string, any>` all around here, as it allows dot-access while still checking that the value is an object and not any other primitive (`string`, `number`, `null`, etc.).

@saschanaz what do you think?

sandersn:
I don't like the readability of `Record<string, any>` compared to `object` or `any`. This is important because VS Code's quick info will show up for JS users as well as TS users.

I vote to just change parameter types, not return types.

sandersn:
I’m going to close this since it’s now kind of stale. 

I’m still not convinced of the value versus readability of `Record<string, any>`, so that’s the first thing we’d need to establish before re-opening.

RReverser:
@sandersn Oh, sorry for never circling back on this. I can definitely see your point about readability.

I wonder if something like this would help?

```typescript
interface AnyObject {
    [key: string]: any;
}
// ...
declare function f(x: AnyObject): AnyObject;
```

This way 1) `f` shows `AnyObject` (name subject to change) in parameters and return types, but still 2) accepts only valid objects and not arbitrary values and 3) allows to get properties on the result and so on.

PR #766, opened 2019-09-30
Add missing `cpx` dependency

RReverser:
It's used by baseline-accept script, but not referenced in dependencies and so fails to run on fresh checkout.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,42 +84,10 @@
         "uri-js": "^4.2.2"
       }
     },
-    "anymatch": {
-      "version": "1.3.2",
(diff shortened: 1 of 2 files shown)


saschanaz:
`cpx` has some unpatched security vulnerabilities, maybe `cpx2` instead?

RReverser:
@saschanaz I don't think it really matters here tbh, but either is fine by me.

saschanaz:
It kinda matters because #742 was to fix the security warning...

RReverser:
As I said, fine by me either way :) Updated PR.

orta:
👍

PR #768, opened 2019-10-02
Random fixes for HTML media and EME

MattiasBuelens:
At work, we found a couple of incorrect and/or missing types while dealing with [HTML media elements](https://html.spec.whatwg.org/multipage/media.html) and [EME](https://www.w3.org/TR/encrypted-media/). This PR fixes some of these types.


saschanaz:
We don't want to add new vendor-prefixed APIs, so I would exclude them here.

MattiasBuelens:
Fair enough. I'll remove that commit.

PR #789, opened 2019-10-16
update types rejection handled

vkrbt:
fixes https://github.com/microsoft/TypeScript/issues/34511


vkrbt:
Hi, I noticed that all check were passed, could this code be merged? I need this fix in my code
Thanks a lot!

saschanaz:
@orta?

orta:
👍 yep, looks like it's correct to the spec

orta:
I think it could be too late to get it into the 3.7 RC though

PR #792, opened 2019-10-20
Updates PermissionName type - Permissions API

maciekgrzybek:
This is PR for updating PermissionName type  - part of Permissions API. The issue can be found [here](https://github.com/microsoft/TypeScript/issues/33923).


saschanaz:
Where did you get the list? Could you file an issue in https://github.com/w3c/permissions/issues so that the standard can be properly updated?

maciekgrzybek:
Ok I'll close it then

WORMSS:
Can someone re-open this PR? it was classed as a Dupe of 735, but 735 was closed.. so, we need this to re-open so we can fix this permissions problem.

saschanaz:
This is quite outdated, could you open a new issue and point which permission string you need?

PR #802, opened 2019-11-23
Add HTML Media types

saschanaz:


saschanaz:
> Have we done that in the past?

Yes, several MS- and WebKit-specific types have been removed since we started to use Web IDL. Some of them are listed in removedTypes.json but generally are just overridden without manual list.

> Which browsers will be broken? If a lot of people are still targetting those browsers then we should maybe keep the MS-specific stuff around.

AFAICT the removed types here are:

1. Old IE-era APIs that have been removed from Edge
2. Some MS-specific DRM APIs which are kept in Edge <= 18 (Not available on Chromium-based Edge)

I think use of DRM should be rare enough.

inputfiles/idlSources.json
```diff
@@ -276,6 +276,10 @@
         "url": "https://html.spec.whatwg.org/multipage/links.html",
         "title": "HTML - Links"
     },
+    {
+        "url": "https://html.spec.whatwg.org/multipage/media.html",
+        "title": "HTML - Media"
+    },
```
orta: This is the controversial one: It's only supported in Safari and old edge today (without flags ) according to caniuse:
 - https://caniuse.com/#search=video%20track
- https://caniuse.com/#search=audio%20track

inputfiles/idlSources.json
```diff
@@ -358,6 +362,10 @@
         "url": "https://w3c.github.io/mediacapture-main/",
         "title": "Media Capture and Streams"
     },
+    {
+        "url": "https://w3c.github.io/media-playback-quality/",
+        "title": "Media Playback Quality"
+    },
```
orta: This one seems totally fine, it's an editor's draft - yep, but every browser seems to be supporting it

inputfiles/idlSources.json
saschanaz: Ah, didn't know that, nice find! But those types are already in lib.d.ts so removing them might affect existing users. Should I remove them anyway?

orta: We think of it as the status quo is wrong, yep, but adding the new spec would make people expect something exists in the runtime which doesn't (without turning on a flag for 90% of the web). 

If it gets removed from  the PR is it still feasible for the media playback to come in?  Then we keep the dodgy existing behavior until the runtime features available for everyone

saschanaz: The spec itself also covers an essential part of HTML - `HTMLAudio/Video/TrackElement`, so I think the spec should still be added, while we can drop specific types mentioned here.

>If it gets removed from the PR is it still feasible for the media playback to come in?

It shouldn't be affected, as it only adds a function.

orta: Yep, I agree - let's do that 👍

saschanaz:
Had to also remove `SourceBuffer.prototype.textTracks` as it turns out both Firefox and Chrome does not have it. I filed this on [BCD](https://github.com/mdn/browser-compat-data/issues/5781).

inputfiles/addedTypes.json
```diff
@@ -241,6 +249,9 @@
     },
     "interfaces": {
         "interface": {
+            "AudioTrack": {
+                "exposed": ""
```
saschanaz: This is a way to properly exclude certain types from unions.

baselines/dom.generated.d.ts
```diff
@@ -2343,47 +2343,6 @@ declare var AudioScheduledSourceNode: {
...
-};
-
-interface AudioTrackListEventMap {
-    "addtrack": TrackEvent;
-    "change": Event;
-    "removetrack": TrackEvent;
-}
-
-/** Used to represent a list of the audio tracks contained within a given HTML media element, with each track represented by a separate AudioTrack object in the list. */
-interface AudioTrackList extends EventTarget {
```
rbuckton, line 2368: Why were these removed? They're part of the DOM standard: https://html.spec.whatwg.org/multipage/media.html#audiotracklist

saschanaz: True, but unfortunately nobody has a stable implementation per MDN and TSJS-lib-generator generally should only include stable ones.

PR #804, opened 2019-11-24
Add HTML Web workers types

### LINKED ISSUE #807: onrejectionhandled / onunhandledrejection are missing in WorkerGlobalScope
This should work:

```typescript
// In a worker

// Handle promise rejection...
self.onunhandledrejection((event: PromiseRejectionEvent) => {
  throw event.reason;
});

// ...so that let main thread know this promise is rejected.
Promise.new((resolve, reject) => {
  throw new Error('Oops!');
});
```

But I get `Property 'onunhandledrejection' does not exist on type 'WorkerGlobalScope & typeof globalThis'.`.

The `unhandledrejection` event is useful for debugging when you write some code that imports modules asynchronously in a Worker.
I am not sure whether these events / handlers in `WorkerGlobalScope` context are defined by specifications, but at least [MDN says "typically, this is the window, but may also be a `Worker`"](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event) and it works in a Worker.

saschanaz:
Closes #807


sandersn:
Not sure what's going on with the new ExtendableMessageEvent after merging from master. It seems like there are some new diff lines in the last commit that weren't there before.

saschanaz:
>I can't figure out where the ExtendableMessageEvent changes came from

The code currently guesses event interface from other types (`eNameToEType` in emitter.ts), it seems the order somehow changed.

baselines/webworker.generated.d.ts
```diff
@@ -2318,10 +2330,6 @@ declare var NavigationPreloadManager: {
     new(): NavigationPreloadManager;
 };
 
-interface NavigatorBeacon {
-    sendBeacon(url: string, data?: Blob | Int8Array | Int16Array | Int32Array | Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array | DataView | ArrayBuffer | FormData | string | null): boolean;
-}
-
```
orta: We're a tad worried about this getting removed, looks like it's being used still: https://caniuse.com/#search=Beacon

Probably just needs: `https://www.w3.org/TR/beacon/` being added as an IDL source

PR #808, opened 2019-11-28
SubtleCrypto type enhancements

jiripudil:
Hello, this PR fixes some issues I've found with SubtleCrypto types:

- microsoft/TypeScript#23056
- microsoft/TypeScript#35408
- microsoft/TypeScript#35409

This also brings the type definitions closer to the [spec](https://www.w3.org/TR/WebCryptoAPI/#subtlecrypto-interface) which uses `AlgorithmIdentifier` and `KeyUsage` everywhere.


mizvyt:
+1. Just had an issue with `wrapKey` not letting me specify an `iv` for AES-GCM encryption. In the MDN docs, you can find an example that shows an example of the correct usage (https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/wrapKey), but I guess the TypeScript definitions were not updated.

Temporary workaround for my project:
```
interface WrapKeyParams extends Algorithm {
    iv: Uint8Array;
}

<...>

const wrapParams: WrapKeyParams = {name: 'AES-GCM', iv: iv};
const wrapPromise = this.subtleCrypto.wrapKey(
    'jwk',
    keyToWrap,
    key,
    wrapParams
);
```

baselines/dom.generated.d.ts
```diff
@@ -15584,24 +15584,24 @@ declare var StyleSheetList: {
...
+    verify(algorithm: AlgorithmIdentifier | RsaPssParams | EcdsaParams | AesCmacParams, key: CryptoKey, signature: Int8Array | Int16Array | Int32Array | Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array | DataView | ArrayBuffer, data: Int8Array | Int16Array | Int32Array | Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array | DataView | ArrayBuffer): PromiseLike<boolean>;
+    wrapKey(format: string, key: CryptoKey, wrappingKey: CryptoKey, wrapAlgorithm: AlgorithmIdentifier | RsaOaepParams | AesCtrParams | AesCbcParams | AesCmacParams | AesGcmParams | AesCfbParams): PromiseLike<ArrayBuffer>;
```
mizvyt: Wonder if it should also have `format: "raw" | "pkcs8" | "spki" | "jwk"`? (for both wrap and unwrap functions)

jiripudil: Good idea! It makes sense to me to still include `string` though, so that the type definitions work even if another format is added to the spec. Plus all the other methods have a `format: string` overload too. What do you think?

mizvyt: Ah, you're right - I didn't check too closely. Then I don't think there is anything to change, at least to maintain the convention. It seems that the formats are only explicitly specified when there are specific constraints to algorithms that can be used.

jiripudil: I think the specific formats are worth adding, if only for the sake of code completion. I was just saying that `string` should still be there so that it's more future-proof, like this: `format: "raw" | "pkcs8" | "spki" | "jwk" | string`

PR #809, opened 2019-11-29
Add SharedWorker to addedTypes

Toxyxer:
Fixes microsoft/TypeScript#24323


sandersn:
#804 contains this change, and pulls it from [the IDL](https://html.spec.whatwg.org/multipage/workers.html), so I think we’re going to take that one.

PR #810, opened 2019-12-05
Use narrower types for setInterval and setTimeout

XenProject:
Infer specific types for arguments instead of using `any[]`


orta:
Hey @XenProject  - thanks for this PR, we'd like to try keep these dom types as simple as possible unless there is a very large advantage for everyone and I don't think these changes pay for themselves in the extra perf hits.

Anhil-111:
@orta with all due respect, do you not consider making these functions type-safe "a large advantage"? Right now, TS happily allows this code to compile: [playground link](https://www.typescriptlang.org/play/#code/GYVwdgxgLglg9mABMOcAUBDATgcwFyIDOUWMYOAlAWCALYBGAplogN4BQiXiMwiaUAJ4AHRnD7YciAIQBeWYgDkxUuUUU2nbtqgALLHADuiMI2MBRLAaxpFuo4ghwQAGwAmiPTEKJGAN2ZEXQxhUTAAfnUAbi0uAF9YxCxGKBAsJAAWACYYhPZCFIAVGFoxECg0FDgAGkQARgAGJtqMiiigA).

Issue #222, opened 2017-03-28
Emit classes instead of vars


SaschaNaz:
Now that we can extend both class instance type and static type, can we now directly emit classes?

sufianrhazi:
The ability to emit classes would help implement my proposed solution to the WebGL declarations' use of empty interfaces: https://github.com/Microsoft/TypeScript/issues/5855#issuecomment-293117701

SaschaNaz:
I will submit a PR if someone gives Accepting PR tag.

RReverser:
On the other hand, emitting classes makes it much harder to extend built-in types with optional extensions, since class declarations cannot be merged together, but interfaces can.

saschanaz:
> class declarations cannot be merged together, but interfaces can.

This works:

```ts
declare class Merge {
    x: string;
}

interface Merge {
    y: string;
}

new Merge().y;
```

RReverser:
@saschanaz Yeah I'm trying to fix that right now actually, because ran into a problem with polyfilling static methods :) https://github.com/microsoft/TSJS-lib-generator/pull/812

RReverser:
To be clear - I came across this issue because I'd also be happy if there was a way to reduce all that clutter and just use `class`, but polyfilling is quite important for Web APIs, and not being able to do it seems like a dealbreaker.

Hopefully this can be somehow fixed on TypeScript side in the future by allowing merging classes with classes, or adding `static` methods to interfaces or something similar.

saschanaz:
There is a relevant issue on TS: https://github.com/microsoft/TypeScript/issues/2957

RReverser:
Actually... I've just realised that something like this might work for extending statics:

```ts
declare class Merge {
    x: string;
    static X: string;
}

interface Merge {
    y: string;
}

declare namespace Merge {
    const Y: string;
}

new Merge().x;
new Merge().y;
Merge.X;
Merge.Y;
```

So if we implement this, then #812 would be unnecessary.

Either option is fine by me and better than current status quo that doesn't allow any extensions, but this one would probably be even cleaner.

RReverser:
FWIW - just so that our work doesn't conflict - I'm giving this a try.

RReverser:
Update: I have a working implementation of this, but waiting for #813 to be merged first.

ExE-Boss:
[...]

PR #813, opened 2019-12-18
Simplify / fix HTMLCollection overrides

RReverser:
Use an extra optional type param on HTMLCollection for `namedItem` override.

This simplifies / removes quite a lot of emitter code that was necessary to maintain just for this override.

Instead, this PR is using regular backward-compatible type system capabilities.

As an additional benefit, this removes a phantom HTMLCollectionOf class that previously appeared as a valid global, but doesn't really exist in JavaScript global object.


RReverser:
I came across this while implementing #222, but decided it's worthwile to separate out this refactoring into its own PR as it's not strictly related to that issue.

baselines/dom.generated.d.ts
```diff
@@ -6533,36 +6533,27 @@ declare var HTMLCanvasElement: {
...
+    namedItem(name: string): E | N | null;
+    [index: number]: E;
 }
 
 declare var HTMLCollection: {
     prototype: HTMLCollection;
     new(): HTMLCollection;
 };
 
-interface HTMLCollectionOf<T extends Element> extends HTMLCollectionBase {
```
HolgerJeromin: I and probably other people have used it in own code (API which should usable with `getElementsByTagNameNS` output for example.
So this interface should probably survive with a `@deprecated` tag.

saschanaz: The API name also is consistent with `NodeListOf`.

RReverser: Added a deprecated `HTMLCollectionOf` alias for backwards compatibility. @saschanaz should I maybe do the same for `NodeListOf` then for consistency?

RReverser: Ping @saschanaz.

saschanaz: I don't have a strong opinion here, maybe @sandersn ?

saschanaz:
Revisiting this after 2 years... 😅

Now that I have a permission to merge, I find this beneficial with no regression. Could you try rebasing? I can cherry-pick this PR if you are too busy. Thank you!

RReverser:
Lol it's been a while, yeah 😅 If you could rebase on your side, that would be great, yeah.

Bashamega:
Can I create a new PR for this @saschanaz

saschanaz:
Please go ahead.

PR #815, opened 2020-01-06
customElement.define's 2nd parameter is a constructor

sandersn:
Previously it was a call signature, but this is not correct. I'm not sure if `HTML - Custom Elements.widl` is wrong or whether our interpretation of it is wrong. A call signature does not match actual usage or the semantics as implemented by Chrome, however.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -19255,7 +19255,7 @@ interface BlobCallback {
 }
 
 interface CustomElementConstructor {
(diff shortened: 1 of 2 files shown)


sandersn:
@saschanaz I'm going to merge this since I'm pretty sure it's right based on [usage from MDN's example repo](https://github.com/mdn/web-components-examples/search?q=customElements.define&unscoped_q=customElements.define), but if you have comments, please do chime in.

PR #818, opened 2020-01-22
Add HTMLElement#attachInternals()

naokikimura:
Fixes https://github.com/microsoft/TypeScript/issues/33218

https://github.com/microsoft/TypeScript/pull/36347 was rejected, so I made this PR.

Rather, please tell me why you removed it.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1840,6 +1840,19 @@ interface VRStageParameters {
     sizeY?: number;
 }
 
+interface ValidityStateFlags {
+    badInput?: boolean;
+    customError?: boolean;
+    patternMismatch?: boolean;
+    rangeOverflow?: boolean;
+    rangeUnderflow?: boolean;
+    stepMismatch?: boolean;
+    tooLong?: boolean;
+    tooShort?: boolean;
+    typeMismatch?: boolean;
+    valueMissing?: boolean;
+}
+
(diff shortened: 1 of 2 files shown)


naokikimura:
@sandersn I would appreciate it if you could review this

saschanaz:
>Rather, please tell me why you removed it.

Those types have only one implementation which is from Chromium, that's why it's removed from here.

naokikimura:
@saschanaz 

Thank you for commenting.

It is already implemented in Firefox Nightly. https://hg.mozilla.org/mozilla-central/rev/e6b60876e668

Does the implementation of that method need to be completed in stable version?

saschanaz:
>Does the implementation of that method need to be completed in stable version?

Yes, as Nightly can always drop it without shipping it to stable.

adriengibrat:
What about moving definition in a separated file so it's up to the user to use it or not (like some ES features) ?

saschanaz:
Currently Definitely Typed is recommended for that purpose.

orta:
I tested support for Firefox and Safari via this glitch app and they're still not available un-flagged: https://web.dev/more-capable-form-controls/#a-working-example

We're still happy to wait until that changes, otherwise as @saschanaz says DT is your best option to get it shipped today

saschanaz:
#1141 and #1149 partially restored ElementInternals types thanks to the new Firefox implementation. Others will be restored when further implementation happens.

@github-actions close

PR #819, opened 2020-01-27
fix(dom): narrow ownerDocument in specific Node interfaces

eps1lon:
> The ownerDocument attribute’s getter must return null, if this is a document, and this’s node document otherwise.

-- https://dom.spec.whatwg.org/#dom-node-ownerdocument

Closes https://github.com/microsoft/TypeScript/issues/36262


eps1lon:
@sandersn Done.

PR #820, opened 2020-01-27
fix(dom): intersect defaultView with globalThis

eps1lon:
Follow-up to #715

With #819 this change allows 
```ts
declare const element: Element;
const isInput = element instanceof element.ownerDocument.defaultView!.HTMLInputElement;
```

Still requires non-null assertion which is fine. Above code will never be sound since defaultView might not be available. This is, at least for my purpose, alright.
Currently I simply have to `@ts-ignore` `element.ownerDocument.defaultView.HTMLInputElement`.


HolgerJeromin:
This change is not unproblematic when used in iframes:
https://github.com/microsoft/TypeScript/issues/41684#issuecomment-733927253

PR #822, opened 2020-02-14
Fix RTCStatsType values

ydz-one:
Change RTCStatsType values to match MDN docs

Fixes https://github.com/microsoft/TypeScript/issues/36650


HolgerJeromin:
https://github.com/microsoft/TSJS-lib-generator/blob/master/inputfiles/README.md says you should not edit this json...

PR #823, opened 2020-02-15
Update HeadersInit type

ArnaudBarre:
Some headers can be conditional like Authorization in my case. Adding the undefined possibility is correct (You can try `await fetch("https://github.com", { headers: { "Authorization": undefined } })` in your console) and simplify the syntax:

```ts
fetch("https://github.com", {
  headers: {
    Authorization: isAuthEndpoint ? `Bearer ${token}` : undefined,
    "Content-Type": "application/json",
  },
});
```

instead of: 

```ts
fetch("https://github.com", {
  headers: {
    ...(isAuthEndpoint ? { Authorization: `Bearer ${token}` } : undefined),
    "Content-Type": "application/json",
  },
});
```


ArnaudBarre:
Found that it actually sends the string undefined. Was just by chance it was working for me

Issue #824, opened 2020-02-19
Missing CSS overscroll-behavior types


MartinBlackburn:
The types for the CSS overscroll-behavior property are missing:
https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior

saschanaz:
@github-actions close

PR #825, opened 2020-02-19
Added CSS overscroll-behavior

MartinBlackburn:
Adding the CSS overscoll-behavior property. This should fix issue #824

### DIFF
--- a/inputfiles/idlSources.json
+++ b/inputfiles/idlSources.json
@@ -85,6 +85,10 @@
         "title": "CSS Overflow"
     },
     {
+        "url": "https://www.w3.org/TR/css-overscroll-1/",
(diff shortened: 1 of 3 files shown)


inputfiles/idlSources.json
```diff
@@ -84,6 +84,10 @@
         "url": "https://www.w3.org/TR/css-overflow-3/",
         "title": "CSS Overflow"
     },
+    {
+        "url": "https://www.w3.org/TR/css-overscroll-1/",
+        "title": "CSS Overscroll"
```
saschanaz: Maybe CSS Overscroll Behavior?

MartinBlackburn: done :-)

PR #827, opened 2020-02-23
Add WritableStream.close()

MattiasBuelens:
The streams standard [has added](https://github.com/whatwg/streams/commit/547428be7c067156a15a8762ff9933193befc753) a new shorthand method to close a writable stream: [WritableStream.close()](https://streams.spec.whatwg.org/#ws-close). It's [already implemented in Chrome 81](https://chromestatus.com/feature/5440098147500032).

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -18827,6 +18827,7 @@ declare var Worklet: {
 interface WritableStream<W = any> {
     readonly locked: boolean;
     abort(reason?: any): Promise<void>;
+    close(): Promise<void>;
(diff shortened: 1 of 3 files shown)


saschanaz:
`WritableStream` lost an implementation so Chromium has the only implementation now, which makes it inappropriate to be added to lib.d.ts....

MattiasBuelens:
Oh right, EdgeHTML is no more, and Gecko and WebKit only support `ReadableStream`... 😞

I'll close this for now. If more browsers add support for the other stream APIs in the future, I can always open a new PR, no big deal. 🙂

PR #828, opened 2020-02-25
Add potential hueristics for the README

orta:
Notes -> Prose from our pairing session

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ To give you a sense of whether we will accept changes, you can use these heurist
 - Does the PR show examples of the changes being used in spec examples or reputable websites like MDN?
 - Did this change come from an IDL update?
 - Does the change appear to be high-impact on a well-used API?
-- Would the changes introduce a lot of breaking changes to existing code in DT?
+- Would the changes introduce a lot of breaking changes to existing code? For example the large corpus of typed code in DefinitelyTyped.
 
 #### Additions
 
--- a/README.md
+++ b/README.md
@@ -51,17 +51,44 @@ The common steps to send a pull request are:
 3. Update the files in the `baselines` folder using the newly generated files
    under `generated` folder (`npm run baseline-accept`).
 
-### When should a DOM API be included here?
+### What are the TypeScript team's heuristics for PRs to the DOM APIs
 
-A feature is considered highly experimental if a DOM API lacks multiple implementations or a formal specification from W3C or WHATWG. Such a feature belongs on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#readme).
-When it gets multiple implementations and a proper specification, it can be added here and removed from DefinitelyTyped.
+Changes to this repo can have pretty drastic ecosystem effects, because these types are included by default in TypeScript. 
+Due to this, we tend to be quite conservative with our approach to introducing changes.
+To give you a sense of whether we will accept changes, you can use these heuristics to know up-front if we'll be open to merging. 
 
+#### Fixes
+
+> For example, changing a type on a field, or nullability references
+
+- Does the PR show examples of the changes being used in spec examples or reputable websites like MDN?
+- Did this change come from an IDL update?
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,18 +51,45 @@ The common steps to send a pull request are:
...
+
+- Does the PR show examples of the changes being used in spec examples or reputable websites like MDN?
+- Did this change come from an IDL update?
+- Does the change appear to be high-impact on a well-used API?
+
+#### Additions
+
+> For example, adding a new spec or subsection via a new or updated IDL file
+
+- Does the new objects or fields show up in [mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)? If not, it's likely too soon
```
sandersn:
```suggestion
- Does the new objects or fields show up in [mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)? If not, it's likely too soon.
```

README.md
```diff
@@ -51,18 +51,45 @@ The common steps to send a pull request are:
...
+- What stage of the [W3C process](https://en.wikipedia.org/wiki/World_Wide_Web_Consortium#Specification_maturation) is the proposal for these changes: We aim for Proposed recommendation, but can accept Candidate recommendation for stable looking proposals.
```
sandersn: I still don't understand if W3C proposals are intended to be mutually exclusive with whatwg additions. If so, we should say something like: "Changes that don't have a W3C proposal should be on whatwg instead."

README.md
```diff
@@ -51,18 +51,45 @@ The common steps to send a pull request are:
...
+- Are the additions available in at least two of Firefox, Safari and Chromium?
+- Could any types added at the global scope have naming conflicts?
+- Are the features going to be used by a lot of people?
+
+#### Removals
+
+> For example, removing a browser-specific section of code
+
+- Do the remove objects or fields show up in [mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)? If so, are they marked as deprecated?
```
sandersn:
```suggestion
- Do the removed objects or fields show up in [mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)? If so, are they marked as deprecated?
```

README.md
saschanaz: Waiting for PR and CR is way too slow, that's why WHATWG is established and maintaining their own versions of specs. I think the browser availability is enough as that means there is good consensus.

README.md
```diff
@@ -51,18 +51,49 @@ The common steps to send a pull request are:
...
+
+#### Additions
+
+> For example, adding a new spec or subsection via a new or updated IDL file
+
+- Does the new objects or fields show up in [mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)? If not, it's likely too soon.
+- Is the IDL source from WHATWG? 
+    - Are the additions available in at least two of Firefox, Safari and Chromium?
+- Is the IDL source from W3C? 
+    - What stage of the [W3C process](https://en.wikipedia.org/wiki/World_Wide_Web_Consortium#Specification_maturation) is the proposal for these changes: We aim for Proposed recommendation, but can accept Candidate recommendation for stable looking proposals.
```
saschanaz: What would be reasons to take W3C specs in different way? W3C also will try not to introduce a breaking change if multiple implementations are already shipped and being used, as doing so will break the web. I know they have their approval steps, but I think that's more about algorithmic details rather than the API form.

orta: I think of this as we're seeing this a little bit more mechanically (because we're not involved in the process) and trying to find heuristics to understand how to react directly from a PR

The main reason they were differentiated was because they came with differing level of ways to verify "done"-ness. A W3C spec can says "this is ready, and won't change" vs a WHATWG which is always living and we have to validate

saschanaz: Makes sense. I guess I'm okay with this as I think [you will allow some exceptions](https://github.com/microsoft/TSJS-lib-generator/pull/802#discussion_r387813428) 👍

orta:
Thanks @saschanaz

PR #829, opened 2020-02-26
Testing CI via github actions

sandersn:
I'm almost certainly doing this the wrong way because I remember it being much easier the first time I added it to a repo.


sandersn:
Never mind, I just realised that none of the tests that are easy to run will show diffs -- just the name of the failing case. We need the extra work that typescript-bot does to create PRs that show the diffs.

I'm going to back out the github actions config and consult with @weswigham again to figure out how to make this happen.

PR #830, opened 2020-03-02
fix(NamedNodeMap): Index signature accepts `string`

ExE-Boss:
See the [**DOM Specification**](https://dom.spec.whatwg.org/#interface-namednodemap) for details.


saschanaz:
The failure explains why it isn't there 😓

PR #834, opened 2020-03-16
Remove Gecko specific Navigator members

### LINKED ISSUE #748: Navigator.oscpu should not be present (or allow undefined)
Navigator.oscpu is a gecko-specific extension as called out in the corresponding spec documents. It should either be removed, or its type should reflect that it will often be undefined in practice.

/cc @chradek

saschanaz:
Fixes #748 

As described in https://html.spec.whatwg.org/multipage/system-state.html#NavigatorID-partial, this is only available by default in Firefox.


orta:
Cool - thanks @saschanaz

PR #835, opened 2020-03-16
Add CSS Cascading and Inheritance

hikarinotomadoi:
Adding the CSS Cascading and Inheritance property.
Fixes https://github.com/microsoft/TypeScript/issues/37215


orta:
Agree, this looks like it's not changing anytime soon. Let's get it in.

PR #836, opened 2020-03-17
Fix duplicates after merge in addedTypes.json

IllusionMH:
Noticed that there are problems with in `addedTypes.json` (thanks VS Code!) which were related to duplicated entries in `interfaces` > `interface`.

I've removed blocks with `cloneNode` and re-run patched helper script to bring them back to existing nodes.

Updated baselines should now contain proper overrides and elements names that were skipped because of duplicated keys later in JSON.

Related to #811

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -416,6 +416,16 @@
                             "type": "MediaEncryptedEvent"
                         }
                     ]
+                },
+                "methods": {
+                    "method": {
+                        "cloneNode": {
+                            "name": "cloneNode",
+                            "override-signatures": [
+                                "cloneNode(deep?: boolean): HTMLMediaElement"
+                            ]
+                        }
+                    }
                 }
             },
             "CSSStyleDeclaration": {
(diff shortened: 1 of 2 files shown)


inputfiles/addedTypes.json
```diff
@@ -3700,19 +3685,6 @@
                     }
                 }
             },
-            "HTMLVideoElement": {
```
sandersn: I don't see an existing entry where this HTMLVideoElement method is added. Was that an incorrect artifact of the merge too?

IllusionMH: I was confused at first, but looks like you've added it by hand after problematic merge.
https://github.com/microsoft/TSJS-lib-generator/pull/811#discussion_r393833305

IllusionMH: Here it is in file https://github.com/microsoft/TSJS-lib-generator/blob/af44ea1019180f60aced843d65a1fd048adb5735/inputfiles/addedTypes.json#L2265

IllusionMH:
Thanks for the review!

I think that at this point all `cloneNode` types are in place, however it's hard to tell if anything from master was affected by merge :(

PR #837, opened 2020-03-20
Add Web Share types

saschanaz:
Fixes https://github.com/microsoft/TypeScript/issues/18642.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1697,6 +1697,12 @@ interface ShadowRootInit {
     mode: ShadowRootMode;
 }
 
+interface ShareData {
+    text?: string;
+    title?: string;
+    url?: string;
+}
+
 interface SpeechSynthesisErrorEventInit extends SpeechSynthesisEventInit {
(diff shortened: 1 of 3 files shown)


orta:
Thanks @saschanaz

Artur-:
The suggestion in the linked issue suggests adding
```
interface Navigator {
  share?: NavigatorShare;
}
```

however this PR adds it as
```
interface Navigator {
  share: NavigatorShare;
}
```

As `share` is not available in most browsers, you want to check for the existence of it before using it but
```
if (navigator.share) {
```

now results in an error
```
This condition will always return true since the function is always defined. Did you mean to call it instead? ts(2774)
```

samtgarson:
@orta This comment 👆 seems correct, is there a reason `share` is always defined here?

orta:
That is how the w3c spec defines it- you're welcome to send a PR making it optional 👍🏻

ferm10n:
any reason why `canShare` wasn't added as well?

saschanaz:
It's added later by #1125.

PR #839, opened 2020-03-20
Update DOMParser type

saschanaz:
The type is moved to HTML.


baselines/dom.generated.d.ts
```diff
@@ -3922,7 +3922,16 @@ declare var DOMMatrixReadOnly: {
 
 /** Provides the ability to parse XML or HTML source code from a string into a DOM Document. */
 interface DOMParser {
-    parseFromString(str: string, type: SupportedType): Document;
+    /**
+     * Parses string using either the HTML or XML parser, according to type, and returns the resulting Document. type can be "text/html" (which will invoke the HTML parser), or any of "text/xml", "application/xml", "application/xhtml+xml", or "image/svg+xml" (which will invoke the XML parser).
+     * 
+     * For the XML parser, if string can be parsed, then the returned Document will contain elements describing the resulting error.
```
HolgerJeromin, line 3928: IMO we need a negation here!
https://github.com/whatwg/html/pull/5540

PR #842, opened 2020-03-25
Remove clone node overrides

sandersn:
They make `@types/tablesorter`, a jquery dependent, invariant in a complex way that I have yet to figure out. I'm reverting this change for 3.9 until I
have time to see how likely other projects are to encounter the problem.

@orta @DanielRosenwasser


david-fong:
Sorry if spam. Just leaving a more explicit paper trail for anybody trying to track down the history: This change reverts #811.

Issue #47, opened 2016-01-13
Affecting the ES6 library


rapropos:
I am attempting to update the WebCrypto interface definitions, where functions return Promises. If I declare these return values as Promises, tsc will fail to build. If I change this to PromiseLike, it builds fine, but if I am targeting es6 (or using the es6 library via other means), it would be nice to have the return values be actual Promises. Is there a way to specify separate signatures for es5 and es6 libraries?

mhegazy:
We should generate a new file dom.es6.d.ts, that include the definitions that contain es6-specific definitions. and we can wire it in the TS build in https://github.com/Microsoft/TypeScript/blob/master/Jakefile.js#L172

rapropos:
If it's in a separate file, won't there be conflicting definitions in the eventual lib.es6.d.ts output? In this case, the method signature changes depending on the language level.

mhegazy:
if the difference is only in the return types, then we need to to have two separate files that include all definitions. then the build can pick one or the other based on the language version.

rapropos:
Perhaps I'm not properly explaining what I'm worried about. As I understand things, when I add to `overridingTypes.json`, these changes are reflected both in `lib.d.ts` and `lib.es6.d.ts` via `dom.generated.d.ts`. Let's say we have:

```
    {
         "kind": "method",
         "interface": "Foo",
         "name": "bar",
         "signatures": ["bar(param:string): PromiseLike<string>"]
     }
```

So if there is a new file, say `overridingTypesEs6.json`, and I put the following version in it:

```
    {
         "kind": "method",
         "interface": "Foo",
         "name": "bar",
         "signatures": ["bar(param:string): Promise<string>"]
     }
```

...and that gets put into a `dom.es6.d.ts` file, won't both versions of the declaration try to end up in `lib.es6.d.ts`? The first version via the existing mechanism, and the second via the proposed new route?

Perhaps there could be an `es6-signatures` key that would provide method signatures to be picked up in lieu of the ones specified in `signatures`?

mhegazy:
what i meant is to generate two files here, `dom.generated.d.ts` and `dom.es6.generated.d.ts`. the second can be just an augmentation to the first, or it can be a replica of it with only different return types. then in the main ts repo, we would wire the build to pick one or the other based on which lib it is building.

SaschaNaz:
Is this fixed by https://github.com/Microsoft/TypeScript/pull/14053?

[...]

PR #843, opened 2020-03-25
Change SubtleCrypto method return types to Promise

taymonbeal:
Previously, these methods returned `PromiseLike`, even though the spec has always required them to return real `Promise`s and browsers have always done so. According to #60, this was done in order to work around #47; now that #47 has been fixed, this workaround is obsolete.

This generally should not be a breaking change since it is making the `SubtleCrypto` type contract stricter and that API is generally implemented by the platform, not by users.


sandersn:
Thanks for following up on this ancient issue @taymonbeal.

PR #844, opened 2020-03-26
Refine webworker interface clients get signature

Rustin-Liu:
fixes https://github.com/microsoft/TypeScript/issues/36705

it should be https://w3c.github.io/ServiceWorker/#dom-clients-get
`  get(id: string): Promise<Client | undefined>; `


Rustin-Liu:
@sandersn  Could you please take a look?

PR #845, opened 2020-03-31
Support web API VisualViewport

Qiyu8:
Fixes [#37585](https://github.com/microsoft/TypeScript/issues/37585)
All tests passed.
The Visual Viewport API is designed to provide an explicit mechanism for developers to query and potentially modify the properties of the visual viewport. It also introduces events that allow the page to listen for changes in the visual viewport, allowing UX that explicitly wants to react to these changes to do so. For example, the page could keep a small text-formatting bar above the OSK.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -16146,6 +16146,32 @@ declare var VideoPlaybackQuality: {
     new(): VideoPlaybackQuality;
 };
 
+interface VisualViewportEventMap {
+    "resize": UIEvent;
+    "scroll": Event;
+}
+
+interface VisualViewport extends EventTarget {
+    readonly height: number;
+    readonly offsetLeft: number;
+    readonly offsetTop: number;
+    onresize: ((this: VisualViewport, ev: UIEvent) => any) | null;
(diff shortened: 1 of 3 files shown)


orta:
We're a little conflicted on this:

### For

 It's in safari and chrome unprefixed, and available via a flag in Firefox

### Against

[It's a community group draft](https://wicg.github.io/visual-viewport/#the-visualviewport-interface), with no goals to go on the standards track

> It is not a W3C Standard nor is it on the W3C Standards Track.

and [MDN pushes](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport) it as being experimental pretty hard

--- 

I think we're gonna be blocked on this until one of the following happens:

 - firefox allows this in prod
 - it gets past draft stage

I'd recommend adding it to definitely typed if it's a blocker

saschanaz:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1551302 this is enabled by default in Firefox Preview on Android.

orta:
OK, thanks @saschanaz - I re-read the spec and if it's in Firefox for android, I think this is considered mainstream enough to be added. It feels like something that is much more useful on mobile browsers vs desktop, so having pretty much all of them doing it in prod works for me.

PR #846, opened 2020-04-01
Fix brokenLinks of speech API

Qiyu8:
The original https://w3c.github.io/speech-api/ has been replaced with new one.


orta:
👍 thanks!

PR #850, opened 2020-04-13
WebSpeech's SpeechRecognition onerror incorrectly reports ErrorEvent as event

shareefalis:
See: https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionErrorEvent

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -15049,7 +15049,7 @@ interface SpeechRecognitionEventMap {
     "audioend": Event;
     "audiostart": Event;
     "end": Event;
-    "error": Event;
+    "error": ErrorEvent;
     "nomatch": SpeechRecognitionEvent;
     "result": SpeechRecognitionEvent;
     "soundend": Event;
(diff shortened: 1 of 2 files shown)


shareefalis:
For a more robust implementation. `SpeechRecognitionErrorEvent` type has to be implemented 

Reference: https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionErrorEvent

saschanaz:
`ErrorEvent` has nothing to do with `SpeechRecognitionErrorEvent`. This is done because no browser implements the interface.

sandersn:
It's OK, I guess, to use a different subtype that's "close enough". I'm not a huge fan, but I appreciate that default is easy even if it's not correct.

In the event that `SpeechRecognitionErrorEvent` is implemented, the PR that switches `ErrorEvent` to `SpeechRecognitionErrorEvent` will need to find and refer to this PR to explain why it's the right change.

saschanaz:
@sandersn Despite the name, `SpeechRecognitionErrorEvent` has no `filename`, `lineno`, `colno` etc. and thus can confuse users more.

sandersn:
@orta and I had a hard time deciding, but it seems like the choice is basically between a type with too many properties, `ErrorEvent` and a type with too few, `Event`. `ErrorEvent` is actually an unrelated subtype of `Event`, but that's mainly confusing to maintainers.

I went with the too-large type because it at least provides `error` and `message` even if `filename/lineno/colno` are also mistakenly present. If you feel strongly about this I will revert it.

saschanaz:
That rationale sounds reasonable. Since the names of the extra properties strongly suggests they are irrelevant here, I don't oppose here. Thanks for the details behind the decision 👍

PR #851, opened 2020-04-14
fix(dom): DOMRECTList index can return undefined

keithamus:
See https://drafts.fxtf.org/geometry-1/#domrectlist

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3992,7 +3992,7 @@ declare var SVGRect: typeof DOMRect;
 interface DOMRectList {
     readonly length: number;
     item(index: number): DOMRect | null;
-    [index: number]: DOMRect;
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
Thanks for helping.
This is a generated file, so you have to adjust the source.

keithamus:
The source that's here? https://github.com/microsoft/TSJS-lib-generator/blob/15a4678e0ef6de308e79451503e444e9949ee849/inputfiles/idl/Geometry%20Interfaces.widl#L79-L82 or something else?

orta:
Hrm, interesting - it's reasonable that the WIDL you posted should generate an `undefined` union.

It's quite possible that the WIDL->DTS does not add the `| undefined` here for the sake of convenience (like we don't force it on array index accessors )  -- it might need an explicit override in the big JSON file

PR #858, opened 2020-05-21
feat: Use `declare class` where possible

ExE-Boss:
This changes emit for interfaces:
```webidl
interface Foo : Bar {
	constructor();
};
Foo includes Baz;
```

From:
```ts
declare interface Foo extends Bar, Baz {}
declare var Foo: {
	readonly prototype: Foo;
	new(): Foo;
}
```

To:
```ts
declare interface Foo extends Baz {};
declare class Foo extends Bar {
	constructor();
]
```

Which has the added benefit of supporting declaration merging for static methods and properties.

---

It also prevents static properties from breaking `// $ExpectType` on constructors, since it just becomes `// $ExpecType typeof Foo`.

---

This shouldn’t be a breaking change as existing consumers will keep working just fine, this just opens up new ways to extends these definitions.

### DIFF
--- a/src/emitter.ts
+++ b/src/emitter.ts
@@ -120,6 +120,16 @@ function isEventHandler(p: Browser.Property) {
     return typeof p["event-handler"] === "string";
 }
 
+const newToConstructorRegExp = /^new(?:<.+?>)?(\(.+?\))(?:: ?[^)]+)?$/u;
+function convertNewToConstructor(signature: string) {
+    const result = newToConstructorRegExp.exec(signature);
+    if (result) {
+        return `constructor${result[1]}`;
+    } else {
+        return signature;
+    }
+}
+
 export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor) {
     // Global print target
     const printer = createTextWriter("\n");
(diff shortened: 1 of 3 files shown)


src/emitter.ts
```diff
@@ -120,6 +120,16 @@ function isEventHandler(p: Browser.Property) {
     return typeof p["event-handler"] === "string";
 }
 
+const newToConstructorRegExp = /^new(?:<.+?>)?(\(.+?\))(?:: ?[^)]+)?$/u;
+function convertNewToConstructor(signature: string) {
```
saschanaz, line 124: This seems only for manual JSON data, in that case I think those should be fixed instead.

sandersn:
This is a possibly gigantic breaking change. I created a PR on Typescript, microsoft/Typescript#39044, so we can run DT and other tests on it.

sandersn:
We discussed this in the design meeting and decided not to take this change.

- The DOM types are not really classes.
- Static-side merging still only works with namespaces, which is not discoverable (and fewer people over time know how to use namespaces at all).
- So far, for cases where we want to enable static-side merging, we manually introduce an interface for the static side. We would take a PR that generates these interfaces for all class-like types.

saschanaz:
Added a comment: https://github.com/microsoft/TypeScript/issues/39054#issuecomment-644299696

PR #860, opened 2020-06-05
Make the `data` property of MessageEvent generic

cawa-93:
Fix https://github.com/microsoft/TypeScript/issues/19370


CKGrafico:
We have the message "MessageEvent" is not generic when
```
window.addEventListener('message', (event: MessageEvent<CustomDataType>) => {})
```

sapter:
I'm also still getting an error Type `'MessageEvent' is not generic`. @CKGrafico have you been able to resolve this or figure out why we're still getting an error? I'm on TS v4.2.2

CKGrafico:
@sapter nop sorry, there is the worst workaround in the world :'(

```
 window.addEventListener('message', (event: MessageEvent) => {
      const data: PanelMessage = event.data;
```

bb-in-hoodie:
There is no error underline in VSCode, but it keeps yield an error message in terminal when the code gets built.
`Type 'MessageEvent' is not generic.`
Is there no fix yet?

saschanaz:
> There is no error underline in VSCode, but it keeps yield an error message in terminal when the code gets built.
> `Type 'MessageEvent' is not generic.`
> Is there no fix yet?

I think your local TypeScript version is low in that case.

bb-in-hoodie:
> I think your local TypeScript version is low in that case.

since which TS version does support this typing?
I'm on v3.9.6 for now.

210618-04-message_event_TS_error

cawa-93:
> since which TS version does support this typing?

@bb-in-hoodie >= v4

bb-in-hoodie:
Thank you 👍

PR #865, opened 2020-06-05
General IDL updates

saschanaz:
General updates that introduce no TS type change.


inputfiles/idl/HTML - Form elements.widl
```diff
@@ -76,9 +76,10 @@ interface HTMLOptGroupElement : HTMLElement {
 };
 
 [Exposed=Window,
- HTMLConstructor,
- NamedConstructor=Option(optional DOMString text = "", optional DOMString value, optional boolean defaultSelected = false, optional boolean selected = false)]
+ LegacyFactoryFunction=Option(optional DOMString text = "", optional DOMString value, optional boolean defaultSelected = false, optional boolean selected = false)]
```
sandersn, line 79: this change resulted in deleting the `Option` constructor. I think it should remain; do we need to update our code to understand `LegacyFactoryFunction` instead (or in addition)?

baselines/dom.generated.d.ts
```diff
@@ -19408,9 +19408,6 @@ declare var Audio: {
 declare var Image: {
     new(width?: number, height?: number): HTMLImageElement;
 };
-declare var Option: {
-    new(text?: string, value?: string, defaultSelected?: boolean, selected?: boolean): HTMLOptionElement;
-};
```
orta, line 19413: This one feels a bit off to me, this still works in a browser:

```js
new Option()
```

orta: Perhaps bad WIDL on the specs?

inputfiles/idl/HTML - Form elements.widl
saschanaz: Oops, yes exactly. I'll revert this change for now and write a new PR for this.

PR #872, opened 2020-06-13
Replace Promises with union types in callbacks

saschanaz:
A cleanup before updating Streams types.


sandersn (changes requested):
I'm worried about this breaking callbacks that only return PromiseLike instead of Promise. Can you explain why this was PromiseLike originally? Was that incorrect? Will these be good breaks?

saschanaz:
No, it turns out that PromiseLike must be used as general thenables should also work.

saschanaz:
Some problems, will reopen soon

saschanaz:
Never mind, my confusion 👀

sandersn:
The diff looks OK now, just need to get the build passing

saschanaz:
@sandersn That's #903, it needs some love.

saschanaz:
Replaced by #1034

PR #873, opened 2020-06-13
Support [LegacyNoInterfaceObject]

saschanaz:
Geolocation changed their interface names to exposed them without conflicts, and Web Authentication introduced some changes including removing a reference to a Geolocation type.


orta (approved):
Looks good to us 👍🏻

PR #877, opened 2020-06-28
Generate Worker iterable types

### LINKED ISSUE #729: Webworker Headers missing entries()
Worker Header interface is missing the `entries()` method:
https://developer.mozilla.org/en-US/docs/Web/API/Headers/entries

> The Headers.entries() method returns an iterator allowing to go through all key/value pairs contained in this object. The both the key and value of each pairs are ByteString objects.

Problem is probably due to it being worker only.

> Note: This method is available in Web Workers.

It's not listed in the `fetch.widl` https://github.com/microsoft/TSJS-lib-generator/blob/a20c41802d11d47dbe27b09f0ad2c9f47a3db421/inputfiles/idl/Fetch.widl#L3-L12

Requires changes to `inputfiles/overridingTypes.json` to implement.

saschanaz:
Fixes #729

### DIFF
--- a/baselines/webworker.iterable.generated.d.ts
+++ b/baselines/webworker.iterable.generated.d.ts
@@ -1,0 +1,139 @@
+/////////////////////////////
+/// Worker Iterable APIs
+/////////////////////////////
+
+interface Cache {
+    addAll(requests: Iterable<RequestInfo>): Promise<void>;
+}
+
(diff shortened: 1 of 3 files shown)


sandersn (approved):

sandersn:
Note that this will require some work on the compiler to make this accessible. @orta is tracking that work on a work item he created on Typescript.

PR #878, opened 2020-06-30
Update overridingTypes.json

lifeiscontent:
Fix setSelectionRange types as `null` is a valid argument and comes from selectionStart and selectionEnd types as well

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -7350,7 +7350,7 @@ interface HTMLInputElement extends HTMLElement {
      * @param end The offset into the text field for the end of the selection.
      * @param direction The direction in which the selection is performed.
      */
-    setSelectionRange(start: number, end: number, direction?: "forward" | "backward" | "none"): void;
+    setSelectionRange(start: number | null, end: number | null, direction?: "forward" | "backward" | "none"): void;
     /**
      * Decrements a range input control's value by the value given by the Step attribute. If the optional parameter is used, it will decrement the input control's step value multiplied by the parameter's value.
      * @param n Value to decrement the value by.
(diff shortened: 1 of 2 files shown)


sandersn (changes requested):
Passing `null` doesn't crash, but [MDN doesn't document null in the type](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange), and `selectionStart/selectionEnd` don't appear to be null on a new textarea. I think it's better to keep the stricter type we have now.

HolgerJeromin:
The spec explicit states `null` is a valid input. But I am unsure if this would be an improvement.

> To set the selection range with an integer or null start, an integer or null or the special value infinity end, and optionally a string direction, run the following steps:

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#set-the-selection-range

lifeiscontent:
@sandersn i can put a test case together if you want me to show you how this change is beneficial. Without this change I need to do a useless null check and pass 0 manually instead just to make typescript happy.

sandersn:
Thanks! An example would definitely help me understand the change.

lifeiscontent:
@sandersn I've left comments in the example to outline the expectations of the demo, as well as the issues related to type signatures.


https://codesandbox.io/s/pedantic-pond-15ec4

lifeiscontent:
@sandersn any thoughts?

lifeiscontent:
@sandersn ping

sandersn:
Thanks for the example. The easiest argument for me to understand is that in HTMLInputElement, `selectionStart: number | null` but `setSelectionRange` has `start: number`. So you can't pass selectionStart to setSelectionRange.

sandersn:
If you can update this with the baseline changes, I think this will be ready to go.

lifeiscontent:
> If you can update this with the baseline changes, I think this will be ready to go.

Sure, I’m not familiar with this project. Do I just need to rebase? Or can you give me a outline of what I need to do to make that happen?

lifeiscontent:
@sandersn figured it out, good to go! :)

sandersn (approved):

saschanaz:
It's not that `null` is explicitly supported here, it's just that `null` is converted to `0` via `Number(null)` in IDL layer. Doing this makes an inconsistency: do we want to allow `null` for all number parameters?

lifeiscontent:
@saschanaz do you have a suggestion on how I could use `0` in my example code above and get a similar level of code ergonomics?

saschanaz:
I would do this:

```tsx
selectionRef.current = [
  inputRef.current.selectionStart!,
  inputRef.current.selectionEnd!,
  inputRef.current.selectionDirection!
];
```

because in this case it's guaranteed that the values are not null. (`null` only occurs when it's not a text input.)

But it's a valid argument that passing `null` value is also actually okay. I wonder we want to generalize this or not.

lifeiscontent:
@saschanaz I see, I would be in favor of this as it allows better ergonomics related to things like this.

PR #879, opened 2020-06-30
Fix HkdfParams use and remove HkdfCtrParams

saghul:
The latter was removed in 2016: https://github.com/w3c/webcrypto/pull/16 with the former being in the current spec: https://w3c.github.io/webcrypto/#hkdf-params

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -9003,12 +9003,6 @@ declare var History: {
     new(): History;
 };
 
-interface HkdfCtrParams extends Algorithm {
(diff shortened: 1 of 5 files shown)


uhoreg:
Coincidentally, I ran into this issue earlier today and came up with essentially the same fix before noticing this PR.

`HkdfCtrParams` was replaced with `HkdfParams` in the [December 2016 WebCrypto Proposed Recommendation](https://www.w3.org/TR/2016/PR-WebCryptoAPI-20161215/), so `HkdfCtrParams` is obsolete by at least 3.5 years.

saghul:
Gentle ping, @sandersn

orta:
Looks good, thanks!

PR #883, opened 2020-07-15
Switch 'name' to 'void', mark as deprecated.

DanielRosenwasser:
Kinda fixes https://github.com/microsoft/TypeScript/issues/18433


HolgerJeromin:
What about "the others"?
The linked issue mentioned this list (I removed `error` from the list):
`["closed", "event", "fdescribe", "name", "length", "location", "parent", "top", ],`

orta:
For someone looking at this, the change means that reads to `name` will now give an error

```
declare const name2: never;
declare const name3: void;

declare function f(str: string): void

f(name2) // Fine
f(name3) // Error
```

PR #884, opened 2020-07-16
Create deprecatedjson from mdn web

ShuiRuTian:
This is the first step to fix https://github.com/microsoft/TypeScript/issues/39514. 

Now mostly use MDN message.


ShuiRuTian:
Some QA:
1.Q: why deprecated.json added a lot of api, but in d.ts only change 6 lines?
A: Some apis in deprecated.json does not exist in d.ts. 
And (we have come to code details) WebIdl.typedefs does not have "comment", it could only be marked as "deprecated". This is indeed could be improved. 
But, in the other way, it is strange, 
for example, there is `"MouseWheelEvent": "Do not use this interface for wheel events.Like MouseScrollEvent, this interface is non-standard and deprecated. It was used in non-Gecko browsers only. Instead use the standard WheelEvent.\n",`
And, now come to d.ts, type MouseWheelEvent = WheelEvent, so is there really difference?

2. Q: Why use mdn rather than W3C?
A: Usually W3C give some more detail messages, but it not have uniform fotmat and I am not familiar with it, which make it diffcult to extract which type is deprecated and the relative info.

Welcome to ask more question! Or is there somewhere I do not explain clearly enough? Just feel free to share your opinion!

README.md
```diff
@@ -103,3 +103,4 @@ To give you a sense of whether we will accept changes, you can use these heurist
 - `overridingTypes.json`: types that are defined in the spec file but has a better or more up-to-date definitions in the json files.
 - `removedTypes.json`: types that are defined in the spec file but should be removed.
 - `comments.json`: comment strings to be embedded in the generated .js files.
+- `deprecatedMessage.json`: the reason why one type is deprecated. The reason why it is a sprecate file rather than merge in comment.json is mdn/apiDescriptions.json would also possiably be deprecated.
```
ghostd, line 106: Did you mean `sprecate` -> `separated`?

ShuiRuTian: whoops, you are right, thanks!

manualAnalytics/README.md
```diff
@@ -0,0 +1,7 @@
+# READ ME!
+All files in this folder aimes to be used by USER rather than CODE.
```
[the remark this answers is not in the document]
sandersn: Actually, deprecatedMessage.json only has 32 elements, and is likely to grow slowly. I think that future updates can be manual, and that the entire manualAnalytics folder doesn't need to be checked in to the repo.

Removing it makes this PR a lot smaller and easier to understand.

src/index.ts
```diff
@@ -110,6 +111,19 @@ function emitDom() {
         return idl;
     }
 
+    function mergeDeprecatedMessage(idl: Browser.WebIdl, descriptions: Record<string, string>) {
+        const namespaces = arrayToMap(idl.namespaces!, i => i.name, i => i);
+        for (const [key, value] of Object.entries(descriptions)) {
+            const target = idl.interfaces!.interface[key] || namespaces[key];
+            if (target) {
+                const tmp = target.comment ?? "";
+                const tmp2 = "\n * @deprecated " + transformVerbosity(key, value);
```
sandersn, line 120: `comment` and `deprecated` would be nicer names for the variables

package.json
```diff
@@ -19,6 +19,7 @@
     "node-fetch": "^2.6.0",
     "print-diff": "^0.1.1",
     "styleless-innertext": "^1.1.2",
+    "ts-node": "^8.10.2",
```
sandersn, line 22: probably don't need ts-node if manualAnalytics is gone, right?

sandersn (changes requested):
Looks good, but please remove manualAnalytics/* and related changes.

package.json
ShuiRuTian: Agree.

.gitignore
```diff
@@ -281,6 +281,7 @@ __pycache__/
 generated/
 lib/
 inputfiles/browser.webidl.json
-.vscode
+.vscode/*
+!.vscode/launch.template.json
```
saschanaz, line 285: I don't think this does anything here 👀

inputfiles/deprecatedMessage.json
```diff
@@ -0,0 +1,33 @@
+{
```
saschanaz, line 1: I don't see any automatic generator for this file. Is this file handcrafted?

ShuiRuTian: No, it is created by a script. It is why I change the .gitignore file ---- I did add the template file, for running the script by vscode quickly.
But sandersn think this file would not update frequently. So it could be maintained manually like others. You could see the previews for more info.

orta (approved):
We think this makes sense as-is 👍🏻

saschanaz:
MDN BCD provides deprecation flags (#915), and then we won't need this. 👀

PR #885, opened 2020-07-22
Replaces several DOM methods with weaker typings and closes microsoft/TypeScript#4689

jun-sheaf:
The following changes have been made:
* All occurences of `getElementById` take an optional type parameter for manual typing.
* All occurences of `getElementsByClassName` take an optional type parameter for manual typing.
* `parentElement` now types to `Element` (See microsoft/TypeScript#4689)


HolgerJeromin:
This will break many code parts. 
Perhaps 
`getElementById<E extends Element = HTMLElement>(elementId: string): E | null;`
will break less.

jun-sheaf:
> This will break many code parts. 
> 
> Perhaps 
> 
> `getElementById<E extends Element = HTMLElement>(elementId: string): E | null;`
> 
> will break less.

I was thinking about this, but the spec states getElementById should return an Element. A simple example is using it on SVGs.

orta:
Yeah, I agree with @HolgerJeromin - I think the move to generics is a good call for usability outside of the common HTML case, but I don't think it should switch to `Element`.

I'll write a longer post in the original TS issue but if this switches to `getElementById<E extends Element = HTMLElement>(elementId: string): E | null;` then I'm cool with getting this merged.

orta:
https://github.com/microsoft/TypeScript/issues/4689#issuecomment-690503791

jun-sheaf:
@orta Sorry for not getting back after a while. I've made the changes you proposed. Your arguments are reasonable, particularly the fact about a possible 'strict-type' module if every necessary.

HolgerJeromin:
Perhaps the parent case can be more complex handled:

forreignObjectHTMLElement parent is svgSvgElement.
The rest HTMLElement parent have HTMLElement

SvgSvgElement can have svgSvgElement or HTMLElement as parent.
All the rest SvgElement have SvgElement as parents.

Or do i miss something?

jun-sheaf:
Sounds reasonable! I'll make these changes.

jun-sheaf:
So the changes made so far:

- All occurrences of `getElementById` take an optional type parameter for manual typing with default as `HTMLElement`.
- All occurrences of `getElementsByClassName` take an optional type parameter for manual typing with default using case-by-case.
  - `Element.getElementsByClassName` returns `HTMLElement` by default.
  - `SVGElement.getElementsByClassName` returns `SVGElement` by default.
- `parentElement` is case-by-case (See microsoft/TypeScript#4689)
  - `Element.parentElement` types to `Element`.
  - `HTMLElement.parentElement` types to `HTMLElement`.
  - `SVGElement.parentElement` types to `SVGElement`.

inputfiles/idl/DOM.widl
```diff
@@ -214,7 +214,7 @@ interface Node : EventTarget {
   readonly attribute Document? ownerDocument;
   Node getRootNode(optional GetRootNodeOptions options = {});
   readonly attribute Node? parentNode;
-  readonly attribute Element? parentElement;
+  readonly attribute HTMLElement? parentElement;
```
HolgerJeromin, line 217: Is this a change from upstream?

jun-sheaf: Thanks for catching this! Happened during a RegExp replace operation..

HolgerJeromin:
Can you add similar typings to `querySelector`, `querySelectorAll` and `closest`?

jun-sheaf:
Scope creep. There is already another discussion for that since the new template string types will change the playing field.

baselines/dom.generated.d.ts
```diff
@@ -4761,16 +4761,16 @@ interface Document extends Node, DocumentAndElementEventHandlers, DocumentOrShad
...
      * Returns a HTMLCollection of the elements in the object on which the method was invoked (a document or an element) that have all the classes given by classNames. The classNames argument is interpreted as a space-separated list of classes.
      */
-    getElementsByClassName(classNames: string): HTMLCollectionOf<Element>;
+    getElementsByClassName<E extends Element = HTMLElement>(classNames: string): HTMLCollectionOf<E>;
     /**
      * Gets a collection of objects based on the value of the NAME or ID attribute.
      * @param elementName Gets a collection of objects based on the value of the NAME or ID attribute.
      */
-    getElementsByName(elementName: string): NodeListOf<HTMLElement>;
+    getElementsByName<E extends Element = Element>(elementName: string): NodeListOf<E>;
```
orta, line 4773: We think this should should go back `HTMLElement` - https://github.com/microsoft/TypeScript/issues/7709 ( of all of the `getElementsX` this one is the least controversial as it can only be HTMLElements anyway)

Other than that, then we're good to merge 👍🏻

jun-sheaf: @orta Changes made! Ready for approval.

orta:
Thanks!

orta:
#970 reverts this PR - so that we can release 4.2 without this.  

The revert isn't permanent, but it is too breaking to DefinitelyTyped and so it's likely to be very breaking to the rest of the TS/JS ecosystem and Anders argued that this technique is probably not what we want either way https://github.com/microsoft/TypeScript/pull/42067#issuecomment-754860862

It's going to need another look (and maybe we need to figure out a way to be able to determine breakability better inside PRs in this repo, because it takes a few steps to find out right now).

felixfbecker:
It is very weird to me that the type parameter would default to `HTMLElement`.
`document.querySelector('#id')` is typed to return `Element` in `dom.d.ts`, why should `document.getElementById('id')` return `HTMLElement` by default (without any casts or type parameters from the user)? That seems incorrect and inconsistent...

I really like that the TypeScript DOM types are a trustworthy reference for how the types in the DOM actually are, often helping uncover and teaching cases that someone may not even know about (and much easier to read and more discoverable than the DOM spec). I think it would be a shame if TypeScript types moved away from modeling the _correct_ types towards only modeling the _majority-of-cases_ types.

It is actually very easy to accidentally write code like e.g.
```ts
document.getElementById(location.hash.slice(1))?.focus()
```
trusting TypeScript that it would give a type error if this was not correct and forgetting that a user-provided ID could very well match an SVG element, which would not have a `.focus()` method and result in a runtime `Cannot call method 'focus' of undefined` error. Imo, that is an error that should be caught by TypeScript, unless the user overrode it with a cast.

I understand the challenges with backwards compatibility, but as a user I'd rather update code to add casts to fix type errors. It wouldn't be the first type error in a dom.d.ts upgrade for more correct types 🙂

baselines/dom.generated.d.ts
```diff
@@ -10807,7 +10808,7 @@ interface Node extends EventTarget {
     /**
      * Returns the parent element.
      */
-    readonly parentElement: HTMLElement | null;
+    readonly parentElement: Element | null;
```
trusktr, line 10811: With this in place, the property on `HTMLElement` can be deleted.

baselines/dom.generated.d.ts
```diff
@@ -14392,7 +14396,7 @@ interface SVGSVGElement extends SVGGraphicsElement, DocumentEvent, SVGFitToViewB
     forceRedraw(): void;
     getComputedStyle(elt: Element, pseudoElt?: string | null): CSSStyleDeclaration;
     getCurrentTime(): number;
-    getElementById(elementId: string): Element;
+    getElementById<E extends Element = HTMLElement>(elementId: string): E | null;
```
trusktr, line 14399: `<svg>` elements can be nested inside SVG trees. `Element` seems like a better default.

baselines/dom.generated.d.ts
```diff
@@ -6615,6 +6615,7 @@ interface HTMLElement extends Element, DocumentAndElementEventHandlers, ElementC
     readonly offsetParent: Element | null;
     readonly offsetTop: number;
     readonly offsetWidth: number;
+    readonly parentElement: HTMLElement | null;
```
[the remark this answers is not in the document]
jun-sheaf: Please read the entire thread and  #1151 before repeating comments.

baselines/dom.generated.d.ts
```diff
@@ -4761,11 +4761,11 @@ interface Document extends Node, DocumentAndElementEventHandlers, DocumentOrShad
      * Returns a reference to the first object with the specified value of the ID attribute.
      * @param elementId String that specifies the ID value.
      */
-    getElementById(elementId: string): HTMLElement | null;
+    getElementById<E extends Element = HTMLElement>(elementId: string): E | null;
```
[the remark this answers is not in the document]
jun-sheaf: Please read the entire thread and  #1151 before repeating comments.

baselines/dom.generated.d.ts
```diff
@@ -4961,7 +4961,7 @@ interface DocumentEvent {
 /** A minimal document object that has no parent. It is used as a lightweight version of Document that stores a segment of a document structure comprised of nodes just like a standard document. The key difference is that because the document fragment isn't part of the active document tree structure, changes made to the fragment don't affect the document, cause reflow, or incur any performance impact that can occur when changes are made. */
 interface DocumentFragment extends Node, NonElementParentNode, ParentNode {
     readonly ownerDocument: Document;
-    getElementById(elementId: string): HTMLElement | null;
+    getElementById<E extends Element = HTMLElement>(elementId: string): E | null;
```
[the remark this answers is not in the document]
jun-sheaf: Please read the entire thread and  #1151 before repeating comments.

baselines/dom.generated.d.ts
```diff
@@ -5147,7 +5147,7 @@ interface Element extends Node, Animatable, ChildNode, InnerHTML, NonDocumentTyp
     /**
      * Returns a HTMLCollection of the elements in the object on which the method was invoked (a document or an element) that have all the classes given by classNames. The classNames argument is interpreted as a space-separated list of classes.
      */
-    getElementsByClassName(classNames: string): HTMLCollectionOf<Element>;
+    getElementsByClassName<E extends Element = HTMLElement>(classNames: string): HTMLCollectionOf<E>;
```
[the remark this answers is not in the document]
jun-sheaf: Please read the entire thread and  #1151 before repeating comments.

baselines/dom.generated.d.ts
jun-sheaf: Please read the entire thread and  #1151 before repeating comments.

baselines/dom.generated.d.ts
jun-sheaf: Please read the entire thread and  #1151 before repeating comments.

PR #669, opened 2019-03-16
Update Streams to spec version 46c3b89

MattiasBuelens:
This updates the type definitions for Streams to [spec version `46c3b89` of 7 November 2018](https://github.com/whatwg/streams/tree/46c3b89dd3aff28b2fc381dd1d397c12b4fb8a16) ([spec text](https://streams.spec.whatwg.org/commit-snapshots/46c3b89dd3aff28b2fc381dd1d397c12b4fb8a16/)).

* `PipeOptions` now takes an optional `AbortSignal`, to allow aborting an ongoing pipe operation started with `ReadableStream.pipeTo()` or `.pipeThrough()`

This also removes the (accidentally exposed) `ReadableStreamBYOBReader` constructor. [The spec defines](https://streams.spec.whatwg.org/commit-snapshots/46c3b89dd3aff28b2fc381dd1d397c12b4fb8a16/#globals) that this constructor should not be exposed.

PR #890, opened 2020-07-28
Update Streams to official WebIDL

### LINKED ISSUE #886: Update Streams API to match the latest IDL
https://github.com/whatwg/streams/pull/1035 did a massive work to introduce Web IDL, and that made our IDL outdated in complex way. Updating our code to use the upstream IDL will ease getting further update from this spec.

Pinging @MattiasBuelens who might be interested.

MattiasBuelens:
Update the types for [the Streams standard](https://streams.spec.whatwg.org/) to use the official WebIDL definitions.

I'm already opening this PR to have a better idea of what sort of (breaking) changes this would cause, and to continue the discussion in #886. Here's what I did so far:
* Aligned all type names with the WebIDL definitions. This includes things like `PipeOptions` ➡️ `StreamPipeOptions` and `ReadableStreamErrorCallback` ➡️ `UnderlyingSourceCancelCallback`. This is *technically* a breaking change, but I don't expect many users will be using these dictionary or callback types directly.
* Removed everything related to readable byte streams. This feature is not implemented in any browser so far. We can easily bring the types back once this feature starts shipping.
* Removed the async iterable declaration from `ReadableStream`. This was not yet part of the generated types, but it is part of the new IDL. Currently, we would incorrectly recognize this as a regular iterable declaration, which would result in incorrect generated code (e.g. a `forEach` method that doesn't exist in practice). See #669 and microsoft/TypeScript#29867.
* Removed `WritableStream.close`. This method did not yet exist in the type definitions, but I explicitly prevent it from being added. We'll wait until more browsers implement it before adding it (by "removing the removal"). See #827.
* Renamed `ReadableStreamReadResult` to `ReadableStreamDefaultReadResult`. The standard also defines `ReadableStreamBYOBReadResult` but, again, I removed byte stream support.
* Refined the `done: true` case for `ReadableStreamDefaultReadResult` to *always* has `value: undefined`. The only case when you could have a value is when using a BYOB reader, which we do not want to support yet.
* Changed `ReadableStreamReader` to be a type alias for `ReadableStreamDefaultReader`. In the future, this will become a type union to include `ReadableStreamBYOBReader`, [as per spec](https://streams.spec.whatwg.org/#typedefdef-readablestreamreader).

I'd like some input regarding what we should and should not change.
* Is it okay to rename types if our initial names (before they were defined in the standard) do not align with the names from the (new) standard? Or should we add type aliases for the old names, to allow for a more gradual migration?
* Should we keep `WritableStream`, `TransformStream` and `pipeTo`/`pipeThrough`, even though only Chromium has implemented them? These types are already used in several projects, and it would be a bit of a shame if they'd need to go back to using third-party type definitions for these.
* Is it okay to remove readable byte stream support? Unlike `WritableStream` and `TransformStream`, this feature is actually not supported *anywhere*, so I don't expect many users to be using these types in practice.

Depends on #920.

Fixes #886.


inputfiles/idl/Streams.commentmap.json
```diff
@@ -0,0 +1,7 @@
...
+    "readablestreamiteratoroptions-preventcancel": "Asynchronously iterates over the chunks in the stream's internal queue.\n\nAsynchronously iterating over the stream will lock it, preventing any other consumer from acquiring a reader. The lock will be released if the async iterator's return() method is called, e.g. by breaking out of the loop.\n\nBy default, calling the async iterator's return() method will also cancel the stream. To prevent this, use the stream's values() method, passing true for the preventCancel option.",
+    "queuingstrategyinit-highwatermark": "Creates a new ByteLengthQueuingStrategy with the provided high water mark.\n\nNote that the provided high water mark will not be validated ahead of time. Instead, if it is negative, NaN, or not a number, the resulting ByteLengthQueuingStrategy will cause the corresponding stream constructor to throw."
+}
```
MattiasBuelens, line 7: This comment map is very broken. The Streams standard does have comments, but it doesn't look like `idlfetcher` understands the comment style.

Do we want to fix `idlfetcher` to handle the comment style of the Streams standard? Or do we want to (somehow) disable comment maps for Streams, and leave them undocumented for now?

saschanaz: The comment generator currently has some issues on other specs too, we probably want to fix it.

MattiasBuelens: I'm trying a couple of things to improve it. Expect a separate PR soon. 😁

MattiasBuelens:
This is (finally) ready for review. 😄 
* I've updated the IDL for Streams and Encoding, since the `GenericTransformStream` mixin has moved from Encoding to Streams. (whatwg/streams#1073)
* I noticed that #915 removes the types for readable byte streams, which I also did in this PR. If that PR were to land first, I'll rebase this one.

Apologies for the large number of force-pushes! I try to keep the commits linear and easy to review, which means a lot of rebases to stay up-to-date with master. 😅

sandersn (approved):
After discussing the intent of the PR, I think we should take this. Streams doesn't seem *well* supported, but we already include it, so it should be kept up to date. (And @orta suspects that it's used by some popular tools even if it's V8-only.)

The changes themselves seem ok? They look reasonable for how little knowledge I have of the standard.

trxcllnt:
@MattiasBuelens now that Chrome [ships with Byte Streams enabled](https://www.chromestatus.com/feature/4535319661641728), can we resurrect them in the TS typings?

MattiasBuelens:
@trxcllnt Good point! We initially removed the types for readable byte streams since they weren't implemented by *any* browser yet, but now they have shipped in Chromium-based browsers.

I'll look into it. 🙂

saschanaz:
Actually no, it needs to be on [*at least two engines*](https://github.com/microsoft/TypeScript-DOM-lib-generator#additions) to be enabled.

MattiasBuelens:
> Actually no, it needs to be on [_at least two engines_](https://github.com/microsoft/TypeScript-DOM-lib-generator#additions) to be enabled.

Right, I remember now! Guess we'll have to wait for Firefox and Safari to show some love for readable byte streams. 😅

trxcllnt:
[caniuse reports](https://caniuse.com/mdn-api_readablestreambyobreader) byte streams are enabled for 65% of browsers. Yes they're all based on Blink, but (if accurate) that's still a huge number.

The main issue is the types existed at one point and now don't. We wrote [libs](https://github.com/apache/arrow/) anticipating native byte streams would ship, told folks to use @MattiasBuelens's `web-streams-polyfill` in the meantime, and now can't update TS compilers because the types are gone.

I understand updating as changes are made in the whatwg spec, but they haven't been removed there. Perhaps they never should've been added here, but that ship has sailed.

MattiasBuelens:
I understand your frustration. With types for readable byte streams removed from TypeScript 4.2 and [`@types/whatwg-streams` deprecated](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38785), users can no longer find these type definitions anywhere.

In retrospect, we should have spotted that this feature wasn't sufficiently supported by browsers, and kept those types out when we added streams to TypeScript's DOM types. But as you said, that ship has sailed. 😞

Is it possible to resurrect `@types/whatwg-streams` after it has been deprecated? Or should we perhaps create a separate `@types` package that *only* adds readable byte streams?

MattiasBuelens:
> Or should we perhaps create a separate `@types` package that _only_ adds readable byte streams?

It looks like this is the idea of #1023. I'm already a fan! 😁

domoritz:
@MattiasBuelens what do you recommend us to do now if we would like to use byte streams? Can you resurrect `@types/whatwg-streams` or make a new `@types` package for readable streams? If we do that and import that package into our package, does that mean we will break in new versions of TS when byte streams are added back to TS? I'm happy to help if you give me some guidance.

PR #892, opened 2020-08-04
Add keyCode attribute to KeyboardEventInit

taejs:
It's needed to fix cross browsing issue(support IE)

Fixes https://github.com/microsoft/TypeScript/issues/15228


HolgerJeromin:
Thanks for helping.

taejs:
Thanks a lot! I finally got how it goes

saschanaz:
I think this type of change should go to DefinitelyTyped with the name something like `@types/ie`.

orta:
Yeah, we agree - let's get this in

PR #893, opened 2020-08-04
Update sources

MattiasBuelens:
This collects the latest updates to the IDL definitions and MDN API descriptions.

Some tooling changes were necessary to deal with a few updated specifications:
* Specifications such as [Screen Orientation](https://www.w3.org/TR/screen-orientation/) seem to be using a newer version of the W3C spec template, which now includes a `<span class="idlHeader">WebIDL</span>` element at the top of every WebIDL block (for example in [this block](https://www.w3.org/TR/screen-orientation/#dom-screen)). With the current code, we would naively include this header in the IDL content, which broke the WebIDL parser. We now strip this header when fetching new IDLs.
* Web IDL has switched from `void` to `undefined`, see heycam/webidl#60 and heycam/webidl#906. Many specifications have already made the switch, so we need to support this in the emitter.

In order to facilitate review, I've done two baseline update commits: first after running `fetch-idl`, and then after running `fetch-mdn`. If you prefer, I can squash these into one commit.


baselines/dom.generated.d.ts
```diff
@@ -11081,28 +11122,60 @@ declare var Notification: {
 interface OES_element_index_uint {
 }
 
+declare var OES_element_index_uint: {
```
MattiasBuelens, line 11126: Hmm, I suppose we would first need #873 to land? These now have `[LegacyNoInterfaceObject]` in their IDL...

sandersn: #873 is now merged.

src/helpers.ts
```diff
@@ -15,6 +15,7 @@ export const baseTypeConversionMap = new Map<string, string>([
     ["object", "any"],
     ["sequence", "Array"],
     ["record", "Record"],
+    ["undefined", "void"],
```
MattiasBuelens, line 18: Web IDL has switched from `void` to `undefined`, see heycam/webidl#60 and heycam/webidl#906. Many specifications have already made the switch, so we need to support this.

For now, I'm mapping this back to TypeScript's `void` type. However, we could also move this into `sameTypes` and actually emit `undefined` in the generated type definitions. Thoughts? 🙂

inputfiles/overridingTypes.json
```diff
@@ -3096,7 +3092,7 @@
     "typedefs": {
         "typedef": [
             {
-                "override-type": "Blob | BufferSource | FormData | URLSearchParams | ReadableStream<Uint8Array> | string",
+                "override-type": "ReadableStream<Uint8Array> | XMLHttpRequestBodyInit",
```
MattiasBuelens, line 3095: `BodyInit` is now [defined](https://fetch.spec.whatwg.org/#bodyinit-unions) as:
```
typedef (Blob or BufferSource or FormData or URLSearchParams or USVString) XMLHttpRequestBodyInit;

typedef (ReadableStream or XMLHttpRequestBodyInit) BodyInit;
```
This change re-aligns our `override-type` with the new definition.

inputfiles/idlSources.json
```diff
@@ -108,6 +108,10 @@
         "url": "https://www.w3.org/TR/css-overscroll-1/",
         "title": "CSS Overscroll Behavior"
     },
+    {
+        "url": "https://www.w3.org/TR/css-pseudo-4/",
+        "title": "CSS Pseudo-Elements"
```
MattiasBuelens, line 113: This is [being used by the Web Animation specification](https://www.w3.org/TR/web-animations-1/#extensions-to-the-pseudoelement-interface), so we need to include its types. Fortunately, [there aren't that many new types](https://www.w3.org/TR/css-pseudo-4/#idl-index).

MattiasBuelens (commented):
Rebased and reworked! 😁

Note that I cheated a bit with the "Update IDL" commit: I updated all IDLs *except* for Streams. These need more tweaking and cleanup in our overrides, which I'm doing in #890. 😉

src/helpers.ts
sandersn: I like this change except in unions. There is one in CustomElementRegistry:

```
    get(name: string): CustomElementConstructor | void;
```

You can't use `?.` from the value returned by the accessor. This is admittedly a failing of Typescript, but `T | void` is rare type for all `T`.

Not sure whether it's easier to fix the emitter or to add an overridingType entry for the accessor

inputfiles/mdn/apiDescriptions.json
```diff
@@ -48,49 +48,48 @@
...
   "DataTransferItemList": "The DataTransferItemList object is a list of DataTransferItem objects representing items being dragged. During a drag operation, each DragEvent has a dataTransfer property and that property is a DataTransferItemList.",
   "DedicatedWorkerGlobalScope": "The DedicatedWorkerGlobalScope object (the Worker global scope) is accessible through the self keyword. Some additional global functions, namespaces objects, and constructors, not typically associated with the worker global scope, but available on it, are listed in the JavaScript Reference. See also: Functions available to workers.",
   "DelayNode": "The DelayNode interface represents a delay-line; an AudioNode audio-processing module that causes a delay between the arrival of an input data and its propagation to the output.",
-  "DeviceAcceleration": "A DeviceAcceleration object provides information about the amount of acceleration the device is experiencing along all three axes.",
+  "DeviceAcceleration": "REDIRECT  DeviceMotionEventAcceleration",
```
sandersn, line 69: can this be turned into `@deprecated` instead of, or in addition to, REDIRECT?

baselines/dom.generated.d.ts
```diff
@@ -5771,6 +5819,10 @@ interface GlobalEventHandlersEventMap {
     "transitionstart": TransitionEvent;
     "volumechange": Event;
     "waiting": Event;
+    "webkitanimationend": Event;
+    "webkitanimationiteration": Event;
+    "webkitanimationstart": Event;
+    "webkittransitionend": Event;
```
sandersn, line 5825: this is probably fine, but it's slightly weird to me that webkit* event handlers are being added. Isn't that webkit-specific?

sandersn (changes requested):

src/helpers.ts
MattiasBuelens: I can add a special case in the emitter to replace `void` with `undefined` if it appears in a union type. Right now, it's just for this one method in `CustomElementRegistry`, but maybe similar methods pop up in the future for which this would apply. 🙂

baselines/dom.generated.d.ts
MattiasBuelens: Apparently these have been added to the HTML standard itself, see whatwg/html#5188. 🤷

Should I add them to `removedTypes.json`? I see there's already a couple of WebKit-specific removals in there.

inputfiles/mdn/apiDescriptions.json
MattiasBuelens: I could change [`mergeApiDescriptions()`](https://github.com/microsoft/TSJS-lib-generator/blob/6c7c2ea7df1105a3bb9749f6b6ba22c7ca817c90/src/index.ts#L96) to first check if the comment from MDN starts with "REDIRECT", and if so ignore the comment and mark it as deprecated instead.

On another note, is there a reason why [`Interface`](https://github.com/microsoft/TSJS-lib-generator/blob/6c7c2ea7df1105a3bb9749f6b6ba22c7ca817c90/src/types.d.ts#L151) does not have a `deprecated` flag? I kind of expected [`markAsDeprecated()`](https://github.com/microsoft/TSJS-lib-generator/blob/6c7c2ea7df1105a3bb9749f6b6ba22c7ca817c90/src/helpers.ts#L279) to also mark the entire interface as deprecated.

saschanaz:
Are we confident that this won't introduce any nonexistent or highly experimental APIs? Updating every spec by a single PR makes hard to check them.

inputfiles/idlSources.json
saschanaz: Neither Chrome nor Firefox provides a stable CSSPseudoElement implementation, so I would just suppress this type.

MattiasBuelens:
> Are we confident that this won't introduce any nonexistent or highly experimental APIs? Updating every spec by a single PR makes hard to check them.

Hmm, good point. I basically ran `npm run fetch-idl` and `npm run fetch-mdn`, thinking that this was the usual process. I have not investigated whether all of these new APIs are actually "stable" yet.

I suppose it would be safer to land the tooling changes first, and then create separate PRs for each updated specification, each with their own review?

MattiasBuelens:
Closing this PR in favor of #920, which contains *only* the tooling changes.

PR #894, opened 2020-08-04
Improve comments

MattiasBuelens:
This improves the comment fetching to better find the correct parent and member. Instead of extracting the member name from the ID of the `<dfn>` element, we now parse it from the text content of the element.

We now also emit comments for constructors (when available).

Depends on #920.


inputfiles/idl/HTML - Canvas.commentmap.json
```diff
@@ -3,11 +3,52 @@
...
     "canvas-toblob": "Creates a Blob object representing a file containing the image in the canvas, and invokes a callback with a handle to that object.\n\nThe second argument, if provided, controls the type of the image to be returned (e.g. PNG or JPEG). The default is \"image/png\"; that type is also used if the given type isn't supported. The third argument applies if the type is an image format that supports variable quality (such as \"image/jpeg\"), and is a number in the range 0.0 to 1.0 inclusive indicating the desired quality level for the resulting image.",
     "canvas-transfercontroltooffscreen": "Returns a newly created OffscreenCanvas object that uses the canvas element as a placeholder. Once the canvas element has become a placeholder for an OffscreenCanvas object, its intrinsic size can no longer be changed, and it cannot have a rendering context. The content of the placeholder canvas is updated by calling the commit() method of the OffscreenCanvas object's rendering context.",
-    "context-2d": "Returns the canvas element.",
+    "context-canvas": "Returns the canvas element.",
```
MattiasBuelens, line 6: We now detect more definitions (yay!), but they're still targeting a non-existent `Context` interface.

[This particular definition](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-canvas) has the following HTML:
```html
<p>The <dfn id=dom-context-2d-canvas><code>canvas</code></dfn> attribute must return the
  value it was initialized to when the object was created.</p>
```
The problem is that the `<dfn>` element does not have a `data-dfn-for` attribute. So we try to extract the parent and member from the `id` attribute, but unfortunately it doesn't match the expect `parent-member` format.

It guess this `data-dfn-for` attribute is not required when using [Wattsi](https://github.com/whatwg/wattsi/blob/master/Syntax.md)? So we can't really rely upon it here. 🤷 

I was thinking we may want to try yet *another* approach for retargeting comments. Instead of parsing the `<dfn>` itself, look for an `<a>` inside of an IDL block that links to the same `<dfn>` and try to parse *that*. That's probably gonna cause some other headaches: there might be multiple links, and we'd need to somehow figure out what IDL member corresponds to this link... 😖 Let's maybe leave that for another time? 😛

MattiasBuelens:
I just realized that the only new comments that actually show up in the output are for constructors. So there's nothing exercising the new logic for attributes or methods yet...

I *swear* this will look more impressive when combined with #890! 😆

MattiasBuelens:
Closing for now.

The `fetch-comment` script is [currently broken](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1097#issuecomment-1004099360), and we're not even sure if we want to fix it or build something else. 🤷

PR #902, opened 2020-08-19
Remove redundant name field from *Types.json

saschanaz:
No need to add name field when overriding existing types.


orta:
I'm game for this, but we'll probably need to do it and get it merged after a bunch of the current PRs are merged - it'll cause a lot of conflicts 👍🏻

saschanaz:
Replaced by #1034

PR #905, opened 2020-09-04
Just activate HTMLVideoElement playsinline attribute.

kahirokunn:
Just activate HTMLVideoElement playsinline attribute for solve this problem.
https://webkit.org/blog/6784/new-video-policies-for-ios

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -8952,6 +8952,10 @@ interface HTMLVideoElement extends HTMLMediaElement {
      */
     height: number;
     /**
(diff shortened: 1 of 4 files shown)


saschanaz:
Rebasing/merging will fix the build failure.

orta:
Cool, this is in Safari since 2016, and Chrome as of 2019 - when I tested making it in Firefox:

```ts
const v = document.createElement("video")
undefined
v.playsInline
undefined
```

It doesn't exist there, but I'm good with 2/3 shipping unprefixed 👍🏻

PR #909, opened 2020-09-10
fix(40454): allow IntersectionObserverInit.root to accept type Docume…

watrool:
…nt and update IntersectionObserver.root to be of type Node

Fix https://github.com/microsoft/TypeScript/issues/40454

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -9611,7 +9611,7 @@ declare var InputEvent: {
 
 /** provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport. */
 interface IntersectionObserver {
-    readonly root: Node | null;
+    readonly root: Element | Document | null;
     readonly rootMargin: string;
     readonly thresholds: ReadonlyArray<number>;
     disconnect(): void;
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -611,7 +611,7 @@ interface IntersectionObserverEntryInit {
 }
 
 interface IntersectionObserverInit {
-    root?: Element | null;
+    root?: Element | Document | null;
     rootMargin?: string;
     threshold?: number | number[];
 }
@@ -9611,7 +9611,7 @@ declare var InputEvent: {
 
 /** provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport. */
 interface IntersectionObserver {
-    readonly root: Element | null;
+    readonly root: Node | null;
     readonly rootMargin: string;
     readonly thresholds: ReadonlyArray<number>;
     disconnect(): void;
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -16,7 +16,8 @@ export const baseTypeConversionMap = new Map<string, string>([
     ["sequence", "Array"],
     ["record", "Record"],
     ["FrozenArray", "ReadonlyArray"],
-    ["EventHandler", "EventHandler"]
+    ["EventHandler", "EventHandler"],
+    ["undefined", "void"]
 ]);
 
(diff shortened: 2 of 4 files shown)


baselines/dom.generated.d.ts
```diff
@@ -9607,7 +9607,7 @@ declare var InputEvent: {
 
 /** provides a way to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport. */
 interface IntersectionObserver {
-    readonly root: Element | null;
+    readonly root: Node | null;
```
watrool, line 9610: image

baselines/dom.generated.d.ts
```diff
@@ -611,7 +611,7 @@ interface IntersectionObserverEntryInit {
 }
 
 interface IntersectionObserverInit {
-    root?: Element | null;
+    root?: Document | Element | null;
```
watrool, line 614: image

saschanaz:
The right way to do this is updating `idlSource.json` to point https://w3c.github.io/IntersectionObserver/ and running `npm run fetch-idl "Intersection Observer" && npm run build && npm run baseline-accept`, so that we can keep syncing with the upstream changes.

watrool:
@saschanaz thanks for the comment! I tried that approach, but it seems `https://w3c.github.io/IntersectionObserver/ ` changed all `void` type to `undefined` type 
image
which makes the script throws
```
Error: Unknown DOM type: undefined
    at convertDomTypeToTsTypeSimple (/Users/troytao/TSJS-lib-generator/lib/emitter.js:301:15)
    at convertDomTypeToTsTypeWorker (/Users/troytao/TSJS-lib-generator/lib/emitter.js:239:28)
    at convertDomTypeToTsType (/Users/troytao/TSJS-lib-generator/lib/emitter.js:233:22)
    at emitSignature (/Users/troytao/TSJS-lib-generator/lib/emitter.js:598:26)
    at /Users/troytao/TSJS-lib-generator/lib/emitter.js:610:45
    at Array.forEach (<anonymous>)
    at emitSignatures (/Users/troytao/TSJS-lib-generator/lib/emitter.js:610:30)
    at emitMethod (/Users/troytao/TSJS-lib-generator/lib/emitter.js:590:13)
    at /Users/troytao/TSJS-lib-generator/lib/emitter.js:620:31
    at Array.forEach (<anonymous>)
```

Is it ok if I add a `['undefined', 'void']` to `baseTypeConversionMap` in `emitter.ts` to fix this ?

saschanaz:
> Is it ok if I add a `['undefined', 'void']` to `baseTypeConversionMap` in `emitter.ts` to fix this ?

Gread find! Yes, that will fix it.

watrool:
@saschanaz updated!

saschanaz (approved):
LGTM 👍

watrool:
@sandersn poke you for merging this PR ...

src/helpers.ts
```diff
@@ -16,7 +16,8 @@ export const baseTypeConversionMap = new Map<string, string>([
     ["sequence", "Array"],
     ["record", "Record"],
     ["FrozenArray", "ReadonlyArray"],
-    ["EventHandler", "EventHandler"]
+    ["EventHandler", "EventHandler"],
+    ["undefined", "void"]
```
saschanaz, line 20: BTW, #920 did this in another way so we don't need this anymore. Sorry, but still was a good work!

saschanaz:
Could you run `npm run fetch-idl "Intersection Observer" && npm run build && npm run baseline-accept` once again? I think that will fix the attribute type also to `Element | Document | null`.

watrool:
@saschanaz thanks for notifying! I just rebased the PR, removed the change in `helpers.ts` and ran `npm run fetch-idl "Intersection Observer" && npm run build && npm run baseline-accept`.

saschanaz:
The TypeScript team will review this when the "review time" comes, so please wait until that happens 🙏

sandersn (approved):

PR #915, opened 2020-09-16
Use BCD to autoremove unimplemented features

saschanaz:
This patch focuses only on removing totally unimplemented features, and defers everything else:

* features only with single implementation should be reevaluated later
* features that are widely-supported but over-restricted by removedTypes.json should also be reevaluated later.

`forceKeepAlive` is quite huge now but it eventually will be removed once reevaluation ends.


baselines/webworker.generated.d.ts
```diff
@@ -2737,29 +2708,6 @@ declare var ReadableStream: {
     new<R = any>(underlyingSource?: UnderlyingSource<R>, strategy?: QueuingStrategy<R>): ReadableStream<R>;
 };
 
-interface ReadableStreamBYOBReader {
```
saschanaz, line 2740: @MattiasBuelens, could you review the Streams part of this removal? Thanks!

saschanaz:
Oops, test fails 👀

baselines/webworker.generated.d.ts
MattiasBuelens: `UnderlyingByteSource`, `ReadableByteStreamControllerCallback` and [this `ReadableStream` constructor overload](https://github.com/microsoft/TSJS-lib-generator/blob/0b6caa7b6a1add40024da034863ae8f6fa6a2bf6/baselines/dom.generated.d.ts#L12345) should be removed as well.

You might want to have a look at 361a87076aef97268105f9b8f85c686d701df69e (part of #890).

MattiasBuelens: Hang on, where did `WritableStream` and `TransformStream` go? There are still types referencing them (like `GenericTransformStream` and `ReadableStream.pipeTo`).

(I understand why they would need to be removed, but it still makes me sad to see them go... 😞)

HolgerJeromin (approved):

baselines/webworker.generated.d.ts
saschanaz: The bright side is that this patch only disables some of your work but not removes them all. It can be enabled again when browsers implement them 👍

saschanaz: BTW, WritableStream and TransformStream are still alive as there is at least a single implementation and thus not covered here. This patch is intentionally only removes totally unimplemented ones to reduce the diff size.

saschanaz:
Filed a set of PRs:

https://github.com/mdn/browser-compat-data/issues/6733
https://github.com/mdn/browser-compat-data/issues/6732
https://github.com/mdn/browser-compat-data/issues/6731
https://github.com/mdn/browser-compat-data/issues/6730
https://github.com/mdn/browser-compat-data/issues/6729
https://github.com/mdn/browser-compat-data/issues/6728
https://github.com/mdn/browser-compat-data/issues/6727
https://github.com/mdn/browser-compat-data/issues/6726
https://github.com/mdn/browser-compat-data/issues/6725
https://github.com/mdn/browser-compat-data/issues/6724
https://github.com/mdn/browser-compat-data/issues/6723
https://github.com/mdn/browser-compat-data/issues/6722
https://github.com/mdn/browser-compat-data/issues/6720
https://github.com/mdn/browser-compat-data/issues/6719
https://github.com/mdn/browser-compat-data/issues/6715
https://github.com/mdn/browser-compat-data/issues/6709
https://github.com/mdn/browser-compat-data/issues/6708
https://github.com/mdn/browser-compat-data/issues/6705
https://github.com/mdn/browser-compat-data/issues/6704
https://github.com/mdn/browser-compat-data/issues/6703
https://github.com/mdn/browser-compat-data/issues/6702
https://github.com/mdn/browser-compat-data/issues/6701
https://github.com/mdn/browser-compat-data/issues/6700
https://github.com/mdn/browser-compat-data/issues/6681
https://github.com/mdn/browser-compat-data/issues/6663

foolip:
@vinyldarkscratch this PR is of interest to our ongoing project. In particular the list of BCD bugs in https://github.com/microsoft/TSJS-lib-generator/pull/915#issuecomment-695333536 are things that can hopefully be fixed by using https://github.com/foolip/mdn-bcd-collector. Where they can't, that might reveal a deficiency in our approach.

baselines/webworker.generated.d.ts
MattiasBuelens: Streams changes LGTM. 👍 Happy to see `WritableStream` and `TransformStream` are still alive! 😁

I think it's best to wait with landing #890 until after this PR lands. I'll see if I can rebase my PR to not *completely* remove the readable byte stream definitions, and instead rely on your work to auto-remove them. 🙂

saschanaz:
Changes are too big, maybe better to remove things progressively. I'll make separate PRs.

sandersn:
@orta and I like this idea -- in fact, we'd rather take this PR than the dozen or so individual-removal PRs. 

- Do they remove the same set of declarations? It looks that way.
- Is this PR ready to go otherwise? I've only scanned the code, but it looks OK at that level of detail.

sandersn (approved):
Approving the code change -- couple of open questions about the resulting DOM changes.

saschanaz:
> @orta and I like this idea -- in fact, we'd rather take this PR than the dozen or so individual-removal PRs.
> 
>     * Do they remove the same set of declarations? It looks that way.
> 
>     * Is this PR ready to go otherwise? I've only scanned the code, but it looks OK at that level of detail.

* Yes, they are already merged in this PR.
* Also yes. It's using older version of BCD since the PR has been rotting, though.

orta:
Cool, IMO we should let's get this in and into the compiler then 👍🏻

sandersn (approved):

sandersn:
I'm going to port this over to Typescript right now and will tag you (@orta and @saschanaz) on the PR when it's ready.

sandersn:
OK, it is merged into TS now.

PR #919, opened 2020-09-23
Add WebRTC Stats spec

trptcolin:
This proposes bringing in the WebRTC Stats spec. The driver here is updating the `RTCStatsType` enum values, as described in https://github.com/microsoft/TypeScript/issues/36650, but there are other changes that the README-recommended `fetch-idl` process brings in as well. 

My initial thinking was that getting all the spec updates in is preferable, but I'm also not totally sure where browser adoption is for this spec, so I'm not sure what level belongs in TypeScript. Could isolate to just `RTCStatsType` to mitigate risk.

Side note: importing this spec via the README instructions did require a bit of editing to remove the text "WebIDL" in front of each WebIDL block, due to a `.idlHeader` element with that text appearing inside the `.idl` code block. It looks like a number of other specs have a similar thing going on:
- `Gamepad`
- `Push`
- `Screen Orientation`
- `Selection`
- `Web Share`
- and this one, `WebRTC Stats`

I'm happy to open a separate issue to suggest some kind of filtering to avoid similar manual IDL-editing effort in the future if that makes sense - just let me know.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1239,17 +1239,36 @@ interface RTCIceCandidatePair {
 interface RTCIceCandidatePairStats extends RTCStats {
     availableIncomingBitrate?: number;
     availableOutgoingBitrate?: number;
+    bytesDiscardedOnSend?: number;
     bytesReceived?: number;
     bytesSent?: number;
+    circuitBreakerTriggerCount?: number;
+    consentExpiredTimestamp?: number;
+    consentRequestsSent?: number;
+    currentRoundTripTime?: number;
+    currentRtt?: number;
+    firstRequestTimestamp?: number;
+    lastPacketReceivedTimestamp?: number;
+    lastPacketSentTimestamp?: number;
+    lastRequestTimestamp?: number;
+    lastResponseTimestamp?: number;
     localCandidateId?: string;
     nominated?: boolean;
+    packetsDiscardedOnSend?: number;
+    packetsReceived?: number;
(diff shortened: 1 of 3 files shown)


saschanaz:
I guess it's relatively safer to add experimental dictionary fields, as it won't cause "undefined is not a function" error.

For `.idlHeader` issue this should work: https://github.com/microsoft/TSJS-lib-generator/pull/893/commits/97c3235142da30a773535283d93ddb16d289b3f8

orta:
We're struggling with getting the adoption info for this also, querying caniuse for some of the identifiers in here either [just in Firefox](https://caniuse.com/?search=RTCRtpStreamStats) or just [in Safari](https://caniuse.com/?search=RTCStatsReport) - IMO, if you constrain it to just `RTCStatsType` we'll merge but until then we'll probably need to see some larger evidence of use 👍🏻

saschanaz:
Using Searchfox and webkit-search can be helpful for that.

* Firefox: https://searchfox.org/mozilla-central/source/dom/webidl/RTCStatsReport.webidl
* Safari: https://webkit-search.igalia.com/webkit/source/Source/WebCore/Modules/mediastream/RTCStatsReport.idl

I can see at least some fields are supported in Firefox, but not all.

orta:
Thanks @saschanaz - those two sites are really useful (I've added them to the README so we can use it more often) - I think on the whole most of these new fields are in FF/Safari which means it's reasonable to merge 👍🏻

PR #920, opened 2020-09-26
Update tooling for new IDL and MDN

MattiasBuelens:
This PR updates the tooling around extracting and transforming the IDL and MDN sources to support recent changes to the specifications and MDN.
* Specifications such as [Screen Orientation](https://www.w3.org/TR/screen-orientation/) seem to be using a newer version of the W3C spec template, which now includes a `<span class="idlHeader">WebIDL</span>` element at the top of every WebIDL block (for example in [this block](https://www.w3.org/TR/screen-orientation/#dom-screen)). With the current code, we would naively include this header in the IDL content, which broke the WebIDL parser. We now strip this header when fetching new IDLs.
* Web IDL has switched from `void` to `undefined`, see heycam/webidl#60 and heycam/webidl#906. [We chose](https://github.com/microsoft/TSJS-lib-generator/pull/893#discussion_r487332968) to have a bit of special handling for this, to better match the existing semantics and expectations around TypeScript's `void` and `undefined` types:
  * Emit `undefined` if it appears in a union type, i.e. `Element | undefined` instead of `Element | void`.
  * Otherwise we emit `void`, i.e. `void` and `Promise<void>` instead of `undefined` and `Promise<undefined>`.
* When an MDN article for an interface redirects to a different one, we now recognize this as a signal that the interface has been deprecated and should be marked as such in the type definitions.

These tooling changes were extracted from #893, but without the *actual* updates to the IDL and MDN sources. These sources can be updated in separate PRs, [as suggested by @saschanaz](https://github.com/microsoft/TSJS-lib-generator/pull/893#issuecomment-691762130).


src/emitter.ts
```diff
@@ -320,6 +320,18 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor, iterator: boo
         };
     }
 
+    function convertDomTypeToTsTypeInUnion(obj: Browser.Typed): { name: string; nullable: boolean } {
```
saschanaz, line 323: I'm not sure why this is needed, what happens without this?

MattiasBuelens: [See this comment.](https://github.com/microsoft/TSJS-lib-generator/pull/893#discussion_r487332968) In most cases, we want to emit `undefined` as `void`, which is handled by 8dc0f62c0175e1c2ef29ab9bc49a8b5102a625a4. However, if this is a type inside of a union type, we want to special-case it and turn it back into `undefined`.

Perhaps it's a bit confusing that we first transform `undefined` to `void` and then back to `undefined`? Do you want me to try to flip the logic around: use `undefined` by default (i.e. add it to `sameTypes` in [helpers.js](https://github.com/microsoft/TSJS-lib-generator/blob/8dc0f62c0175e1c2ef29ab9bc49a8b5102a625a4/src/helpers.ts#L8)), but add a special case for when it does *not* appear in a union type?

saschanaz: > Do you want me to try to flip the logic around: use `undefined` by default (i.e. add it to `sameTypes` in [helpers.js](https://github.com/microsoft/TSJS-lib-generator/blob/8dc0f62c0175e1c2ef29ab9bc49a8b5102a625a4/src/helpers.ts#L8)), but add a special case for when it does _not_ appear in a union type?

Hmm yes, that sounds more intuitive. 👍

src/emitter.ts
```diff
@@ -285,18 +285,18 @@ export function emitWebIdl(webidl: Browser.WebIdl, flavor: Flavor, iterator: boo
     function convertDomTypeToTsType(obj: Browser.Typed): string {
         if (obj["override-type"]) return obj["override-type"]!;
         if (!obj.type) throw new Error("Missing type " + JSON.stringify(obj));
-        const type = convertDomTypeToTsTypeWorker(obj);
+        const type = convertDomTypeToTsTypeWorker(obj, false);
         return type.nullable ? makeNullable(type.name) : type.name;
     }
 
-    function convertDomTypeToTsTypeWorker(obj: Browser.Typed): { name: string; nullable: boolean } {
+    function convertDomTypeToTsTypeWorker(obj: Browser.Typed, inUnion: boolean): { name: string; nullable: boolean } {
```
MattiasBuelens, line 292: Now I need an extra flag to know whether we're converting a type inside of a union type... 🤷

saschanaz: I have a suggestion: just use `undefined` inside the converter, and in the method/callback emitter we detect whether the conversion result is a simple `undefined` or `Promise<undefined>` and then replace it as `void` / `Promise<void>`.

This is based on an assumption that those two form are tho only proper uses of `void`.

saschanaz: Yay it worked 🎉

MattiasBuelens: That also works! 🙂

Even better, applying this change to #890, we will now generate the following for [`ReadableStreamGenericReader`](https://streams.spec.whatwg.org/#readablestreamgenericreader):
```typescript
interface ReadableStreamGenericReader {
    readonly closed: Promise<undefined>;
}
```
Since `closed` is a property, it keeps its `Promise<undefined>` type, whereas `cancel()` is changed to return `Promise<void>`. I think this is quite neat! 😄

orta (approved):

orta:
Thanks!

I-Cat (commented):
What is 
 "pre.idl:not(.extract):not(.example)", // bikeshed and ReSpec	
"bikeshed" ?

PR #921, opened 2020-10-01
Add Property 'sheet' on type 'SVGStyleElement'

Daliak:
This is a first attempt to fix this issue: https://github.com/microsoft/TypeScript/issues/36896

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -14457,6 +14457,7 @@ declare var SVGStringList: {
 interface SVGStyleElement extends SVGElement {
     disabled: boolean;
     media: string;
+    sheet: CSSStyleSheet;
     title: string;
     type: string;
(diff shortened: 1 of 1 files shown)


DanielRosenwasser (approved):
Looks good, but I don't know 100% how this repo works.

inputfiles/browser.webidl.preprocessed.json
```diff
@@ -74396,6 +74396,13 @@
                             "type": "DOMString",
                             "content-attribute-reflects": 1
                         },
+                        "sheet": {
```
sandersn, line 74399: This is a generated file too, so this change should go in addedTypes.json.

(The format there is slightly different there, I think, unfortunately.)

sandersn (changes requested):
This produces the right output but the code change needs to go in addedTypes.json

saschanaz:
#922 fixed this.

PR #922, opened 2020-10-01
SVGStyleElement implements LinkStyle

conniefan:
SVGStyleElement implements LinkStyle, which will give it the 'sheet' property 
https://github.com/microsoft/TypeScript/issues/36896


baselines/dom.generated.d.ts
```diff
@@ -19025,6 +19024,7 @@ declare var console: Console;
 
 /** Holds useful CSS-related methods. No object with this interface are implemented: it contains only static methods and therefore is a utilitarian interface. */
 declare namespace CSS {
+    function escape(ident: string): string;
     function escape(ident: string): string;
```
sandersn, line 19028: this duplicate `escape` is probably in addedTypes.json or overridingTypes.json, and should now be removed.

sandersn (changes requested):
1. There are two copies of CSS.escape now, so one probably needs to be deleted.
2. caniuse.com only lists Edge as supporting LinkStyle, but doesn't have information for other browsers. Can you figure out if it's available in Firefox or Safari, for example? We only want to add things that are supported by two or more browsers.

baselines/dom.generated.d.ts
conniefan: LinkStyle is already included by this entry in idlSources, so removing the entry from my commit:  {
        "url": "https://drafts.csswg.org/cssom/",
        "title": "CSS Object Model"
    }
That was also the cause of the duplicate CSS.escapes

baselines/dom.generated.d.ts
```diff
@@ -19462,7 +19461,6 @@ interface SVGElementTagNameMap {
     "rect": SVGRectElement;
     "script": SVGScriptElement;
     "stop": SVGStopElement;
-    "style": SVGStyleElement;
```
sandersn, line 19465: should this be deleted? It might be necessary to re-add it in addedTypes.json

baselines/dom.generated.d.ts
```diff
@@ -14454,8 +14454,7 @@ declare var SVGStringList: {
 };
 
 /** Corresponds to the SVG <style> element. */
-interface SVGStyleElement extends SVGElement {
-    disabled: boolean;
```
sandersn, line 14458: should this be deleted? If it's not standard, or not widely implemented, then it's fine to delete, but if not, it would need to be re-added in addedTypes.json

sandersn (changes requested):
Question about two removed properties

baselines/dom.generated.d.ts
conniefan: This website suggests it is not widely used: https://www.chromestatus.com/metrics/feature/timeline/popularity/. But I see it in this spec, https://developer.apple.com/documentation/webkitjs/svgstyleelement. So it will be added back.

baselines/dom.generated.d.ts
conniefan: it is re-added

sandersn (approved):

sandersn:
Thanks @conniefan!

baselines/dom.generated.d.ts
saschanaz: The apple documentation is not the spec. The right one does not has it: https://svgwg.org/svg2-draft/styling.html#InterfaceSVGStyleElement

sandersn: @saschanaz to make sure I understand, we just need to revert the last two commits, 6fa0fbc and f1f5a57, right?

saschanaz: Exactly!

PR #924, opened 2020-10-02
Update Web Speech API to match idl

elyseyao:
Changed the type of the SpeechRecognition event handler for error from ErrorEvent to SpeechRecognitionErrorEvent to match the [API](https://wicg.github.io/speech-api). This adds the types `SpeechRecognitionErrorEvent` and `SpeechRecognitionErrorCode` back in and resolves [this issue](https://github.com/microsoft/TypeScript/issues/37046). 

If it's a potential problem, this reverses some of the changes made [here](https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/1f859ddee3b17f409837694ff89d8d0e7d70f13f).

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -15374,7 +15374,6 @@ interface StorageEvent extends Event {
      * Returns the URL of the document whose storage item changed.
      */
     readonly url: string;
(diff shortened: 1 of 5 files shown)


saschanaz (approved):
Seems okay, as 1) we had them before 2) Blink does have them with webkit prefix (`webkitSpeechRecognitionEvent`, `webkitSpeechRecognitionError`). If someone were using aliases referencing these types, that use case make sense to me.

sandersn (approved):

PR #925, opened 2020-10-02
Added event handlers for gamepadconnected GamepadEvent

elyseyao:
Added event handlers ongamepadconnected and ongamepaddisconnected to add their corresponding events to lib.dom.d.ts and resolve [this issue](https://github.com/microsoft/TypeScript/issues/39425). The event handler names were taken from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/ongamepadconnected).


saschanaz:
This is not the right way to add them, as the spec does not define `ongamepadconnected`/`ongamepaddisconnected`. We need to file a spec issue first.

sandersn (approved):
This is widely supported and the spec change hasn't happened yet. Let's take this now and remove the `addedTypes` override when the spec change gets merged.

PR #926, opened 2020-10-02
Removed incorrect info from getElementById comment

JoshuaKGoldberg:
Fixes #https://github.com/microsoft/TypeScript/issues/40874.


orta (approved):

ulrichstark:
maybe we should leave the notice, that the parameter matches case-SENSITIVE?

JoshuaKGoldberg:
@ulrichstark perhaps, but most parameters are case sensitive, so I think it's probably a safe assumption on the side of the reader. What do you think?

RyanCavanaugh:
Agree that case-sensitive is the default behavior in most programming applications, so no comment is necessary

Issue #908, opened 2020-09-10
Missing API: ResizeObserver


wKich:
Spec: https://drafts.csswg.org/resize-observer/#resize-observer-interface
Mdn: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

MatthiasKunnen:
Will be fixed by #847.

wKich:
@MatthiasKunnen, thank you 🙏

PR #847, opened 2020-04-09
feat: add resize-observer typs

manucorporat:
However, i got an error when generating the types for Resize Observer:
```
  bareMessage: 'Attributes cannot accept sequence types',
  context: 'Syntax error at line 23, since `interface ResizeObserverEntry`:\n' +
    ' attribute sequence<ResizeObserverSize> borderBoxSize;\n' +
```

fixes https://github.com/microsoft/TypeScript/issues/37861

PR #948, opened 2020-10-26
Add resize observer

timocov:
This PR is based on #847, but with replacing the `sequence` type with `FrozenArray` (see https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/847#issuecomment-636933119, https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/847#issuecomment-659427016, https://github.com/w3c/webidl2.js/issues/392#issuecomment-524596523).

I'm not sure whether it's correct to do so in this way, but it seems that #847 is inactive so far and I'm not sure why we can't change the type according the drafts.

Please let me know if I did something wrong and/or close this PR if required.

fixes microsoft/TypeScript#37861
fixes https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/908


inputfiles/idlSources.json
```diff
@@ -438,6 +438,10 @@
         "url": "https://www.w3.org/TR/referrer-policy/",
         "title": "Referrer Policy"
     },
+    {
+        "url": "https://www.w3.org/TR/resize-observer/",
```
saschanaz, line 442: https://drafts.csswg.org/resize-observer/ is a better choice if the old version is faulty. Manually fixing IDL will require same fix every time anyone call `npm run fetch-idl`, which is not we want.

saschanaz: BTW I also filed an issue to remove `ResizeObservation` which is not supposed to be here: https://github.com/w3c/csswg-drafts/issues/5660

timocov: > drafts.csswg.org/resize-observer is a better choice if the old version is faulty. Manually fixing IDL will require same fix every time anyone call npm run fetch-idl, which is not we want.

Ah, sure! Didn't know about that, thanks!

> BTW I also filed an issue to remove ResizeObservation which is not supposed to be here: w3c/csswg-drafts#5660

Should I leave it here until it will be removed from IDL, right?

saschanaz: You can remove it by add a name in `removedTypes.json`. It makes sure we ignore the type whenever we update the IDL file 👍

baselines/dom.generated.d.ts
```diff
@@ -12642,6 +12646,51 @@ declare var Request: {
...
+declare var ResizeObserver: {
+    prototype: ResizeObserver;
+    new(callback: ResizeObserverCallback): ResizeObserver;
+};
+
+interface ResizeObserverEntry {
+    readonly borderBoxSize: ReadonlyArray<ResizeObserverSize>;
+    readonly contentBoxSize: ReadonlyArray<ResizeObserverSize>;
+    readonly contentRect: DOMRectReadOnly;
+    readonly devicePixelContentBoxSize: ReadonlyArray<ResizeObserverSize>;
```
saschanaz, line 12664: We also want to remove this specific property `devicePixelContentBoxSize` as it has no browser support. `removedTypes.json` may include 

```json
"ResizeObserverEntry": {
  "properties": {
    "property": {
      "devicePixelContentBoxSize": null
    }
  }
}`.

inputfiles/idlSources.json
timocov: Done in a85006d and 1c32701.

Thank you so much @saschanaz!

baselines/dom.generated.d.ts
timocov: Is it possible to mark it as optional? Or we don't want to have such properties until they come to spec and if someone wants to use them - they can use declaration merging for that?

saschanaz: > they can use declaration merging for that?

Yup. It's safer not to include it as unsupported properties are considered as highly experimental (which means it can be in a totally different form when added to browsers)

timocov: Agree. Done in 1601cca672e5e70a28355deecc13ac6cf5800d90

saschanaz (approved):
👍

Note that merging this may take some time as AFAIK it typically happens when new TS release is about to happen.

mdingena:
Thanks for being proactive and get things moving again. 👏  I'm looking forward to seeing this merged!

anandtiwary:
@sandersn can we merge this?

EricRabil:
@saschanaz there's been a few releases since your previous comment, is there any update on when this will be merged?

saschanaz:
By new release I mean a new minor version bump. I don't have any further information unfortunately.

a-tarasyuk (approved):

sandersn (approved):
This looks good. Once it's merged from master, it's ready to go in.

timocov:
Just merged master branch with resolving conflicts.

orta:
Thanks!

PR #949, opened 2020-11-04
Add composition events to GlobalEventHandlers

marijnh:
`GlobalEventHandlers` were missing `"compositionstart"`, `"compositionupdate"` and `"compositionend"`, which caused errors when trying to use a function type that relied on getting even object types from that type.

(I should note that I don't really know what I'm doing here, and just searched for the other event names and added similar declarations in what looked like the appropriate place.)

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -70,6 +70,18 @@
                 "attributeless-events": {
                     "event": [
                         {
+                            "name": "compositionstart",
+                            "type": "CompositionEvent"
+                        },
+                        {
+                            "name": "compositionupdate",
+                            "type": "CompositionEvent"
+                        },
+                        {
+                            "name": "compositionend",
+                            "type": "CompositionEvent"
+                        },
+                        {
                             "name": "focusin",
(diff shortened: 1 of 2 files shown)


inputfiles/addedTypes.json
```diff
@@ -93,6 +93,18 @@
                             "name": "change",
                             "type": "Event"
                         },
+                        {
+                            "name": "compositionstart",
```
saschanaz, line 97: Since there is no corresponding `oncompositionstart` etc., these should go to `attributeless-events` instead. Also, please run `npm i && npm run build && npm run baseline-accept` and commit the results.

justmeforyou (commented):
Thanks for

inputfiles/addedTypes.json
marijnh: Thanks for the pointer! I've updated the patch.

marijnh:
Is there anything else I can do to move this forward?

orta:
Looks very available to me: https://caniuse.com/mdn-api_element_compositionstart_event - thanks

PR #954, opened 2020-11-17
Add playoutDelayHint to RTCRtpReceiver.

### LINKED ISSUE #953: Add playoutDelayHint to RTPRtcReceiver.
**Search Terms**
playoutDelayHint, playoutDelay

**Suggestion**
Add playoutDelayHint to RTPRtcReceiver.

The API playoutDelayHint as a member to RTCRtpReceiver was introduced in https://henbos.github.io/webrtc-timing/#dom-rtcrtpreceiver-playoutdelayhint and implemented by Chrome. It was adopted by WebRTC extension as defined in https://w3c.github.io/webrtc-extensions/#dom-rtcrtpreceiver-playoutdelay with a different name, even though Chrome has not updated the implementation, and no other major browsers have implemented it.

It would be good to add playoutDelayHint to RTCRtpReceiver for now to enable Chrome. A TODO for the future is to deprecate it and rename it to playoutDelay.

**Use Cases**
In WebRTC based calling, one can use this to adjust the jitter buffer to flavor streaming or conferencing applications.

**Examples**
fippo/webrtc@6f77f10

minyuel:
Fix #953

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -12264,6 +12264,14 @@ declare var RTCPeerConnectionIceEvent: {
 
 /** This WebRTC API interface manages the reception and decoding of data for a MediaStreamTrack on an RTCPeerConnection. */
 interface RTCRtpReceiver {
+    /**
+     * Chrome exclusive hint to control jitter buffer delay.
+     * See https://henbos.github.io/webrtc-timing/#dom-rtcrtpreceiver-playoutdelayhint.
+     * This attribute later got renamed to playoutDelay and defined in WebRTC extensions.
+     * See https://w3c.github.io/webrtc-extensions/#dom-rtcrtpreceiver-playoutdelay.
+     * It is expected that playoutDelayHint be playoutDelay in the future.
+     */
+    playoutDelayHint: number | null;
     readonly rtcpTransport: RTCDtlsTransport | null;
     readonly track: MediaStreamTrack;
     readonly transport: RTCDtlsTransport | null;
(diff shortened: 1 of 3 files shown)


saschanaz:
Files in `inputfiles/idl` are not for manual modifications. You might want to add `https://w3c.github.io/webrtc-extensions/` to idlSources.json and run `npm run fetch -- "WebRTC Extensions" && npm run build && npm run baseline-accept`.

minyuel:
Hi Kagami,

Thanks for the review! WebRTC Extensions contains a lot more than playoutDelay, but that might be just good.

But I need to ask you for advice for a special situation. WebRTC Extensions contains playoutDelay but not playoutDelayHint, which is an old version of playoutDelay, and is already in Chrome.

saschanaz:
Oh, is Chrome the only implementation of playoutDelayHint? In that case it shouldn't be included here just yet, [as features need to be implemented on several engines](https://github.com/microsoft/TypeScript-DOM-lib-generator#additions) to make sure they are stable enough.

minyuel:
Hi,

Thanks for the reply!

This might have been well thought of, but to me, if a feature has to be implemented on multiple browsers to be added to TS DOM lib, it sounds a big blocker to the adoption of TS, as many features may be implemented at different paces on different browsers (interop is certainly considered a prerequisite). And on Chrome, some can first undergo an experiment before being settled.

How does TS handle these cases?

saschanaz:
So far the best way to support experimental features is through DefinitelyTyped, so that devs can explicitly know it's not available everywhere.

minyuel:
Ok, thanks! I think that may be what I am really after. Could you point me to how to do PR there?

saschanaz:
Sure. There are several w3c-prefixed packages e.g. [`w3c-generic-sensor`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/w3c-generic-sensor), so you can maybe add `w3c-webrtc-extensions` by `npx dts-gen --dt --name w3c-webrtc-extensions --template global` [per the documentation](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package).

minyuel:
Thanks a lot! For now we can close this PR.

PR #956, opened 2020-11-21
Correct type of Window.frameElement

djcsdy:
`frameElement` is `null` if the window is the top-level window, or if the window is embedded in a document that has a different origin.

See https://developer.mozilla.org/en-US/docs/Web/API/Window/frameElement

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -441,8 +441,7 @@
                             "deprecated": 1
                         },
                         "frameElement": {
-                            "name": "frameElement",
-                            "override-type": "Element | null"
+                            "nullable": 1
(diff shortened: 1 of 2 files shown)


inputfiles/overridingTypes.json
```diff
@@ -440,6 +440,10 @@
                             "override-type": "Event | undefined",
                             "deprecated": 1
                         },
+                        "frameElement": {
+                            "name": "frameElement",
+                            "override-type": "Element | null"
```
saschanaz, line 445:
```suggestion
                            "nullable": 1
```

orta:
Cool, yep, this looks accurate to me

Issue #826, opened 2020-02-21
Separete WebAssembly types from `lib.dom.d.ts` into their own file


ExE-Boss:
Currently, if you wish to use **WebAssembly**, you need to include `lib.dom.d.ts`, even in cases where that’s incorrect (e.g.: **NodeJS** or **Deno**).

The solution would be to move non‑web‑specific¹ **WebAssembly** types to a new `lib.wasm.d.ts` file.

---

¹ Web‑specific **WebAssembly** types are those defined in <https://webassembly.github.io/spec/web-api/index.html>.
Non‑web‑specific **WebAssembly** types are those defined in <https://webassembly.github.io/spec/js-api/index.html>.

Semigradsky:
+1

Currently I can't define `AbortController`/`AbortSignal`/`EventTarget`/`Event` for node@15.x because typings include `lib.dom.d.ts` (with its own implementation of `AbortController`/`AbortSignal`, etc.) (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48981)

ExE-Boss:
@Semigradsky That’s unrelated to this.

PR #959, opened 2020-12-02
feat: Emit WebAssembly types separately from DOM types

ExE-Boss:
Fixes <https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/826>

---

Note that I’d prefer if there was a way to filter based on the source **WebIDL** file, rather than the current `Flavor` const enum approach.

### DIFF
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -3113,6 +3113,11 @@
             "name": "WebAssembly",
             "methods": {
                 "method": {
+                    "compileStreaming": {
+                        "excluded-from": [
+                            "webassembly"
+                        ]
+                    },
                     "instantiate": {
                         "override-signatures": [
                             "instantiate(bytes: BufferSource, importObject?: Imports): Promise<WebAssemblyInstantiatedSource>",
@@ -3132,6 +3137,9 @@
                             {
                                 "type": "Imports"
                             }
+                        ],
+                        "excluded-from": [
+                            "webassembly"
                         ]
                     }
                 }
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -19,6 +19,7 @@ export const baseTypeConversionMap = new Map<string, string>([
     ["EventHandler", "EventHandler"]
 ]);
 
+// FIXME: This should be named `deepFilter`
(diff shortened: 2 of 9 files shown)


inputfiles/overridingTypes.json
```diff
@@ -2566,6 +2566,33 @@
...
+            },
+            "RuntimeError": {
+                "extends": "Error",
+                "constructor": {
+                    "override-signatures": [
+                        "new(message?: string): RuntimeError",
+                        "(message?: string): RuntimeError"
+                    ]
+                }
             }
```
ExE-Boss, line 2596: These types are actually defined using **ECMA‑262**’s <code><var>NativeError</var></code> object structure:
- <https://webassembly.github.io/spec/js-api/#error-objects>
- <https://tc39.es/ecma262/#sec-nativeerror-object-structure>

src/helpers.ts
```diff
@@ -64,7 +65,7 @@ export function merge<T>(target: T, src: T, shallow?: boolean): T {
                 const targetProp = target[k];
                 const srcProp = src[k];
                 if (Array.isArray(targetProp) && Array.isArray(srcProp)) {
-                    mergeNamedArrays(targetProp, srcProp);
+                    mergeNamedArrays(targetProp as any, srcProp as any);
```
saschanaz, line 68: Could be better to wait for https://github.com/microsoft/TypeScript/issues/40463 because the fix will probably be done earlier than the merge here 😁

saschanaz: Oh I'm wrong, the merges just happened!

inputfiles/overridingTypes.json
ExE-Boss: Extracted to: <https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/962>

ExE-Boss:
I’ve marked this as a draft, because this depends on <https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/962> and <https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/963>.

saltman424:
@sandersn Any chance this can get reviewed?

ExE-Boss:
This PR needs to be rewritten, as there were many changes to this since I last touched this.

PR #961, opened 2020-12-11
feat: Use `declare class` for namespaced types

ExE-Boss:
Unlike <https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/858>, which failed because `class`es declared in the global scope exist only on the global Declarative Environment Record, `class`es exported from **TypeScript**’s `namespace` construct exist as own properties of the `namespace` object, which matches the **WebIDL** behaviour.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -19031,81 +19031,46 @@ declare namespace CSS {
 }
 
 declare namespace WebAssembly {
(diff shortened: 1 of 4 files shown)


sandersn:
This doesn't address https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/858#issuecomment-644294982, does it? That should be the first step in this process.

PR #963, opened 2020-12-12
feat(WebAssembly): Make `WebAssembly.Global` generic

ExE-Boss:
This makes `WebAssembly.Global` generic to better match the runtime behaviour:
- <https://webassembly.github.io/spec/js-api/#globals>

---

**Q:** Should I change this to use something more like:
```ts
declare namespace WebAssembly {
	class Global<T extends ValueType = ValueType> {
		constructor(descriptor: GlobalDescriptor<T>, v?: ValueTypeMap[T]);
		value: ValueTypeMap[T];
		valueOf(): ValueTypeMap[T];
	}

	type ValueType = keyof ValueTypeMap;
	interface ValueTypeMap {
		i32: number;
		i64: bigint;
		f32: number;
		f64: number;
	}
}
```


baselines/dom.generated.d.ts
```diff
@@ -19146,14 +19146,21 @@ declare namespace WebAssembly {
         maximum?: number;
     }
     
+    interface ValueTypeMap {
+        f32: number;
+        f64: number;
+        i32: number;
+        i64: bigint;
+    }
+
```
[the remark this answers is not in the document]
ExE-Boss: This is the generated file, the changes would have to be done in `addedTypes.json`: <https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/f9eecea20b0141a17ce35e27783cfeb347f51cbc/inputfiles/addedTypes.json#L2372-L2400>

inputfiles/addedTypes.json
```diff
@@ -2368,6 +2368,35 @@
...
+                        "i32": {
+                            "name": "i32",
+                            "override-type": "number",
+                            "required": 1
+                        },
+                        "i64": {
+                            "name": "i64",
+                            "override-type": "bigint",
+                            "required": 1
+                        }
```
00ff0000red, line 2397:
```suggestion
                        }
                        "externref": {
                            "name": "externref",
                            "override-type": "unknown",
                            "required": 1
                        },
                        "anyfunc": {
                            "name": "anyfunc",
                            "override-type": "Function",
                            "required": 1
                        },
                        "funcref": {
                            "name": "funcref",
                            "override-type": "Function",
                            "required": 1
                        },
                        "v128": {
                            "name": "v128",
                            "override-type": "never",
                            "required": 1
                        }
```
¯\\_(ツ)\_/¯ would this work?
Also, I'm not sure if you want `funcref` in there, it's only stage 2.

ExE-Boss: Actually, `funcref` is already the canonical name: <https://github.com/WebAssembly/reference-types/pull/85>

---

That said, I’d rather hold off on this suggestion until the **Reference Types** proposal is merged into the **WebAssembly** spec, since with this PR, it will be easy for downstream code to extend `ValueTypeMap` by using interface merging:
```ts
declare global {
	namespace WebAssembly {
		interface ValueTypeMap {
			externref: unknown;
			funcref: Function;
		}
	}
}
```

00ff0000red: I see, `anyfunc` isn't even standard yet (I've been using experimental flags forever), I agree, just drop the references types, on the other hand, "v128" is much closer, do you want to keep it?

sandersn:
The types look like an improvement to me. @saschanaz would it be better to fix this upstream somewhere instead of overriding it here? If not I think this would be appropriate here.

saschanaz:
The IDL syntax does not support this kind of generics, so I think we need to override them. Looks good to me, feel free to merge 👍

sandersn (approved):

PR #964, opened 2020-12-15
Add beforeinput event to GlobalEventHandlersEventMap

marijnh:
`"beforeinput"`, as defined [here](https://w3c.github.io/uievents/#event-type-beforeinput) and implemented in Chrome for a long time and Firefox nightly recently, was missing from the event types.

This (trivially) conflicts with #949 since it adds to the same JSON structure in the same place. It'd be wonderful if someone could merge the both of them (since these don't seem controversial).


orta:
Looks like caniuse says it's in chrome and safari, and if it's in a FF nightly, then I'm good to have this in 👍🏻

PR #965, opened 2020-12-15
Allow URL in ServiceWorkerContainer#register

noelebrun:
Allow URL object to be used in `ServiceWorkerContainer#register`. Similar as #600 which was for Worker constructor.

Fixes microsoft/TypeScript#41980


orta:
I'm a little conflicted on this, because the docs and spec all assume string - but I think it's the same underlaying thing as this: https://github.com/microsoft/TypeScript/issues/15246#issuecomment-294948887 - so we'll take it 👍🏻

PR #966, opened 2020-12-22
Fix RTCPeerConnection override

### LINKED ISSUE #875: Parameter of  RTCPeerConnection.setLocalDescription should be optional
The parameter is optional, if omitted "the WebRTC runtime tries to automatically do the right thing.".
See [spec definition](https://w3c.github.io/webrtc-pc/#interface-definition) and the corresponding [mdn page](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setLocalDescription#Parameters).

It seems like [webidl2.js](https://github.com/w3c/webidl2.js) somehow ignores the `optional` in front of the parameter?

yousefamar:
The signature for RTCPeerConnection.setLocalDescription has been overriden, as of #386, to have a mandatory argument. This is no longer the case; see specs here:

https://w3c.github.io/webrtc-pc/#dom-peerconnection-setlocaldescription
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setLocalDescription#Parameters

This PR makes it optional, as it should be.

Resolves #875
Resolves https://github.com/microsoft/TypeScript/issues/41939
Resolves https://github.com/microsoft/TypeScript/pull/39142

Props to @jcc10 and @Simonwep for finding this bug before me, but trying to fix it by modifying the generated declaration files, instead of here (as per [these instructions](https://github.com/Microsoft/TypeScript/blob/master/CONTRIBUTING.md#modifying-generated-library-files)). I think this should solve the problem for good and finally stop TypeScript from complaining about the argument count?

That being said, I still think that [webidl2](https://github.com/w3c/webidl2.js) needs to be looked at, as @Simonwep said in #875, because in [this file](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/master/inputfiles/browser.webidl.preprocessed.json) after line 44565 there should be `"optional": 1`, (like #386, this is all from 2018) but I deliberatiely didn't touch that since I'm thinking it needs to be fixed upstream.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -12288,7 +12288,7 @@ interface RTCPeerConnection extends EventTarget {
     removeTrack(sender: RTCRtpSender): void;
     setConfiguration(configuration: RTCConfiguration): void;
     setIdentityProvider(provider: string, options?: RTCIdentityProviderOptions): void;
-    setLocalDescription(description: RTCSessionDescriptionInit): Promise<void>;
+    setLocalDescription(description?: RTCSessionDescriptionInit): Promise<void>;
     setRemoteDescription(description: RTCSessionDescriptionInit): Promise<void>;
     addEventListener<K extends keyof RTCPeerConnectionEventMap>(type: K, listener: (this: RTCPeerConnection, ev: RTCPeerConnectionEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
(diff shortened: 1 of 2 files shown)


orta:
Thanks, @yousefamar  - I agree that this change matches the docs and we should take it 👍🏻 

I guess we'll see what happens with the webidl change too

patelnav:
Do we know which version of Typescript will get this fix? 

https://github.com/microsoft/TypeScript/blob/408c804103d5b4f3ce97722190eab9d428740d80/lib/lib.dom.d.ts#L12310

PR #968, opened 2021-01-03
fix: Mark `prototype` property of constructors as `readonly`

ExE-Boss:
This more closely matches runtime behaviour.

---

It also causes code that would try to re‑assign the `prototype` property of **WebIDL** constructors to have the error caught at compile time, rather than at run time.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1997,7 +1997,7 @@ interface AbortController {
 }
 
 declare var AbortController: {
-    prototype: AbortController;
+    readonly prototype: AbortController;
     new(): AbortController;
 };
 
@@ -2019,7 +2019,7 @@ interface AbortSignal extends EventTarget {
 }
 
 declare var AbortSignal: {
-    prototype: AbortSignal;
+    readonly prototype: AbortSignal;
     new(): AbortSignal;
 };
 
@@ -2047,7 +2047,7 @@ interface AbstractRange {
 }
 
 declare var AbstractRange: {
-    prototype: AbstractRange;
+    readonly prototype: AbstractRange;
     new(): AbstractRange;
 };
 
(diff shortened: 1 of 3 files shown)


orta:
We think this is probably a bit too big of a change for its value, I agree that it better reflects reality (example below) but it's hard to gauge a potential breaks from something like this 

```
AbortController.prototype = () => {}
> () => {}

new AbortController()
> AbortController {signal: AbortSignal}

const a = new AbortController()
undefined

a
>  AbortController { 
  signal: AbortSignal {aborted: false, onabort: null} 
  __proto__: AbortController
}
```

saschanaz:
> We think this is probably a bit too big of a change for its value, I agree that it better reflects reality (example below) but it's hard to gauge a potential breaks from something like this.

I'm not sure I understand the reasoning here, your example looks like it will benefit from `readonly`? The reason it's silently ignoring reassignation of `.prototype` is because it's not in strict mode.

```
"use strict"; AbortController.prototype = () => {}
> Uncaught TypeError: "prototype" is read-only

saschanaz:
(But I guess some polyfill can legally assign things to `.prototype`)

petamoriken:
Can we move this PR forward?

saschanaz:
We nowadays have better testing and hopefully the CI in microsoft/TypeScript side repo can catch potential breakage @orta was talking about on DefinitelyTyped. PR welcome (assuming one would also try the CI on TS repo too)

jakebailey:
Before doing this, you probably need to consider #58296 as it may become necessary to remove the `readonly` modifier on some prototypes.

saschanaz:
Not sure what those some would be. Would that be because of compatibility issue with packages?

jakebailey:
No, it's because the PR would actually enforce that you can't assign an object with a readonly property to one that has the same property but mutable, and that relationship also applies to interfaces extending other interfaces. So, you can't inherit a mutable prototype and then mark it as readonly or vice versa.

But, maybe it would work out to mark _all_ prototypes as readonly, it just doesn't seem like that's the way things are going based on the discussions about the PR.

Issue #899, opened 2020-08-16
Incorrect "this" value in event handlers


tyler1205:
In:
```typescript
let el = document.createElement('p')
el.ontransitionend = function(){let r = this}
```

The this value of r is actually el or typeof el, but typescript tells me that r is "GlobalEventHandlers". This is very annoying because even if I try to replace it with:
```typescript
el.ontransitionend = function(this: HTMLParagraphElement, ev: TransitionEvent){let r = this}
```

TypeScript still incorrectly throws an error saying that the "this" types are incompatible...

If TS cannot correctly assume/infer the this type, it should at least default to the "any" type or something so that I can specify my own this type without getting errors. I'm not sure if this happens with all events or just those that are "GlobalEventHandlers" related, but I did notice that in the following variations of my example, the this types are more or less correctly assumed/inferred:

```typescript
ontransitionend = function(){let r = this} // r is "Window & typeof globalThis"
window.ontransitionend = function(){let r = this} // r is "Window & typeof globalThis"
```

PR #969, opened 2021-01-04
Event type for readystatechange is not specific enough

jonhue:
Fixes https://github.com/microsoft/TypeScript/issues/41775

I was a little bit unsure whether its best to use `ProgressEvent<Document>` vs making `Event` generic (`Event<T extends EventTarget = EventTarget>`) and then using `Event<Document>`.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4446,7 +4446,7 @@ interface DocumentEventMap extends GlobalEventHandlersEventMap, DocumentAndEleme
     "fullscreenerror": Event;
     "pointerlockchange": Event;
     "pointerlockerror": Event;
-    "readystatechange": Event;
+    "readystatechange": ProgressEvent<Document>;
     "visibilitychange": Event;
 }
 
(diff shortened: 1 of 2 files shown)


HolgerJeromin:
This is a bit of a cherrypicking of all the unspecific typed events (even if you only look at `document`).
even `document.onreadystatechange` is not changed... 8-/

ref #899 and https://github.com/Microsoft/TypeScript/issues/299

jonhue:
@HolgerJeromin Thank you for these pointers!

I actually would be in favor of turning the `Event`type into something like `Event<T extends EventTarget = EventTarget>`.
That shouldn't really break existing code except altering some error messages.

But this would be a larger change, so without someone from the TypeScript team chiming in and telling us they are looking for something like that, I'm somewhat hesitant to implement it. At least as part of this pr.

orta:
Looking at all the past issues with trying to make these types more explicit, we're very wary of changes here which adds generics to types which are re-used all of the place. I know it's not a perfect representation of the underlaying DOM as it is, but we're making perf vs memory vs accuracy trade-offs and I think we'll probably not take this.

jonhue:
@orta makes sense :) should we then close the original issue, so that we don't spend more time on this in the future?

PR #971, opened 2021-01-07
Make ImportMeta's url match the version in @types/node

orta:
Correctly implements the hotfix I added to https://github.com/microsoft/TypeScript/pull/42230


sandersn:
Looks good, especially since this version fixes the webworker change, which microsoft/TypeScript#42230 does not.

PR #972, opened 2021-01-13
Use this: this in event handlers in mixin types and this: Window & typeof globalThis

weswigham:
…in global event handlers.

Should fix https://github.com/microsoft/TypeScript/issues/39057

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2056,10 +2056,10 @@ interface AbstractWorkerEventMap {
 }
 
 interface AbstractWorker {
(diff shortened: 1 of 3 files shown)


orta (approved):
This reads good to me. I've rebased some changes which will test this against the compiler's test suite, so Danger might have something to say - we'll see.

sandersn (approved):
I think this is a good idea, but can you first create a PR on the Typescript repo to test performance?

saschanaz:
Is there a PR there or can I go ahead and open one?

sandersn:
No, I'm pretty sure there's no PR.

PR #975, opened 2021-01-31
fix: `window.opener` should be not be typed as any

Sapfir0:
Fixes [microsoft/TypeScript#42433](https://github.com/microsoft/TypeScript/issues/42433)

I have not found a suitable IDL for window.opener, but you can find [here ](https://html.spec.whatwg.org/multipage/browsers.html#dom-opener) that window.opener can be specified as WindowProxy or null.


orta:
Agree based on reading the spec, this looks good to me 👍🏻

PR #977, opened 2021-02-13
fix: add loading attributes

1000ch:
- Add [`loading` attribuets for `<iframe>` element](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element).
- Add a description of [`loading` attribuets for `` element](https://html.spec.whatwg.org/#the-img-element).
- Use [`'eager' | 'lazy'` instead of `string` as a type for `loading` attributes](https://html.spec.whatwg.org/#lazy-loading-attributes)

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -7084,6 +7084,10 @@ interface HTMLIFrameElement extends HTMLElement {
      */
     height: string;
     /**
+     * Sets or retrieves the policy for loading iframe elements that are outside the viewport.
+     */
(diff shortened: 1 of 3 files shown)


1000ch:
@sandersn could you please take a look? 🙂

orta:
I think this is probably too early as it looks to only be implemented in Chrome: https://caniuse.com/mdn-html_elements_iframe_loading

1000ch:
Let me close this PR to separate PR.

PR #984, opened 2021-02-25
add performanceservertiming

keithamus:
In the hopes of getting a discussion going around [PerformanceServerTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceServerTiming) this implements the basic types as specified in [the WD spec](https://w3c.github.io/server-timing/#the-performanceservertiming-interface).

From the contributor guidelines:

 - [x] Does the new objects or fields show up in mdn/browser-compat-data? If not, it's likely too soon. (See [browser-compat-data](https://github.com/mdn/browser-compat-data/blob/master/api/PerformanceServerTiming.json))
- [ ] Is the IDL source from WHATWG? (No, W3C)
    - [x] Are the additions available in at least two of [Firefox](https://searchfox.org/mozilla-central/search?q=&path=), [Safari](https://webkit-search.igalia.com/webkit/search?q=&path=) and Chromium?
- [x] Is the IDL source from W3C?
    - [ ] What stage of the [W3C process](https://en.wikipedia.org/wiki/World_Wide_Web_Consortium#Specification_maturation) is the proposal for these changes: We aim for Proposed recommendation, but can accept Candidate recommendation for stable looking proposals. (Working Draft)
    - [x] If it's at Working draft the additions available in all three of Firefox, Safari and Chromium (Not Safari yet, but Chrome, Edge & Firefox)
- [x] Could any types added at the global scope have naming conflicts? (Doubtful unless someone is polyfilling these?)
- [x] Are the new features going to be used by a lot of people? (We'd like to use this on GitHub :shrug:)

From the [browser compat table on mdn](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceServerTiming#browser_compatibility) this API is available in Firefox 61, Chrome 65, Edge 79

[browser compat table](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceServerTiming#browser_compatibility)

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -11687,6 +11687,7 @@ interface PerformanceResourceTiming extends PerformanceEntry {
     readonly responseEnd: number;
     readonly responseStart: number;
     readonly secureConnectionStart: number;
+    readonly serverTiming: ReadonlyArray<PerformanceServerTiming>;
     readonly transferSize: number;
     readonly workerStart: number;
     toJSON(): any;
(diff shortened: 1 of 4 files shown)


orta (approved):
This looks good to me, thanks for the well written PR description - we should consider making a PR template with that?

sandersn (approved):

PR #990, opened 2021-03-08
Use eslint+prettier

saschanaz:
This also removes src/preprocess.ts because nothing uses it anymore.


orta:
Cool, sorry open PRs - me and Nathan are happy to deal with conflicts from the fallout

PR #996, opened 2021-03-28
feat(emitter): support differing accessor types

saschanaz:
Thanks to https://github.com/microsoft/TypeScript/pull/42425

This turns out to be less interesting than I expected because of the setter type limitation where the getter type must be assignable to the setter type. Thus the following can't be supported:

```webidl
[Exposed=Window]
interface CSSStyleRule : CSSRule {
  attribute CSSOMString selectorText;
  [SameObject, PutForwards=cssText] readonly attribute CSSStyleDeclaration style;
};
```

```ts
interface CSSStyleRule extends CSSRule {
    selectorText: string;
    get style(): CSSStyleDeclaration;
    set style(cssText: string); // error!
}
```

```ts
document.body.style = "display: none" // thus still an error 🤔
```

Probably worth to file an issue, or does TS team think this is not a very interesting case?

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4538,7 +4538,8 @@ interface Document extends Node, DocumentAndElementEventHandlers, DocumentOrShad
     /**
      * Contains information about the current URL.
      */
-    location: Location;
+    get location(): Location;
+    set location(href: string | Location);
     onfullscreenchange: ((this: Document, ev: Event) => any) | null;
     onfullscreenerror: ((this: Document, ev: Event) => any) | null;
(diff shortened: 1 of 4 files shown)


orta (approved):
I like this, this is definitely the case I hit the most often with the DOM and it shows of 4.3's feature well

sandersn (approved):

sandersn:
I think you should file an issue, because during the design of this feature, we decided on a conservative loosening of the rules. It would be good to have an issue to gather evidence for further loosening.

@orta I'm going to merge this, but it's a little late for putting DOM changes into 4.3. Do you agree?

orta:
Grudgingly agree

PR #997, opened 2021-03-28
feat: add serviceworker and sharedworker libs

### LINKED ISSUE #505: Create ServiceWorker lib
Service workers basically extends from general worker but has their own types e.g. Client, ExtendableEvent, FetchEvent, etc. Will you accept a PR for this?

### LINKED ISSUE #675: Unable to define a `.js` file as ServiceWorkerGlobalScope
When a `.js` file is a service-worker, there's no way to define its scope/`self`/`this` as `ServiceWorkerGlobalScope`. This means service-worker only global APIs like `self.skipWaiting()` are undefined and fail Typescript's `checkjs` run.

**Minimal test-case**

https://github.com/wilsonpage/vscode-scope-test-case

**Twitter thread w/ @DanielRosenwasser**

https://twitter.com/drosenwasser/status/1113905694028132352

saschanaz:
Closes #675, closes #505


HolgerJeromin:
CC @jakearchibald and @MattiasBuelens

src/index.ts
```diff
@@ -280,14 +279,22 @@ function emitDom() {
...
-    flavor: Flavor.Window,
-    global: "Window",
+    global: ["Window"],
     outputFolder,
   });
   emitFlavor(webidl, new Set(knownTypes.Worker), {
     name: "webworker",
-    flavor: Flavor.Worker,
-    global: "Worker",
+    global: ["Worker", "DedicatedWorker", "SharedWorker", "ServiceWorker"],
```
MattiasBuelens, line 287: Should we have a new `dedicatedworker` flavor that only contains `["DedicatedWorker", "Worker"]`? It would be nice if you're writing code for a dedicated worker to have the type checker disallow using service worker-only or shared worker-only APIs. But I can also understand that it may be difficult to teach users the difference between the "webworker" and "dedicatedworker" flavors.

saschanaz: Sounds good to me and it was my initial plan. In the ideal world we could ultimately deprecate `webworker` in that case. I wonder what the maintainers think.

saschanaz:
Replaced by #1034

PR #1005, opened 2021-04-27
feat: Add `SubmitEvent`

cawa-93:
I try to add `SubmitEvent` interface. [MDN](https://developer.mozilla.org/docs/Web/API/SubmitEvent).

Resolve https://github.com/microsoft/TypeScript/issues/40811


PaperStrike:
#1005 has added `SubmitEvent`, while not using it in "submit" events, which seems like [a bug](https://github.com/whatwg/html/issues/6948) pulled from the standard. 🤔

PR #1011, opened 2021-05-14
Merge comment and depreciation notice

Scharkee:
Fixes https://github.com/microsoft/TypeScript/issues/44039 

If there is an entity comment and a depreciation comment, they will be merged into one.

[This is a diff ](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1011/files#diff-d57c51d024219b9f37e6b42df127db25f700acbaade23ce1edcecda2853b995c)for one of the generated files, `dom.generated.d.ts`, before and after this change.

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -826,14 +826,21 @@ export function emitWebIdl(
     entity: { comment?: string; deprecated?: boolean },
     print: (s: string) => void
   ) {
-    if (entity.comment) {
-      if (entity.comment.startsWith("/*")) {
-        entity.comment.split("\n").forEach(print);
-      } else {
-        print(`/** ${entity.comment} */`);
-      }
+    let comment = entity.comment;
+    if (
+      entity.comment &&
+      entity.deprecated &&
+      !entity.comment?.includes("@deprecated") &&
+      entity.comment.startsWith("/*")
+    ) {
+      comment = entity.comment.replace("*/", "* @deprecated\n */");
     }
-    if (entity.deprecated && !entity.comment?.includes("@deprecated")) {
+
+    if (comment) {
+      comment.split("\n").forEach(print);
+    }
+
+    if (entity.deprecated && !comment?.includes("@deprecated")) {
       print(`/** @deprecated */`);
     }
   }
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -837,13 +837,11 @@ export function emitWebIdl(
         }
 
         split.forEach(print);
+      } else if (entity.deprecated && !entity.comment.includes("@deprecated")) {
+        // Convert into multi-line comment with deprecation notice.
+        print(`/**\n * ${entity.comment}\n * @deprecated\n */`);
       } else {
-        if (entity.deprecated && !entity.comment.includes("@deprecated")) {
-          // Convert into multi-line comment with deprecation notice.
-          print(`/**\n * ${entity.comment}\n * @deprecated\n */`);
-        } else {
-          print(`/** ${entity.comment} */`);
-        }
+        print(`/** ${entity.comment} */`);
       }
     } else if (entity.deprecated) {
       print(`/** @deprecated */`);
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -826,21 +826,22 @@ export function emitWebIdl(
     entity: { comment?: string; deprecated?: boolean },
     print: (s: string) => void
   ) {
-    let comment = entity.comment;
     if (
-      entity.comment &&
       entity.deprecated &&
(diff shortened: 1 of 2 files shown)


sandersn:
This fix is important because recent versions of Typescript use only the jsdoc directly above a function for generating the documentation. For example: https://www.typescriptlang.org/play?#code/PQKhAIFkEMGsFMDO4Au8AeLwCd7QDb4Ce4ARgJYDmAdOCMAFCgQACAJvAA64DG0abOowBmAVwB2PFOQD24slQAUiAFzhEKbOXGUAlGo1ad4AN4NwOeClHZ5iahUqLdDAL4NHigOQALcl90gA

src/build/emitter.ts
```diff
@@ -826,6 +826,14 @@ export function emitWebIdl(
     entity: { comment?: string; deprecated?: boolean },
     print: (s: string) => void
   ) {
+    if (
+      entity.deprecated &&
+      entity.comment?.endsWith(`*/`) &&
+      !entity.comment?.includes("@deprecated")
```
sandersn, line 832: when is this true? I thought that all the `@deprecated` tags would be inserted on line 845.

Edit: I see that the previous code also includes this check, but I think it's worth taking it out and where extra `@deprecated` are coming from, if any.

src/build/emitter.ts
```diff
@@ -826,6 +826,14 @@ export function emitWebIdl(
     entity: { comment?: string; deprecated?: boolean },
     print: (s: string) => void
   ) {
+    if (
+      entity.deprecated &&
+      entity.comment?.endsWith(`*/`) &&
```
sandersn, line 831: why not use `entity.comment?.startsWith("/*")` like line 838?

src/build/emitter.ts
```diff
@@ -826,6 +826,14 @@ export function emitWebIdl(
...
+      !entity.comment?.includes("@deprecated")
+    ) {
+      entity.comment = entity.comment.replace("*/", "* @deprecated\n */");
+    }
+
     if (entity.comment) {
```
sandersn, line 837: Instead of a new, side-effecting section, I'd rather see the existing code refactored into exclusive comment/no-comment sections:

```ts
if (entity.comment) {
  // previous code, plus code to insert `@deprecated`
}
else if (entity.deprecated) {
  // previous code
  print("/** @deprecated")
}

**Review by sandersn — CHANGES_REQUESTED** [review]
This is a great fix. I'd like to see it better integrated with the existing code, though.

File: src/build/emitter.ts (hunk 1 above)
Comment (Scharkee) [reply src/build/emitter.ts:832]: Some entity comments have deprecated tags already included in the comment block. Some examples would be ScriptProcessorNode and PerformanceTiming. This condition is true if the entitity is deprecated, and has a multi-line comment, but does not yet contain the @deprecated tag. I will clean it up to be more readable.

File: src/build/emitter.ts (hunk 2 above)
Comment (Scharkee) [reply src/build/emitter.ts:831]: This seemed to be more apt given the @deprecated tag gets inserted at the closing tag. In principle, given that a comment is valid and has an opening tag, it should have a closing tag a well, but there is no harm in checking.

File: src/build/emitter.ts (hunk 3 above)
Comment (Scharkee) [reply src/build/emitter.ts:829]: Done in c5b68a87d5bebdb19061dc78f11dff05dd95e573. A few edge cases where conversion from single to multi-line comment was required are now also handled.

**Review by Scharkee — COMMENTED** [review]
Changes implemented.

File: src/build/emitter.ts [hunk 4]
```diff
@@ -828,12 +828,24 @@ export function emitWebIdl(
...
+        // Merge in the deprecated notice, if required.
+        if (entity.deprecated && !entity.comment.includes("@deprecated")) {
+          split[split.length - 1] = " * @deprecated";
+          split.push(" */");
+        }
+
+        split.forEach(print);
       } else {
-        print(`/** ${entity.comment} */`);
+        if (entity.deprecated && !entity.comment.includes("@deprecated")) {
```
sandersn, line 841: formatting: you can merge `else { if (...` to `else if (...` and drop an indent level

sandersn (changes requested):
All ready to go except a minor formatting request (unless it's required by prettier, I can't remember if it's running on this repo)

orta:
This is good work 👍🏻

src/build/emitter.ts
Scharkee: Done

sandersn (approved):

sandersn:
OK, let's see if the `@types/web` publisher works!

sandersn:
Yep, it's now available as 0.0.6!

PR #1013, opened 2021-05-26
feat: Add `CSSAnimation` and `CSSTransition` interfaces

### LINKED ISSUE #1012: Add interfaces from CSS Animations Level 2 and CSS Transitions Level 2
These specs define the [`CSSAnimation`](https://drafts.csswg.org/css-animations-2/#cssanimation) and [`CSSTransition`](https://drafts.csswg.org/css-transitions-2/#csstransition) interfaces respectively which are implemented in all three major browser engines.

I believe we simply need to add the following URLs to [`idlSources.json`](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/master/inputfiles/idlSources.json):

* https://drafts.csswg.org/css-animations-2/
* https://drafts.csswg.org/css-transitions-2/

birtles:
Fixes #1012.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2620,6 +2620,19 @@ declare var CDATASection: {
     new(): CDATASection;
 };
 
+interface CSSAnimation extends Animation {
+    readonly animationName: string;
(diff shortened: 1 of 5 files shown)


birtles:
As best I can tell, those CI test failures are bogus.

PR #1019, opened 2021-06-05
Deploy @webref/elements and partially @webref/idl

saschanaz:
This adds SVG Animations spec and removes quite a lot of lines from `inputfiles/`.


saschanaz:
Replaced by #1034

PR #1022, opened 2021-06-19
fix(44560): Make `PaymentRequest.show` method param optional

thunder-coding:
This PR makes `detailsPromise` param of `PaymentRequest.show` optional

Fixes microsoft/TypeScript#44560


HolgerJeromin:
Thanks for helping.

mayeedwin:
Getting the same error. Looks at [TypeScript-DOM-lib-generator](https://github.com/mayeedwin/TypeScript-DOM-lib-generator) and seems like the [bug is fixed](detailsPromise). Maybe not published yet. **But for now**, you can manually edit on your dev environment to keep going as you await for the release.

cc @thunder-coding

PR #1025, opened 2021-06-22
Add the ability to upload to @types/x

orta:
Supports mapping the generated files from the generator to different `@types` packages via a deploy process. I ended up more or less using the exact same infra as [tsconfig/bases](https://github.com/tsconfig/bases) but ported from Deno to Node. 

Roughly, as you'll see it during code review:
 - Adds a yml for deploys on master merges. Runs through building and tests, then creates packages and deploys.
 - `deploy/createTypesPackages.mjs` is responsible for generating all possible packages as complete artefacts
 - `deploy/deployChangedPackages.mjs` is responsible for comparing the built packages against the versions on npm, if they differ - then deploy

TODO: 

 - [x] I probably want the ability to have a custom README for `@types/web` (at least) but maybe I just vendor each README in the repo so they can can hand crafted to better describe their domain.

 - [x] I'm still a little undecided on what should be done with the `.iterator` files. E.g. should they be included with the corresponding non-iterator. I lean on _yes_, but I need to verify with others.

Iterators are included by default in pretty much every case but ES5:
 
Screen Shot 2021-06-22 at 2 19 35 PM


deploy/deployChangedPackages.mjs
```diff
@@ -0,0 +1,84 @@
...
+// @ts-ignore
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const go = async () => {
+  const uploaded = [];
+
+  // Loop through generated packages, deploying versions for anything which has different
+  // .d.ts files from the version available on npm.
+  const generatedDir = join(__dirname, "generated");
```
saschanaz, line 23: Passing URL value to fs methods can skip the path conversions.

deploy/createTypesPackages.mjs
```diff
@@ -0,0 +1,122 @@
...
+  for (const pkg of packages) {
+    const folderName = pkg.name.replace("@", "").replace("/", "-");
+    const packagePath = join(generatedDir, folderName);
+
+    if (fs.existsSync(packagePath)) fs.rmSync(packagePath, { recursive: true });
+    fs.mkdirSync(packagePath, { recursive: true });
+
+    // Migrate in the template files
+    for (const templateFile of fs.readdirSync(templateDir)) {
+      if (templateFile.startsWith(".")) continue;
```
sandersn, line 47: Is this line used? Is it for .DS_Store?

deploy/createTypesPackages.mjs
```diff
@@ -0,0 +1,122 @@
...
+    const semverMarkers = npmPackage["dist-tags"].latest.split(".");
+    const bumpedVersion = `${semverMarkers[0]}.${semverMarkers[1]}.${
+      Number(semverMarkers[2]) + 1
+    }`;
+
+    if (isBumpedVersionHigher(version, bumpedVersion)) {
+      version = bumpedVersion;
+    }
+  } catch (error) {
+    // NOOP, this is for the first deploy, which will set it to 0.0.1
```
sandersn, line 96: I'd rather have us seed each package with 0.0.1 manually and raise the error here, since errors are probably more frequent than new packages.

Also I'm not sure which statement is expected to error on first deploy. Just `const npmResponse = ...`, right?

Also, if `fetch` fails AND the file has new changes, then by default we'll try to publish 0.0.1, which will fail. I think it's better to stop trying to publish entirely instead.

deploy/template/README.md
```diff
@@ -0,0 +1,12 @@
+# Installation
+> `npm install --save {{name}}`
```
sandersn, line 2: I didn't see where these get replaced with real values.

sandersn (changes requested):
Mostly questions about error handling.

I vote to have custom READMEs, since I don't think we'll have that many different packages (at first, especially).

saschanaz:
>     * is responsible for comparing the built packages against the versions on npm, if they differ - then deploy

Can we do this locally by comparing between git tags in `baselines/`?

orta:
> Can we do this locally by comparing between git tags in baselines/?

Probably not if we want to be able to deploy just `@types/web-sharedworker` and not `@types/web` when only shared worker has changed, tags would only work if all version numbers were kept in sync

saschanaz:
> Probably not if we want to be able to deploy just `@types/web-sharedworker` and not `@types/web` when only shared worker has changed, tags would only work if all version numbers were kept in sync

Webref have tags for each package https://github.com/w3c/webref/releases and I wonder it can also be done here. This way we can find the latest tag and see if the corresponding file changed from that tag.

orta:
Nice, scoping the tags via a name seems pretty reasonable to me. 

Should be pretty easy to get access to the raw files for a release via the GH API 👍🏻

saschanaz:
FYI this is what I'm doing on types-web: https://github.com/saschanaz/types-web/blob/7b2675fc37d568d2d99e288eb63125f8c55581af/src/changelog.ts#L5-L13

`git describe --match` seemingly also supports filtering by tag name👍

deploy/deployChangedPackages.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+    // Look through each .d.ts file included in a package to
+    // determine if anything has changed
+    let upload = false;
+    for (const file of dtsFiles) {
+      const generatedDTSPath = join(generatedDir, dirName, file);
+      const generatedDTSContent = fs.readFileSync(generatedDTSPath, "utf8");
+      try {
+        const unpkgURL = `https://unpkg.com/${pkgJSON.name}/${file}`;
+        const npmDTSReq = await fetch(unpkgURL);
```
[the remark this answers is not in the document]
orta: Heh, I've only once had an issue with unpkg on the playground for the last 2 years, and that was somewhat my fault (I passed `//` instead of `/` and it stopped resolving) - either way, @saschanaz's idea for tags means we don't need this dependency.

deploy/deployChangedPackages.mjs
```diff
@@ -0,0 +1,84 @@
...
+        upload = upload || npmDTSText !== generatedDTSContent;
+      } catch (error) {
+        // Not here, definitely needs to be uploaded
+        upload = true;
+      }
+    }
+
+    // Publish via npm
+    if (upload) {
+      if (process.env.NODE_AUTH_TOKEN) {
```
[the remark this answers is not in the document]
orta: We do log at the end that it was a dry run, so I think this is probably fine, but I can expand on the logs for the dry run.

deploy/deployChangedPackages.mjs
```diff
@@ -0,0 +1,84 @@
...
+const go = async () => {
+  const uploaded = [];
+
+  // Loop through generated packages, deploying versions for anything which has different
+  // .d.ts files from the version available on npm.
+  const generatedDir = join(__dirname, "generated");
+  for (const dirName of fs.readdirSync(generatedDir)) {
+    const localPackageJSONPath = join(generatedDir, dirName, "package.json");
+    const newTSConfig = fs.readFileSync(localPackageJSONPath, "utf-8");
+    const pkgJSON = JSON.parse(newTSConfig);
```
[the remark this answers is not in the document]
orta: its not even tsconfig too ;)

deploy/createTypesPackages.mjs
orta: Yep, on macOS they appear all over the place and I just always do this

orta:
Using git to make the comparison won't realistically work without a bunch of git faff on every push, which I don't think is worth it  - we'd be comparing the files which are meant to be `.gitignored`, so we'd have to have CI ship a version with the gitignore changed, then undo it. 

I think we're fine to use unpkg, but I'll make the release tagging (and once we're up-to-date with types-web, I'll make the CHANGELOG feed into that in GitHub releases)

saschanaz:
Could you elaborate why it would do anything with every push?

orta:
I've been using the files in the npm package as the source of truth and both the npm packages aren't included in the git repo nor are the files which they use in `generated`.

I could maybe go backwards through the generator metadata object get back to the originals, and use the baseline test files instead - but this way means comparing the exact same file in the package to the exact same file package.

A safe alternative which can be relied on it in git is:

 - Force add the generated packages into git
 - Commit that with a `[no ci]`, use this for the tag reference
 - Make a commit undoing the generated packages

But that's more effort than just downloading and unzipping the npm tarball and looking at the raw file contents instead of rely on unpkg to do that for us

orta:
I've asked in the TypeScript chat room about including dom.iterator.d.ts in types/web - I'm on the site of "yes", we'll see -this can be un-blocked before then. Given it's all deploy stuff, I can only know it works when it hits reality.

saschanaz:
> I've been using the files in the npm package as the source of truth and both the npm packages aren't included in the git repo nor are the files which they use in `generated`.

Are you going to use `git diff`, because otherwise the files don't have to be in the git repo? Files can be loaded via `fs.readFile`/`git show` and then be `===`ed instead.

saschanaz:
Anyway, I guess it can be a separate work if the current solution works.

deploy/deployChangedPackages.mjs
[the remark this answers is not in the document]
sandersn: :+1: more logs are good

sandersn (approved):

orta:
I think we need to keep the 'if no package exists, then deploy' logic - none of us are logged in as the types account on npm to initially add the dependency

RandinAk (commented):
He****

PR #1028, opened 2021-06-25
Add support for User-Agent Client Hints

rowan-m:
Provides `navigator.userAgentData` for User-Agent Client Hints values as per https://wicg.github.io/ua-client-hints/

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -859,6 +859,11 @@ interface NavigationPreloadState {
     headerValue?: string;
 }
 
+interface NavigatorUABrandVersion {
+    brand?: string;
+    version?: string;
+}
+
 interface NotificationAction {
     action: string;
     icon?: string;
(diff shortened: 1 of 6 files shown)


rowan-m:
From the heuristics for inclusion, I'm afraid this is currently only available in Chromium-based browsers. However with the hope of generally moving away from use of `navigator.userAgent` it would be helpful to have the new API available here. So I'm hoping for addition based on the criteria: _"new features going to be used by a lot of people"._

rowan-m:
This was previously raised here: https://github.com/microsoft/TypeScript/issues/43604 back in April, but with this now being available in stable it felt like a good opportunity to revisit.

orta:
Thanks - I'm afraid we're probably going to be waiting on this until we see some other browsers implement it

rowan-m:
Sounds fair to me. I'm fine with this either waiting or just closing and we can try again when there's additional support. Repo instructions were very clear, so thanks for making it so simple to contribute.

SNDST00M:
As noted in the April issue, https://wicg.github.io/ua-client-hints/#interface

rowan-m:
Cleaning up as this is a fair few years old now and it probably needs a fresh pass if the feature request is still valid.

PR #1035, opened 2021-07-01
Gives @saschanaz the ability to merge PRs which affect src/baselines/inputfiles

orta:
By using https://github.com/OSS-Docs-Tools/code-owner-self-merge we get to give enough rights to @saschanaz while still conforming to the Microsoft security guidelines


sandersn (approved):
We should hold off on merging (or @saschanaz should hold off on merging PRs) until 

1. the backport from types_web is merged and published as 0.0.2
2. the merge-to-publish action is in.

That way we never group a lot of changes into a single version of `@types/web` again.

PR #1034, opened 2021-06-29
Merge types_web back into TypeScript DOM lib generator

orta:
> This comparison is big! We’re only showing the most recent 250 commits

This brings in all the changes, and merges the changes to the TS repo at the same time. The output .d.ts files post-merge basically match the version in the compiler with the exceptions of the PRs I submitted to types_web which I haven't got into a mergeable state yet.

Me and @sandersn have gone through the types-web d.ts files with a pretty fine brush in prep for migrating to use that version for 4.4 ( https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1023 ) so I'm pretty happy with the output. 

The merge was pretty rough, so I tended to prefer the types-web version for source code, then try make sure both was applied to anything in the JSON.

PR #1036, opened 2021-07-02
fix(1017) - SVGPathElement extends geometry

### LINKED ISSUE #1017: SVGPathElement should extend SVGGeometryElement and not SVGGraphicsElement
see also: https://github.com/microsoft/TypeScript/issues/44336

NotWoods:
Fixes #1017

Added the paths page found on MDN, which just includes this one interface. That change removes the deprecated property (it doesn't show up anywhere, removed in Chrome 48) and removes the duplicate methods (they are present on SVGGeometryElement).

For some reason `"path"` was removed from the `SVGElementTagNameMap`. I'm not sure how to fix that and any support would be appreciated!

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2874,6 +2874,7 @@ interface CSSStyleDeclaration {
     cssFloat: string;
     cssText: string;
     cursor: string;
+    d: string;
     direction: string;
     display: string;
     dominantBaseline: string;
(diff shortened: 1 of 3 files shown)


orta:
Kudos @NotWoods - but your timing was a tad off, we've just merged a **massive** update which includes this  in with https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1034 - so I'm going to close this PR

PR #1037, opened 2021-07-02
fix(1030) - Add CSP disposition field

NotWoods:
Added https://www.w3.org/TR/CSP3/

This makes some optional properties required.


saschanaz:
I think #1034 made this redundant. Feel free to ping if I'm wrong.

@github-actions close

PR #1038, opened 2021-07-02
fix(994) - WebAssembly errors extend Error

### LINKED ISSUE #994: WebAssembly.{CompileError,LinkError,RuntimeError} should extend Error
# Problem

[Current lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/fb60c9f46e827f83a09b076bfb35cf703692f1f1/lib/lib.dom.d.ts#L19134) says both `WebAssembly.CompileError`, `WebAssembly.LinkError`, and `WebAssembly.RuntimeError` are not an `Error`. But [the spec](https://www.w3.org/TR/wasm-js-api-1/#error-objects) defines that they behave as a `NativeError` like `TypeError` etc.

Due to this inconsistency, I can't get even `message` when catching a `WebAssembly.CompileError`. For example:

```ts
try {
  WebAssembly.compile(wasmBin);
} catch(e) {
  if (e instanceof WebAssembly.CompileError){
    console.log(e.message); // Type Error!
  } else {
    throw e;
  }
}
```

# Suggested Changes

So I suggest to change:

```ts
declare namespace WebAssembly {
    interface CompileError {
    }
    ...
    interface LinkError {
    }
    ...
    interface RuntimeError {
    }
    ...
}
```

into:

```ts
declare namespace WebAssembly {
    interface CompileError extends Error {
    }
    ...
    interface LinkError extends Error {
    }
    ...
    interface RuntimeError extends Error {
    }
    ...
}
```

# Question

I've checked both https://github.com/microsoft/TypeScript/blob/master/CONTRIBUTING.md#modifying-generated-library-files and https://github.com/microsoft/TypeScript-DOM-lib-generator#contribution-guidelines , but I can't get sure of the right instructions to fix. I tried `npm run fetch-idl "WebAssembly JavaScript Interface" && npm run build && npm run baseline-accept`, but couldn't make any change to fix it. So should I edit `overridingTypes.json`, right?

NotWoods:
Fixes #994

[Errors in WebAssembly](https://webassembly.github.io/spec/js-api/#exceptiondef-compileerror) are located under the WebAssembly namespace and can't be expressed in WebIDL, according to the spec. Here, they've been overriden to extend the Error class as specified in the spec.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -18776,7 +18776,7 @@ declare namespace CSS {
 }
 
 declare namespace WebAssembly {
-    interface CompileError {
+    interface CompileError extends Error {
     }
 
     var CompileError: {
@@ -18803,7 +18803,7 @@ declare namespace WebAssembly {
         new(module: Module, importObject?: Imports): Instance;
     };
 
-    interface LinkError {
+    interface LinkError extends Error {
     }
 
     var LinkError: {
@@ -18832,7 +18832,7 @@ declare namespace WebAssembly {
         imports(moduleObject: Module): ModuleImportDescriptor[];
     };
 
-    interface RuntimeError {
+    interface RuntimeError extends Error {
     }
 
     var RuntimeError: {
--- a/inputfiles/overridingTypes.json
+++ b/inputfiles/overridingTypes.json
@@ -2615,6 +2615,21 @@
                         }
                     }
                 }
+            },
+            "CompileError": {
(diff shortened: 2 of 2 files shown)


orta:
Kudos @NotWoods - but your timing was a tad off, we've just merged a **massive** update which includes this  in with https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1034 which includes this change - so I'm going to close this PR

PR #1049, opened 2021-07-02
Update dependencies

saschanaz:


orta:
I think we _just_ missed #1047 being merged at the same time which added you to the package lock, try rebase

orta:
The bot is set to only comment when a PR opens (again, it's very simple) the auth should be working right though, so try the LGTM when its green

saschanaz:
I guess we need a way to escape L1GTM 😄

orta:
7GTM

saschanaz:
LGTM

orta:
ace

PR #1051, opened 2021-07-03
Restore WebGLRenderingContextBase#canvas

### LINKED ISSUE #1050: Removed 'canvas' property from 'IWebGL2RenderingContext'
I tried building `xterm.js` recently, and got the following error:

```
addons/xterm-addon-webgl/src/GlyphRenderer.ts:305:26 - error TS2339: Property 'canvas' does not exist on type 'IWebGL2RenderingContext'.

305     gl.viewport(0, 0, gl.canvas.width, gl.canvas.height);   
```

Looks like the [`canvas`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/canvas) property was removed as per https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029, but was that intentional? MSDN says

> WebGL 2 is an extension to WebGL 1. The WebGL2RenderingContext interface implements all members of the WebGLRenderingContext interface. Some methods of the WebGL 1 context can accept additional values when used in a WebGL 2 context. You will find this info noted on the WebGL 1 reference pages.

saschanaz:
Closes #1050.

### DIFF
--- a/inputfiles/addedTypes.json
+++ b/inputfiles/addedTypes.json
@@ -179,6 +179,17 @@
                     ]
                 }
             },
+            "WebGLRenderingContextBase": {
+                "properties": {
+                    "property": {
+                        "canvas": {
+                            "name": "canvas",
+                            "exposed": "Window",
+                            "type": "HTMLCanvasElement"
+                        }
+                    }
+                }
+            },
             "WindowEventHandlers": {
                 "events": {
                     "event": [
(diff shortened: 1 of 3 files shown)


inputfiles/addedTypes.json
```diff
@@ -179,6 +179,17 @@
                     ]
                 }
             },
+            "WebGLRenderingContextBase": {
+                "properties": {
+                    "property": {
+                        "canvas": {
+                            "name": "canvas",
```
DanielRosenwasser, line 186:
```suggestion
                            "name": "canvas",
                            "readonly": true,
```

saschanaz:
LGTM

PR #1056, opened 2021-07-06
Generate changelogs per release

orta:
Add support for re-using the CHANGELOG infra for the releases - https://github.com/microsoft/TypeScript-DOM-lib-generator/releases

### DIFF
--- a/deploy/createTypesPackages.mjs
+++ b/deploy/createTypesPackages.mjs
@@ -3,7 +3,7 @@
 // node deploy/createTypesPackages.mjs
 
 // prettier-ignore
-const packages = [
+export const packages = [
     {
       name: "@types/web",
       description: "Types for the DOM, and other web technologies in browsers",
@@ -71,8 +71,6 @@ const go = async () => {
   }
 };
 
-go();
-
 async function updatePackageJSON(packagePath, pkg, gitSha) {
   const pkgJSONPath = join(packagePath, "package.json");
   const packageText = fs.readFileSync(pkgJSONPath, "utf8");
@@ -130,4 +128,8 @@ function copyREADME(pkg, pkgJSON, writePath) {
     );
 
   fs.writeFileSync(writePath, readme);
+}
+
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
+  go();
 }
--- a/deploy/createTypesPackages.mjs
+++ b/deploy/createTypesPackages.mjs
@@ -131,5 +131,5 @@ function copyREADME(pkg, pkgJSON, writePath) {
 }
 
 if (process.argv[1] === fileURLToPath(import.meta.url)) {
-  go();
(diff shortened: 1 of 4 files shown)


src/changelog.ts
```diff
@@ -169,14 +177,9 @@ export function generate(): string {
   }
 
   const output = outputs.join("\n\n");
-
-  if (!output.length) {
```
saschanaz, line 173: Could you move this logic to the caller?

saschanaz (approved):
Approving with a nit.

I still think this can be done without unpkg by simply replacing the unpkg part with the `getLatestTag()`, but that's a separate issue.

deploy/deployChangedPackages.mjs
```diff
@@ -93,8 +102,10 @@ Assuming that this means we need to upload this package.`);
 
       uploaded.push(dirName);
     }
-  }
 
+    console.log("\n# Release notes:");
```
sandersn, line 106: I don't know the code well enough -- but it looks like this prints the release notes but doesn't upload them. Is that the intent?

orta: It does! I just figured it'd be nice to have that in the CI logs also

orta:
OK, this is updated to use the git tags and all feedback is resolved 👍🏻

deploy/createTypesPackages.mjs
```diff
@@ -131,3 +129,7 @@ function copyREADME(pkg, pkgJSON, writePath) {
 
   fs.writeFileSync(writePath, readme);
 }
+
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
+  go();
```
saschanaz, line 134:
```suggestion
  await go();
```

Using top level `await` can be helpful as it changes exit code when throws. [It's supported by Node.js 14+.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#browser_compatibility)

deploy/deployChangedPackages.mjs
```diff
@@ -40,28 +44,51 @@ const go = async () => {
...
+      );
+
       const generatedDTSPath = join(generatedDir, dirName, file);
       const generatedDTSContent = fs.readFileSync(generatedDTSPath, "utf8");
-      const unpkgURL = `https://unpkg.com/${pkgJSON.name}/${file}`;
+
+      // This assumes we'll only _ever_ ship patches, which may change in the
+      // future someday.
+      const [maj, min, patch] = pkgJSON.version.split(".");
+      const olderVersion = `${maj}.${min}.${patch - 1}`;
```
saschanaz, line 72: (Just to note for future reference: `git describe --tags --match "@types/web@*" --abbrev=0` can do the same thing)

saschanaz (approved):
Cool!

PR #1057, opened 2021-07-06
Documents the release to TS process, and adds support for easily generating a CHANGELOG for 2 arbitrary versions 

orta:
https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/97f733707a840650c8995f290a28dc3ee44b502a is the only commit which matters (I just wanted to build on top of #1056 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ This tool is used to generate the web-based `lib.dom.d.ts` file which is include
 
 ## Why is my fancy API still not available here?
 
-A feature needs to be supported by more than two major browser engines to be included here, to make sure there is a good consensus among vendors: __Gecko__ (Firefox), __Blink__ (Chrome/Edge), and __WebKit__ (Safari).
+A feature needs to be supported by two or more major browser engines to be included here, to make sure there is a good consensus among vendors: __Gecko__ (Firefox), __Blink__ (Chrome/Edge), and __WebKit__ (Safari).
 
-If the condition is met but still is not available here, please [file an issue](hthttps://github.com/microsoft/TypeScript-DOM-lib-generator/issues/new).
+If the condition is met but still is not available here, first check the heuristics below and then please [file an issue](hthttps://github.com/microsoft/TypeScript-DOM-lib-generator/issues/new).
 
 ## Build Instructions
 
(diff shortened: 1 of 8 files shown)


deploy/versionChangelog.mjs
```diff
@@ -0,0 +1,37 @@
...
+    const npmToFileText = await npmToFileReq.text();
+
+    const title = `\n## \`${file.path.slice(1)}\`\n`;
+    const notes = generateChangelogFrom(npmBeforeFileText, npmToFileText);
+
+    console.log(title);
+    console.log(notes.trim() === "" ? "No changes" : notes);
+  }
+};
+go();
```
saschanaz, line 37:
```suggestion
await go();
```

Or maybe even without `go()` since top level await is already a thing in Node.js.

README.md
```diff
@@ -4,9 +4,9 @@ This tool is used to generate the web-based `lib.dom.d.ts` file which is include
 
 ## Why is my fancy API still not available here?
 
-A feature needs to be supported by more than two major browser engines to be included here, to make sure there is a good consensus among vendors: __Gecko__ (Firefox), __Blink__ (Chrome/Edge), and __WebKit__ (Safari).
+A feature needs to be supported by two or more major browser engines to be included here, to make sure there is a good consensus among vendors: __Gecko__ (Firefox), __Blink__ (Chrome/Edge), and __WebKit__ (Safari).
 
-If the condition is met but still is not available here, please [file an issue](hthttps://github.com/microsoft/TypeScript-DOM-lib-generator/issues/new).
+If the condition is met but still is not available here, first check the heuristics below and then please [file an issue](hthttps://github.com/microsoft/TypeScript-DOM-lib-generator/issues/new).
```
saschanaz, line 9: I don't think we need that heuristics anymore. If something is supported well by browsers then it should be supported here, there is no good reason not to do so IMO.

deploy/versionChangelog.mjs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const go = async () => {
+  const allFiles = `https://unpkg.com/${name}/?meta`;
+  const npmFileReq = await fetch(allFiles);
+  const npmDTSFiles = await npmFileReq.json();
+  const dtsFiles = npmDTSFiles.files.filter((f) => f.path.endsWith(".d.ts"));
+
+  for (const file of dtsFiles) {
+    const beforeURI = `https://unpkg.com/${name}@${before}${file.path}`;
```
saschanaz, line 22: Okay this is a new code and I think it can use git tags. The required tools are already in `version.ts`.

deploy/versionChangelog.mjs
```diff
@@ -0,0 +1,37 @@
...
+    const npmBeforeFileReq = await fetch(beforeURI);
+    const npmBeforeFileText = await npmBeforeFileReq.text();
+
+    const toURI = `https://unpkg.com/${name}@${to}${file.path}`;
+    const npmToFileReq = await fetch(toURI);
+    const npmToFileText = await npmToFileReq.text();
+
+    const title = `\n## \`${file.path.slice(1)}\`\n`;
```
saschanaz, line 30: No need for this declaration since it has one single use?

saschanaz (changes requested):

orta:
OK, updated - let me keep the heuristics request in the README for now and I'll bring it up in the discord about dropping it. I think I'm on side 'drop' but wanna cover bases.

orta:
OK, I'm going to merge this because #1064 has all the changes for the debate 👍🏻

PR #1059, opened 2021-07-09
Allow strings in deprecated field

saschanaz:
No more mixed deprecation notes in `comment`.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2230,8 +2230,10 @@ declare var AudioParamMap: {
     new(): AudioParamMap;
 };
 
-/** The Web Audio API events that occur when a ScriptProcessorNode input buffer is ready to be processed.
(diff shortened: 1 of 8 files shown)


baselines/dom.generated.d.ts
```diff
@@ -6302,7 +6304,7 @@ declare var HTMLDetailsElement: {
     new(): HTMLDetailsElement;
 };
 
-/** @deprecated - this is not available in most browsers */
+/** @deprecated this is not available in most browsers */
```
saschanaz, line 6307: @orta Do you agree with this change? It seems more like [the standard format](https://jsdoc.app/tags-deprecated.html).

> ```js
> /**
>  * @deprecated since version 2.0
>  */
> function old() {
> }
> ```

orta: Yep, the `-` kinda gets put in an odd place in IDEs

Screen Shot 2021-07-09 at 1 30 51 PM

orta (approved):

saschanaz:
LGTM

orta:
haha

PR #1061, opened 2021-07-09
Use JSONC file extension for input json files

saschanaz:
Comments in JSON cause warnings on both IDE and GitHub and thus degrades dev experience. Changing the file extension should fix it.


package.json
```diff
@@ -54,7 +54,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^3.4.0",
     "jsdom": "^16.6.0",
-    "json5": "^2.2.0",
+    "jsonc-parser": "^3.0.0",
```
saschanaz, line 57: This is seemingly the parser used by VSCode for "JSON with Comments" parsing mode.

PR #1062, opened 2021-07-09
Adds targets for the other environments

orta:
Adds support for deploying:

 - `@types/webworker`
 - `@types/serviceworker`
 - `@types/audioworklet`
 - `@types/sharedworker`

[`@types/sharedworker` already exists](https://www.npmjs.com/package/@types/sharedworker) and was removed from DT by me back in 3.9 - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43366 - the deployment infra will see this and release this version as `0.0.30`.


orta:
Going to give this the weekend for TS team members to give any feedback, will merge on Monday 👍🏻

deploy/createTypesPackages.mjs
```diff
@@ -13,6 +13,42 @@ export const packages = [
         { from: "../generated/dom.iterable.generated.d.ts", to: "index.iterable.d.ts" }
       ],
     },
+    {
+      name: "@types/webworker",
```
saschanaz, line 17: Probably need to resolve https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/997#discussion_r602899470 first before publishing webworker type, since currently it's a mixture of all types of workers:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/bc78f9cffbc8cb9bcfce12a1780a77e9e2620344/src/build.ts#L253-L257

orta: OK, cool, I'll remove this one from the object right now which should unblock the PR

deploy/createTypesPackages.mjs
```diff
@@ -10,7 +10,34 @@ export const packages = [
...
+      { from: "../generated/dom.iterable.generated.d.ts", to: "index.iterable.d.ts" },
+    ],
+  },
+  {
+    name: "@types/serviceworker",
+    description: "Types for the global scope of Service Workers",
+    readme: "./readmes/serviceworker.md",
+    files: [
+      { from: "../generated/serviceworker.generated.d.ts", to: "index.d.ts" },
+      { from: "../generated/serviceworker.iterable.generated.d.ts", to: "index.iterable.d.ts" },
```
saschanaz, line 22: While you are at it: how can `index.iterable.d.ts` be referenced? 👀 Should the step be added to readme?

orta: Hrm, I thought it would be automatically be included in all builds because it lives in @types and we read all the files in there. Haven't _actually checked though.

saschanaz: Somehow it doesn't work like that in my environment:

orta: Agree, the build pipeline will probably need to add something like `/// <reference path="..." />` to the `index.dt.s`

A user can probably get these imported into their globals via `import "web/index.iterable"` but that's weaksauce

saschanaz: But that would mean packages will require es6+ 🤔

orta: I'm still figuring out how to have it happen automatically `/// <reference types="web/index.iterable" />` would also work in any target

this pattern may be a good replacement for `@types/ie11` too.

saschanaz: Yeah, since it's probably time to concentrate on es6+ 😁

PR #1063, opened 2021-07-09
Don't run the TS tests unless baselines change

orta:
Should reduce CI times for a some PRs

### DIFF
--- a/.github/workflows/test_typescript.yml
+++ b/.github/workflows/test_typescript.yml
@@ -1,5 +1,12 @@
 name: Runs with TypeScript Tests
-on: [push, pull_request]
+on:
+  push:
+    paths:
+      - "baselines/*"
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


saschanaz:
I wonder doing this will skip the tests when someone only updates `inputfiles/` and forgets to update baselines.

saschanaz:
Probably should also cover `inputfiles` and `package-lock.json` since updating webref and bcd also affects baselines.

saschanaz:
Or maybe better to filter out irrelevant paths by `paths-ignore`.

orta:
Any changes to the other files without a corresponding baseline change would mean that the normal `npm test` CI would fail, making the build red anyway. At that point there's no need to run the run these more intense tests.

saschanaz:
Ah yeah, I misunderstood 👀 Okay then.

PR #1064, opened 2021-07-09
Update README to match the new code

saschanaz:
This should match better with the code after #1034.


README.md
```diff
@@ -34,82 +34,37 @@ To deploy:
...
 For each pull request, we will run the script and compare the generated files with the baseline files.
 In order to make the tests pass, please update the baseline as well in any pull requests.
 
-It's recommended to first check which spec the wrong type belongs to. Say we are to update `IntersectionObserver` which belongs to [`Intersection Observer`](https://www.w3.org/TR/intersection-observer/) spec, and then we can do:
+### When the type is missing
 
-1. First check we have the spec name `Intersection Observer` in `inputfiles/idlSources.json`. If not, add it.
-2. Run `npm run fetch-idl "Intersection Observer" && npm run build && npm run baseline-accept`.
+It's possible that the automated algorithm decided that it's not well supported by browsers and thus removed it. Say we we want to add a new interface named `Foo`. Check if there is a document about that interface in [MDN](https://developer.mozilla.org/). If there is, see the browser compatibility section and check whether it's supported by two or more browsers. (Chrome, Firefox, and Safari are the relevant browsers. Supports from multiple Chromium-based browsers are counted as one support.)
```
HolgerJeromin, line 47:
```suggestion
It's possible that the automated algorithm decided that it's not well supported by browsers and thus removed it. Say we we want to add a new interface named `Foo`. Check if there is a document about that interface in [MDN](https://developer.mozilla.org/). If there is, see the browser compatibility section and check whether it's supported by two or more browsers. (Chrome, Firefox, and Safari are the relevant browsers. Support from multiple Chromium-based browsers are counted as one support.)
```

orta: Maybe "Support from multiple Chromium-based browsers are counted as one browser" - using support at the end  doesn't really fit in English grammar.

saschanaz:
@orta @sandersn Do you see something to add or modify here? 👀

README.md
```diff
@@ -38,82 +38,37 @@ npm run test
...
 
-0. Open or refer to an issue in the [TypeScript repo](https://github.com/Microsoft/TypeScript) which might get moved back to this repo.
-1. Update an IDL or add missing elements to `inputfiles/addedTypes.json`, overriding elements to `inputfiles/overridingTypes.json`, or elements to remove to `inputfiles/removedTypes.json`.
-2. Run the build script locally to obtain new `dom.generated.d.ts` and `webworker.generated.d.ts`.
-3. Update the files in the `baselines` folder using the newly generated files under `generated` folder (`npm run baseline-accept`).
+It's possible that the type is too specific or too general. First you need to check whether `inputfiles/overridingTypes.jsonc` or `inputfiles/addedTypes.jsonc` have a relevant item, which can be fixed if exists. If they don't, add one in `overridingTypes.jsonc`. Run `npm run build && npm run baseline-accept` to make sure the resulting changes are what you want.
 
-### What are the TypeScript team's heuristics for PRs to the DOM APIs
-
-Changes to this repo can have pretty drastic ecosystem effects, because these types are included by default in TypeScript.
```
sandersn, line 66: This paragraph is mostly still true, right? I know we are less conservative than before, but the idea is that every update to `@types/web` will *eventually* ship with Typescript by default means that there are some changes that will not be OK until there's a way to publish other packages from here.

sandersn (changes requested):
I'm not too sure about the removal of heuristics, which still somewhat apply. @orta can you take a look too?

README.md
```diff
@@ -38,82 +38,37 @@ npm run test
...
 ## Input Files
 
-- `browser.webidl.preprocessed.json`: a JSON file generated by Microsoft Edge. **Do not edit this file**.
-    - Due to the different update schedules between Edge and TypeScript, this may not be the most up-to-date version of the spec.
 - `mdn/apiDescriptions.json`: a JSON file generated by fetching API descriptions from [MDN](https://developer.mozilla.org/en-US/docs/Web/API). **Do not edit this file**.
-- `addedTypes.json`: types that should exist in either browser or webworker but are missing from the Edge spec. The format of the file mimics that of `browser.webidl.preprocessed.json`.
-- `overridingTypes.json`: types that are defined in the spec file but has a better or more up-to-date definitions in the json files.
-- `removedTypes.json`: types that are defined in the spec file but should be removed.
+- `addedTypes.jsonc`: types that should exist but are missing from the spec data.
+- `overridingTypes.jsonc`: types that are defined in the spec but has TypeScript-friendly modifications in the json files.
```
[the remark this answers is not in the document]
saschanaz: It can't be "more up to date" anymore since the data itself is being constantly autoupdated while the JSON file is handled manually. Removing that word makes it "have better definitions" which is a vague expression.

README.md
saschanaz: Yes, maybe it's good to keep the sentence for maintainers. It has minimal meaning for contributors though, since nearly all changes will now be from webref and MDN rather than manual PRs.

saschanaz: >I know we are less conservative than before

I'm not sure, as quite a lot types are removed for lack of broad browser support?

orta (approved):
OK, yeah, I think this probably what we want

README.md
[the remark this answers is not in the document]
saschanaz:
```suggestion
- `overridingTypes.jsonc`: types that are defined in the spec but have TypeScript-friendly modifications in the json files.
```

saschanaz:
Cool, I'll merge this if @sandersn also agrees.

sandersn (approved):

saschanaz:
LGTM

PR #1065, opened 2021-07-09
Use new standard URL objects for paths

saschanaz:


deploy/createTypesPackages.mjs
```diff
@@ -4,22 +4,21 @@
 
 // prettier-ignore
```
saschanaz, line 5: Why ignore this?

orta: I'm a 1 line 1 expression kinda person, wanted this to read more naturally given the 80 char constraints from the defaults

saschanaz: Probably better to add this directly to `files` then, or is there a reason to keep it here?

orta: No reason - though #1062 expands on it

saschanaz: Hmm, I'll just keep it so that no need to duplicate it.

orta: I remembered why actually, it's so that this config is considered part of the deployment infra and can't get changed without our ✅

Issue #1067, opened 2021-07-15
Google Feedback on 4.4 Beta


DanielRosenwasser:
Opening on the behalf of @h-joo from https://github.com/microsoft/TypeScript/issues/45047:

> ### Readonly attributes in DOMPoint, DOMMatrix, DOMRect
> 
> Some properties of DOMPoint, DOMMatrix, DOMRect became readonly, and this no
> longer match the specification. This is because these types no longer override
> properties from the extended types to be read/write.
> 
> Compare(DOMPoint):
> 
> *   https://github.com/microsoft/TypeScript/blob/release-4.3/lib/lib.dom.d.ts#L3979
> *   https://github.com/microsoft/TypeScript/blob/release-4.4/lib/lib.dom.d.ts#L3886
> 
> Compare(DOMMatrix):
> 
> *   https://github.com/microsoft/TypeScript/blob/release-4.3/lib/lib.dom.d.ts#L3854
> *   https://github.com/microsoft/TypeScript/blob/release-4.4/lib/lib.dom.d.ts#L3783
> 
> Compare(DOMRect):
> 
> *   https://github.com/microsoft/TypeScript/blob/release-4.3/lib/lib.dom.d.ts#L4026
> *   https://github.com/microsoft/TypeScript/blob/release-4.4/lib/lib.dom.d.ts#L3929
> 
> ### Removed Types
> 
> SpeechRecognition, RTCError got removed from lib.dom.d.ts. There are existing
> code which use this API, and it would be difficult to migrate without proper
> type support either from the compiler or from definitelyTyped.
> 
> *   https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionEvent
> *   https://developer.mozilla.org/en-US/docs/Web/API/RTCError
> 
> ### Offscreen Rendering API changes
> 
> OffscreenCanvas still exists, but it's marked as deprecated. Though it is no
> longer a value, i.e. code calling `new OffscreenCanvas` no longer works, it's
> just an interface. We think this might be unintentional. Though, we would still
> be able to migrate via referring to the types defined in
> [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped).
> 
> *   https://github.com/microsoft/TypeScript/blob/release-4.3/lib/lib.dom.d.ts#L11087
> *   https://github.com/microsoft/TypeScript/blob/release-4.4/lib/lib.dom.d.ts#L10832
> 
[...]

PR #1071, opened 2021-07-17
Add HTMLElement-returning getElementById to DocumentFragment

saschanaz:
This is copying what `document.getElementById` and `document.body.ownerDocument` currently looks like.

Regressed by b598925.

Fixes the DocumentFragment issue raised in #1067.


saschanaz:
LGTM

PR #1073, opened 2021-07-17
Restore `document.caretRangeFromPoint()`

### LINKED ISSUE #1072: `caretRangeFromPoint` return type should be `Range | null`
# `caretRangeFromPoint` Update Request

## Configuration Check

My compilation *target* is `ES2020` and my *lib* is `caretRangeFromPoint`.

## Missing / Incorrect Definition

`caretRangeFromPoint` is currently defined as:
```
caretRangeFromPoint(x: number, y: number): Range;
```
but should be
```
caretRangeFromPoint(x: number, y: number): Range | null;
```

## Sample Code

```
// Calculate range and friends here after we've made our fake textarea/input divs.
range = document.caretRangeFromPoint(ev.clientX, ev.clientY);
const startNode = range.startContainer;
```
This code should have a type error but currently doesn't since library doens't show return type as nullable. This happens quite often in the wild.

## Documentation Link

https://developer.mozilla.org/en-US/docs/Web/API/Document/caretRangeFromPoint also shows null as possible return type.

saschanaz:
Fixes #1072

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4439,6 +4439,8 @@ interface Document extends Node, DocumentAndElementEventHandlers, DocumentOrShad
     /** @deprecated */
     captureEvents(): void;
     /** @deprecated */
+    caretRangeFromPoint(x: number, y: number): Range | null;
+    /** @deprecated */
     clear(): void;
     /**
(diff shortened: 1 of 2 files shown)


saschanaz:
LGTM

PR #1074, opened 2021-07-17
Update webref/idl to 2.2.8

saschanaz:
@orta could you double check that the token below has the `public-repo` permission, since [it's failing to open PRs](https://github.com/microsoft/TypeScript-DOM-lib-generator/runs/3092249820?check_suite_focus=true)? Thanks!

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/1f0846bdf369f4d78876465ec7d50ec66aae3cfa/.github/workflows/update-core-deps.yml#L35


saschanaz:
LGTM

PR #1078, opened 2021-07-29
improve fetch body

septs:
the not is a break changes, but can improve developer experience

```typescript
const response = fetch(...)
// old method
const result = response.body() as Promise<Result>
// new method
const result = response.body<Result>()
```

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2427,7 +2427,7 @@ interface Body {
     arrayBuffer(): Promise<ArrayBuffer>;
     blob(): Promise<Blob>;
     formData(): Promise<FormData>;
-    json(): Promise<any>;
+    json<T = any>(): Promise<T>;
     text(): Promise<string>;
 }
 


saschanaz:
This is not a recommended pattern, see https://github.com/definitelytyped/definitelytyped#common-mistakes

> * `getMeAT<T>(): T`:
>   If a type parameter does not appear in the types of any parameters, you don't really have a generic function, you just have a disguised type assertion.
>   Prefer to use a real type assertion, e.g. `getMeAT() as number`.
>   Example where a type parameter is acceptable: `function id<T>(value: T): T;`.
>   Example where it is not acceptable: `function parseJson<T>(json: string): T;`.
>   Exception: `new Map<string, number>()` is OK.

septs:
@sandersn the pr can improve developer experience

```typescript
const response = fetch(...)
// old method
const result = response.body() as Promise<Result>
// new method
const result = response.body<Result>()
```

PR #1082, opened 2021-08-01
Mark legacy aliases with @deprecated

saschanaz:
This is kinda new, @orta @sandersn do you think this is good to have?

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3033,49 +3033,49 @@ interface CSSStyleDeclaration {
     userSelect: string;
     verticalAlign: string;
     visibility: string;
-    /** @deprecated */
+    /** @deprecated This is a legacy alias of `alignContent`. */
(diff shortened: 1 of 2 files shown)


baselines/dom.generated.d.ts
```diff
@@ -17223,7 +17226,9 @@ interface Window extends EventTarget, AnimationFrameProvider, GlobalEventHandler
     readonly orientation: number;
     readonly outerHeight: number;
     readonly outerWidth: number;
+    /** @deprecated This is a legacy alias of `scrollX`. */
     readonly pageXOffset: number;
+    /** @deprecated This is a legacy alias of `scrollY`. */
     readonly pageYOffset: number;
```
orta, line 17232: Interesting, TIL'd

orta (approved):
I think this is a good idea

saschanaz:
LGTM

👍

PR #1084, opened 2021-08-01
Pass string to createRelease()

saschanaz:
Fixes #1083

octokit was a phantom dependency 😱

PR #1085, opened 2021-08-01
Lint deploy/ scripts

saschanaz:
This changes .mjs to .js because

* jsconfig does not support .mjs
* they don't need to be .mjs because there is `"type": "module"` in package.json already.


orta:
Looks good, I can't merge due to conflicts from #1084  though 👍🏻

saschanaz:
#1084 broke the deploy action because of octokit version, this should fix it.

orta:
Ace

PR #1089, opened 2021-08-03
Removes the nullable attribute from window.parent

orta:
Re: https://github.com/microsoft/TypeScript/pull/44684#discussion_r681871052

I read the spec, and it really didn't seem like null was possible - it is either the parent, or itself:

https://dev.w3.org/html5/spec-LC/browsers.html#dom-parent

> The parent IDL attribute on the Window object of a Document in a browsing context b must return the WindowProxy object of the parent browsing context, if there is one (i.e. if b is a child browsing context), or the WindowProxy object of the browsing context b itself, otherwise (i.e. if it is a top-level browsing context or a detached nested browsing context).

Maybe this could be true in environments other than dom?


saschanaz:
This is the latest spec: https://html.spec.whatwg.org/multipage/browsers.html#dom-parent

It has an example, which I modified to allow copy-pasting to console:

```js
var element = document.createElement("iframe");
document.body.append(element);
var iframeWindow = element.contentWindow;
element.remove();

console.log(iframeWindow.top === null);
console.log(iframeWindow.parent === null);
console.log(iframeWindow.frameElement === null);
```

All returns true. But admittedly a rare example, so your call. I guess at least it should be in the comment that it can be null?

orta:
Yeah, thanks for digging!

I'll add a comment saying it's possibly null - I think this is probably worth the ergonomics change

saschanaz:
Should it say why it's not nullable in the lib file?

baselines/dom.generated.d.ts
```diff
@@ -17230,7 +17230,8 @@ interface Window extends EventTarget, AnimationFrameProvider, GlobalEventHandler
     readonly pageXOffset: number;
     /** @deprecated This is a legacy alias of `scrollY`. */
     readonly pageYOffset: number;
-    readonly parent: WindowProxy | null;
+    /** Refers to either the parent WindowProxy, or itself. If you create a child context, then remove the parent you could get null though. */
```
saschanaz, line 17233: What about:

> Refers to either the parent WindowProxy, or itself.
>
> It can rarely be null e.g. for `window` for an iframe that is already removed from the parent.

orta: Having it multiline doesn't really work because it's in two places with different indentations, so I made this a single line

saschanaz: I think `\n` automatically does the indentation.

saschanaz: e.g. postMessage

inputfiles/comments.json
```diff
@@ -1796,6 +1796,15 @@
                         }
                     }
                 }
+            },
+            "Window": {
+                "properties": {
+                    "property": {
+                        "parent": {
+                            "comment" : "Refers to either the parent WindowProxy, or itself. It can rarely be null e.g. for contentWindow of an iframe that is already removed from the parent."
```
saschanaz, line 1804:
```suggestion
                            "comment" : "/**\n * Refers to either the parent WindowProxy, or itself.\n *\n * It can rarely be null e.g. for contentWindow of an iframe that is already removed from the parent.\n */"
```

Sigh, the multiline comment code is weird, it only autoindents comments that manually deal with `/** */`.

saschanaz (approved):
LGTM

PR #1088, opened 2021-08-03
Make ChildNode and ParentNode extend Node again

DanielRosenwasser:
Fixes #1087

PR #1090, opened 2021-08-03
Remove unnecessary 'name' properties.

DanielRosenwasser:
Addresses some tech debt @saschanaz brought up in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1088#discussion_r681360802.

I don't necessarily know if I got everything, but I figure it's a good start.


saschanaz:
Thanks! LGTM.

PR #1092, opened 2021-08-04
Always emit "| undefined" for optional properties

sandersn:
Makes exactOptionalPropertyTypes work better with `Window` and `window`


saschanaz:
LGTM

PR #1093, opened 2021-08-04
Remove static abort from AbortSignal

sandersn:
@types/node doesn't have it, and it's not worth it to make @types/node create a new typingsVersion.

### DIFF
--- a/deploy/createTypesPackages.js
+++ b/deploy/createTypesPackages.js
@@ -90,6 +90,7 @@ const go = async () => {
     });
 
     prependAutoImports(pkg, packagePath);
+    postProcessDTSFiles(pkg, packagePath);
 
     // Setup the files in the repo
     const newPkgJSON = await updatePackageJSON(pkg, packagePath);
@@ -186,6 +187,30 @@ function prependAutoImports(pkg, packagePath) {
 
   let indexText = fs.readFileSync(index, "utf-8");
   fs.writeFileSync(index, `${toPrepend}\n\n${indexText}`);
+}
+
+/**
+ * Handles any post-processing we do for deployment.
+ * @param {Package} pkg
+ * @param {URL} packagePath
+ */
+function postProcessDTSFiles(pkg, packagePath) {
+  iterateThroughFiles((content) => {
+    return content.replace(
+      "abort(): AbortSignal;",
+      "// abort(): AbortSignal; - To be re-added in the future"
+    );
+  });
+
+  /** @param {(str:string) => string} contentReplacer */
+  function iterateThroughFiles(contentReplacer) {
+    pkg.files.forEach((fileRef) => {
+      const dtsFileURL = new URL(fileRef.to, packagePath);
+      let dtsContent = fs.readFileSync(dtsFileURL, "utf-8");
+      dtsContent = contentReplacer(dtsContent);
+      fs.writeFileSync(dtsFileURL, dtsContent);
+    });
+  }
 }
 
 if (process.argv[1] === fileURLToPath(import.meta.url)) {
--- a/inputfiles/removedTypes.jsonc
+++ b/inputfiles/removedTypes.jsonc
@@ -166,14 +166,6 @@
             "Text": {
                 "implements": ["GeometryUtils"]
             },
-            // This is hard to get in sync with @types/node
-            "AbortSignal": {
-                "methods": {
-                    "method": {
-                        "abort": null
-                    }
-                }
-            },
             "WebGLBuffer": {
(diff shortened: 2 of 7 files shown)


inputfiles/removedTypes.jsonc
```diff
@@ -166,6 +166,14 @@
             "Text": {
                 "implements": ["GeometryUtils"]
             },
+            // This is hard to get in sync with @types/node
+            "AbortSignal": {
+                "methods": {
+                    "method": {
+                        "abort": null
```
saschanaz, line 173: I think this one should be preferred over node, as Node is just following the web standard? What happened there?

SimonSchick:
But node does have this method?
```
Type ".help" for more information.
> AbortSignal.abort()
AbortSignal { aborted: true }
```

saschanaz:
> But node does have this method?

Web browsers and Node.js have the method defined as https://dom.spec.whatwg.org/#interface-AbortSignal .

orta:
Yeah, agree - its more that we didn't think adding this was worth forcing `@types/node` to have to switch to TS 4.5 for their types and then requiring _many_ projects to update for this one property. Not that it's incorrect

saschanaz:
Can it be fixed locally in TS rather than affecting @types/web?

orta:
It's a normal assignability bug, not a change in TypeScript (you can see a [tonne in here](https://github.com/typescript-bot/TypeScript/pull/197/files#diff-01b0ab09324fb295447d930f7e8edddbe4093fa413b59a6d2bfc2d66762dc712R3))

```
node_modules/@types/node/globals.d.ts(85,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.
```

Basically it means that anyone with an older version of `@types/node` cannot upgrade to TS 4.5 without updating their node types, and it means that the `@types/node` would need to bump their minimum TS version to 4.5 for this one property

saschanaz:
I mean, we'll have to add that to node one day, so it would be good if TS have a local patch that removes abort() so that anyone need it can install @types/web?

saschanaz:
That can happen in deploy script IMO.

SimonSchick:
Can we not change the type of `AbortSignal` to be an interface and use declaration merging?

saschanaz:
> Can we not change the type of `AbortSignal` to be an interface and use declaration merging?

Sounds like an option tbh.

sandersn:
Does the code have the ability to generate an interface for the static side too? If not, I guess we could add that.

saschanaz:
No such ability, I think AbortSignal can be special cased.

sandersn:
oh, I forgot about addedTypes.jsonc. That should work.

orta:
So, I think we're going to have to take this PR as-is because I don't think we're going to get a good answer which doesn't force users to update node or force a typesVersion on node before the release today. We shipped a build to TypeScript _with_ these changes in, because we don't want to force the node update on anyone migrating to the new TS build, and I want to have corresponding `@types/web` to refer to it.

I'll make a PR reverting it after we're sure the RC dom is going to be what we ship in TS. 

I brought up https://github.com/microsoft/TypeScript/issues/43972 in a TS language design meeting on Wednesday and there was general agreement to explore finding the common subset and extract it in here which is going to probably be tricky work in this repo. https://github.com/microsoft/TypeScript/issues/45328

saschanaz:
> I want to have corresponding `@types/web` to refer to it.

I'd prefer manual patching in deploy script if this isn't a strict requirement.

orta:
It's not

saschanaz:
Looks good, please reformat 😉

saschanaz:
LGTM

deploy/createTypesPackages.js
```diff
@@ -188,6 +189,30 @@ function prependAutoImports(pkg, packagePath) {
...
+/**
+ * Handles any post-processing we do for deployment.
+ * @param {Package} pkg
+ * @param {URL} packagePath
+ */
+function postProcessDTSFiles(pkg, packagePath) {
+  iterateThroughFiles((content) => {
+    return content.replace(
+      "abort(): AbortSignal;",
+      "// abort(): AbortSignal; - To be re-added in the future"
```
saschanaz, line 201: This maaayy advertise `@types/web` but your call really.

orta:
OK, let's get this in - which will unblock merging some of the other PRs

PR #1095, opened 2021-08-08
fix(createObjectURL): use MDN type declaration

### LINKED ISSUE #916: Add argument types to `URL.createObjectURL`
The function `URL.createObjectURL` currently has an argument of type `any`. However, this is not in line with the types it would like to be parsed.

I suggest changing the argument from type `object: any` to `object: File | Blob | MediaSource`.

This would make make the type more helpful and make it inline with the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL#Parameters) .

I might be missing a type of `null` or `undefined` on the union type but I'm not sure if there is a use case for this.

Friendseeker:
Changes `createObjectURL(object: any): string` to `createObjectURL(object: File | Blob | MediaSource): string`.

Makes type declaration inline with the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL).

Fixes #916.

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -402,19 +402,6 @@
                     }
                 }
             },
-            "URL": {
-                "name": "URL",
-                "methods": {
-                    "method": {
-                        "createObjectURL": {
-                            "exposed": "Window Worker",
-                            "overrideSignatures": [
-                                "createObjectURL(object: File | Blob | MediaSource): string"
-                            ]
-                        }
-                    }
-                }
-            },
             "NodeListOf": {
                 "name": "NodeListOf",
                 "typeParameters": [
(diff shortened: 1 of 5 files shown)


inputfiles/addedTypes.jsonc
```diff
@@ -409,7 +409,7 @@
                         "createObjectURL": {
                             "exposed": "Window Worker",
                             "overrideSignatures": [
-                                "createObjectURL(object: any): string"
+                                "createObjectURL(object: File | Blob | MediaSource): string"
```
saschanaz, line 412: This is actually not needed anymore, please remove the whole URL item here (from line 405 to line 417) and then rebuild.

saschanaz (changes requested):
Thank you for catching this!

inputfiles/addedTypes.jsonc
Friendseeker: Roger. I removed URL, re-ran `npm run build && npm run baseline-accept` and sent out a new commit.

Appreciate your friendly feedback!

saschanaz:
LGTM

PR #1100, opened 2021-08-11
Make DOMException extend Error

### LINKED ISSUE #1099: DOMException should extend Error
Fun fact that code below evaluates to `true`.

```js
DOMException.prototype instanceof Error
&& new DOMException() instanceof Error
```

I'm not familiar with the technical details, but noticed that Web IDL suggest:

> Additionally, if an implementation gives native Error objects special powers or nonstandard properties (such as a `stack` property), it should also expose those on [DOMException](https://heycam.github.io/webidl/#idl-DOMException) instances.
>
> — [DOMException custom bindings](https://heycam.github.io/webidl/#es-DOMException-specialness)

which sounds just like "interface DOMException extends Error".

And this matters in cases like https://github.com/typescript-eslint/typescript-eslint/issues/3712. 👀

Thank you!

PaperStrike:
Fixes #1099

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3675,7 +3675,7 @@ declare var CustomEvent: {
 };
 
 /** An abnormal event (called an exception) which occurs as a result of calling a method or accessing a property of a web API. */
-interface DOMException {
+interface DOMException extends Error {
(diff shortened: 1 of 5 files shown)


saschanaz:
LGTM

Issue #1029, opened 2021-06-27
TypeScript 4.4 libdom changes


orta:
### Beta

With TypeScript 4.4, we coordinated some pretty [fundamental infrastructure work](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1023) to prepare for the version of `dom.d.ts` with the future of the generator repo. This work resulted in two note-worthy items:

 - The `dom.d.ts` files typically accessed inside TypeScript are now _also_ [deployed to `@types/web`](https://www.npmjs.com/package/@types/web) starting with version [`0.0.1`](https://www.npmjs.com/package/@types/web/v/0.0.1) which corresponds to the version of `libdom.d.ts` in TypeScript 4.3. This means an update to TypeScript 4.4 _does not_ mean you have to update all of your web `.d.ts` files.
 
   The name change is to more accurately reflect the scope of the APIs, for a long time the file has provided much more APIs than just the DOM and `@types/web` reflects that. If you're interested in switching to use `@types/web`, there are instructions in the [npm README](https://www.npmjs.com/package/@types/web).

- The 4.4 `dom.d.ts` file which is included with TypeScript includes quite a lot of changes, with a lot of APIs which have been marked as deleted by the web standards committees. You can see the full list [in this issue](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029#issuecomment-869224737)
  
   We've been testing these changes with DefinitelyTyped, and have been individually rolling back changes which we think are a bit too breaking, this is more of an art than a science however.  If you think we should rollback a particular change we're open to discussing change in [this issue](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029) or the [`dom-lib-generator` channel in the TypeScript Community Discord](https://discord.gg/typescript)

We plan to have thorough support for replacing `lib.dom.d.ts` with `@types/web` in 4.5 with https://github.com/microsoft/TypeScript/issues/44795

orta:
# High Level Changes from 4.3 to 4.4

## New interfaces

* `BlobEvent`
* `CSSAnimation`
* `CSSCounterStyleRule`
* `CSSTransition`
* `ClipboardItem`
* `FileSystem`
* `FileSystemDirectoryEntry`
* `FileSystemDirectoryReader`
* `FileSystemEntry`
* `FileSystemFileEntry`
* `FontFace`
* `FontFaceSet`
* `FontFaceSetLoadEvent`
[...]

PR #1102, opened 2021-08-14
Track typedefs and generics when resolving promises

saschanaz:
Fixes a clipboard type issue reported in #1029.


saschanaz:
LGTM

PR #1108, opened 2021-08-17
Run the deploy every night

orta:
This should mean that merged PRs via code-owners-self-merge trigger a deploy (eventually)


saschanaz (approved):
The `push:` thing now looks redundant, but the schedule thing looks good. LGTM

orta:
Agree, dropped it so that there's one 'canonical' way to deploy (and document)

Issue #1110, opened 2021-08-20
`npm i` installs older dependencies


saschanaz:
Because package.json is not being updated. This can be a hurdle for contributors.

orta:
Dependabot doesn't seem to be making PRs, https://github.com/microsoft/TypeScript-DOM-lib-generator/pulls?q=is%3Apr+author%3Aapp%2Fdependabot

ExE-Boss:
Refs <https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1173#issuecomment-932809285>:
> Unfortunately https://github.com/OSS-Docs-Tools/code-owner-self-merge does not support partial ownership of package.json 😬 <https://github.com/OSS-Docs-Tools/code-owner-self-merge/issues/29>

PR #1111, opened 2021-08-20
Update bcd to 4.0.0

saschanaz:
I guess #1110 is because the major version change? I'll close it if it still happens after this.


orta:
👍🏻

PR #1112, opened 2021-08-22
fix(41212,41334): removed duplicate interface extensions

neotan:
Fixes [#41212](https://github.com/microsoft/TypeScript/pull/41212) and [#41334](https://github.com/microsoft/TypeScript/issues/41334)

### DIFF
--- a/src/build/helpers.ts
+++ b/src/build/helpers.ts
@@ -325,3 +325,10 @@ export function followTypeReferences(
     }
   }
 }
+
+export function assertUnique(list: string[]): string[] {
+  if (new Set(list).size < list.length) {
+    throw new Error(`Duplicate items found in the list: ${list}`);
+  }
+  return list;
+}
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -6048,7 +6048,7 @@ interface HTMLElementEventMap extends ElementEventMap, DocumentAndElementEventHa
 }
 
 /** Any HTML element. Some elements directly implement this interface, while others implement it via an interface that inherits it. */
-interface HTMLElement extends Element, DocumentAndElementEventHandlers, ElementCSSInlineStyle, ElementCSSInlineStyle, ElementContentEditable, GlobalEventHandlers, HTMLOrSVGElement {
+interface HTMLElement extends Element, DocumentAndElementEventHandlers, ElementCSSInlineStyle, ElementContentEditable, GlobalEventHandlers, HTMLOrSVGElement {
     accessKey: string;
     readonly accessKeyLabel: string;
     autocapitalize: string;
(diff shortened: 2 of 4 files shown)


saschanaz (changes requested):
Oh, the real fix would be removing such duplicates from files like `addedTypes.jsonc`. Please remove `HTMLElement` and `SVGElement` from there instead.

But `uniq` seems still helpful with a different form, how about `assertUnique` that throws when there is a duplicate?

src/build/helpers.ts
```diff
@@ -325,3 +325,8 @@ export function followTypeReferences(
     }
   }
 }
+
+export function uniq(list: string[] = []): string[] {
```
saschanaz, line 329: `=[]` seems redundant when all callers pass a valid array.

src/build/helpers.ts
```diff
@@ -325,3 +325,21 @@ export function followTypeReferences(
     }
   }
 }
+
+export function assertUnique(list: string[]): string[] {
+  if (!Array.isArray(list)) return list;
```
saschanaz, line 330:
```suggestion
```

The type requirement `: string[]` is sufficient here (as this is just an internal function and thus is not exposed as a part of JavaScript library)

src/build/helpers.ts
neotan: Hi @saschanaz, 
Thanks for your comments! I've just started using TS these months, this is my first PR for TS ecosystem XD actually :P.
I updated the codes based on your comment, for the `assertUnique`, didn't use while/for/map for higher performance, but it's easier to understand.

Pls let me know if anything needs to be done. Thanks ;)

src/build/helpers.ts
```diff
@@ -325,3 +325,21 @@ export function followTypeReferences(
...
+    .forEach((str, _, trimmed) => {
+      if (trimmed.indexOf(str) !== trimmed.lastIndexOf(str)) result.add(str);
+    });
+
+  if (result.size) {
+    throw new Error(
+      `Duplicate items [${[...result].join(", ")}] found in the list.`
+    );
+  }
+  return list;
```
saschanaz, line 344: Good but IMO a bit overkill, checking whether it has a duplicate is enough IMO. A human should easily see the duplications in the list, given that callers are only expected to pass small lists.

```suggestion
  if (new Set(list).size < list.length) {
    throw new Error(`Duplicate items found in the list: ${list}`);
  }
  return list;
```

saschanaz:
Thank you, looks good 👍

@orta I think I'm often blocked by "First-time contributors need a maintainer to approve running workflows" thing, what do you think about https://github.com/OSS-Docs-Tools/code-owner-self-merge/issues/26 ?

orta:
( I agree. I can build that this week, I was thinking about giving access to closing PRs and maybe issues too 👍🏻 )

saschanaz:
> 
> 
> ( I agree. I can build that this week, I was thinking about giving access to closing PRs and maybe issues too 👍🏻 )
> 
> `<edit>`and then I saw you had already got that ball rolling - thanks

That's https://github.com/OSS-Docs-Tools/code-owner-self-merge/pull/25

saschanaz:
LGTM

PR #1113, opened 2021-08-23
Update HTMLMediaElement.preload type and comment

MartinJohns:
Fixes microsoft/TypeScript#45458

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -2792,6 +2792,16 @@
                         }
                     }
                 }
+            },
+            "HTMLMediaElement": {
+                "properties": {
+                    "property": {
+                        "preload": {
+                            "name": "preload",
+                            "overrideType": "\"none\" | \"metadata\" | \"auto\""
+                        }
+                    }
+                }
             }
         }
     },
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -2797,8 +2797,7 @@
                 "properties": {
                     "property": {
                         "preload": {
-                            "name": "preload",
-                            "overrideType": "\"none\" | \"metadata\" | \"auto\""
+                            "overrideType": "\"none\" | \"metadata\" | \"auto\" | \"\""
                         }
                     }
                 }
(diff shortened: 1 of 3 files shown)


MartinJohns:
This PR contains a **breaking change**, which is acceptable according to [Ryans comment](https://github.com/microsoft/TypeScript/issues/45458#issuecomment-900493474).

inputfiles/overridingTypes.jsonc
```diff
@@ -2792,6 +2792,16 @@
                         }
                     }
                 }
+            },
+            "HTMLMediaElement": {
+                "properties": {
+                    "property": {
+                        "preload": {
+                            "name": "preload",
```
saschanaz, line 2800:
```suggestion
```

"name" is not needed in this case.

inputfiles/overridingTypes.jsonc
```diff
@@ -2792,6 +2792,16 @@
...
+                            "overrideType": "\"none\" | \"metadata\" | \"auto\""
```
saschanaz, line 2801:
```suggestion
                            "overrideType": "\"none\" | \"metadata\" | \"auto\" | \"\""
```

Since empty string is also valid per both the MDN document and the spec: https://html.spec.whatwg.org/multipage/media.html#attr-media-preload

saschanaz (changes requested):

inputfiles/overridingTypes.jsonc
MartinJohns: Is it not? Can you elaborate when it is needed and when not? I've seen it plenty of times in the code. Or is the usage just deprecated (if name matches) and these are leftovers?

inputfiles/overridingTypes.jsonc
MartinJohns: Neither [Apple](https://developer.apple.com/documentation/webkitjs/htmlmediaelement/1633059-preload) nor [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement#properties) lists it, so I left it out. Next time I'll check the spec. :-)

saschanaz:
LGTM

PR #1116, opened 2021-08-23
Restore HTMLDocument constructor

### LINKED ISSUE #1115: Add `HTMLDocument` constructor back
Turned up an error in https://github.com/microsoft/TypeScript/issues/45540. https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/91408baa65a1e2fe6f51709bbc20175714774401 added the interface back (as deprecated), but not the value, which is still very much present in runtimes (`document.constructor === HTMLDocument`).

sandersn:
Detected in https://github.com/microsoft/TypeScript/issues/45540, where it breaks ant-design. `x instanceof HTMLDocument` no longer works.

This is likely worth fixing before the release of 4.4.

Fixes #1115

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -6044,6 +6044,12 @@ interface HTMLDocument extends Document {
     removeEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | EventListenerOptions): void;
 }
 
+/** @deprecated */
+declare var HTMLDocument: {
+    prototype: HTMLDocument;
+    new(): HTMLDocument;
+};
+
 interface HTMLElementEventMap extends ElementEventMap, DocumentAndElementEventHandlersEventMap, GlobalEventHandlersEventMap {
 }
 
@@ -16000,7 +16006,6 @@ interface WindowEventMap extends GlobalEventHandlersEventMap, WindowEventHandler
 
 /** A window containing a DOM document; the document property points to the DOM document loaded in that window. */
 interface Window extends EventTarget, AnimationFrameProvider, GlobalEventHandlers, WindowEventHandlers, WindowLocalStorage, WindowOrWorkerGlobalScope, WindowSessionStorage {
-    HTMLDocument: Document;
     /** @deprecated This is a legacy alias of `navigator`. */
     readonly clientInformation: Navigator;
     /** Returns true if the window has been closed, false otherwise. */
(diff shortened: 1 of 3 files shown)


sandersn:
@saschanaz You are probably interested in this too.

saschanaz:
Perhaps a good chance to add a small test here: https://github.com/microsoft/TypeScript-DOM-lib-generator/tree/main/unittests/files

baselines/dom.generated.d.ts
```diff
@@ -17035,7 +17040,6 @@ declare var Image: {
 declare var Option: {
     new(text?: string, value?: string, defaultSelected?: boolean, selected?: boolean): HTMLOptionElement;
 };
-declare var HTMLDocument: Document;
```
andrewbranch, line 17038: Ooh, I’m guessing this was an incorrect attempt at back compat, right? So until now the `HTMLDocument` global value appeared to actually be an _instance_ of an `HTMLDocument`?

andrewbranch (approved):

baselines/dom.generated.d.ts
orta: Not quite, the general pattern for a lot of DOM types is a `var` with an `interface` - as that's globally extendable and generally represents the runtimey flexiness of JS. This still represented the 'classlike' ness if you will, kind surprised that we didn't get a constructor though

orta (approved):
Green if it gets a test 👍🏻

baselines/dom.generated.d.ts
andrewbranch: That’s not what this says though... this var doesn’t have a constructor-ish type like the one added back by the PR, it has an instance-ish type. There’s no class-likeness here, which is why ant-design had an error.

unittests/files/htmldocument.ts
```diff
@@ -0,0 +1,2 @@
+const d = new HTMLDocument()
+const result = d instanceof HTMLDocument
```
saschanaz, line 2: `new HTMLDocument()` should ideally be prevented since it's illegal, I guess this is enough:

```suggestion
document instanceof HTMLDocument
```

baselines/dom.generated.d.ts
andrewbranch: I think I’d be ok with that approach, but personally don’t see a strong reason to remove `HTMLDocument` (or even to mark it deprecated)—I must be missing the original context.

saschanaz: `HTMLDocument` is not defined as a separate interface anymore by the HTML spec and served as a legacy alias of `Document`. But since both Firefox and Chrome haven't implemented things as the spec says, maybe this also make sense as-is.

saschanaz:
LGTM

saschanaz:
> Sorry @saschanaz, you don't have access to merge: /unittests/files/htmldocument.ts.

@orta could you add the directory to CODEOWNERS 👀

orta:
I can

PR #1122, opened 2021-08-27
Show change for all files

saschanaz:
Currently release notes are generated only for `.iterable`s, because of a regression from #1084 😬

This also removes hackish `\n` from title and makes sure things have double newlines between each other.


orta:
Nice, thanks!

PR #1134, opened 2021-09-04
Fix ReadableStream incorrectly marked as a sync iterable

MattiasBuelens:
It looks like a regression slipped into the `ReadableStream` interface. In TypeScript 4.4, it incorrectly has a `[Symbol.iterator]` method, along with automatic `forEach`/`entries`/`keys`/`values` methods. In reality [`ReadableStream` is an `async iterable<any>`](https://streams.spec.whatwg.org/#readablestream), meaning it has a `[Symbol.asyncIterator]` and `values` method (both returning an async iterator).

Since we don't (yet) have a way to express this, I've changed the emitter to instead ignore `async iterable` declarations, so at least we don't *add* any non-existent methods. Alternatively, we could do this properly: tackle microsoft/TypeScript#29867 and introduce `lib.dom.asynciterable.d.ts`.


saschanaz (approved):
It seems nobody really has implemented async iterable for ReadableStream, so lgtm.

saschanaz:
@github-actions, lgtm again?

PR #1135, opened 2021-09-05
Add async iterable flavors

MattiasBuelens:
Add support for `async iterable` declarations [per Web IDL](https://heycam.github.io/webidl/#idl-async-iterable), and emit them as `.asynciterable.d.ts` flavors. This is an alternative solution for #1134, and a first step towards microsoft/TypeScript#29867.

Since this is currently only used by a single interface (`ReadableStream`), I understand that there's limited use for this. That's why I'm opening this as a draft PR for now: as a proof-of-concept on what this *could* look like, and as a starting point to discuss *if* we want to proceed with this. (If we don't want this, that's fine too, no hard feelings. 🙂)

### DIFF
--- a/baselines/audioworklet.asynciterable.generated.d.ts
+++ b/baselines/audioworklet.asynciterable.generated.d.ts
@@ -1,0 +1,8 @@
+/////////////////////////////
+/// AudioWorklet Async Iterable APIs
+/////////////////////////////
+
+interface ReadableStream<R = any> {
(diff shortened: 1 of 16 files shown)


saschanaz:
Hmm, I'd accept this after anyone actually implement that. For now I guess no real value is there.

MattiasBuelens:
Uhhh, yeah, I just realized there are currently 0 browser implementations of `ReadableStream[Symbol.asyncIterator]`. Woops. 😅 

I'll try again next year or something. 😛

PR #1140, opened 2021-09-06
Use special type for MediaRecorder error event

### LINKED ISSUE #1138: Should MediaRecorderEventMap use MediaRecorderErrorEvent?
The new MediaRecorderEventMap currently maps the `error` event to a basic `Event` type:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/fa11d129b123258298af34059d8c4726d512fb80/baselines/dom.generated.d.ts#L9016

If I read the spec correctly ([error summary section](https://w3c.github.io/mediacapture-record/#event-summary)), the error event should use `MediaRecorderErrorEvent`.

Is there a reason for this or is this a bug?

frigus02:
The spec (https://w3c.github.io/mediacapture-record/#event-summary) specifies a specific event type MediaRecorderErrorEvent for the error event.

This ensures we're using it, similar to the dataavailable event.

Fixes #1138 

(I hope this is the correct way to do this 🙂)

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -9013,7 +9013,7 @@ declare var MediaQueryListEvent: {
 
 interface MediaRecorderEventMap {
     "dataavailable": BlobEvent;
-    "error": Event;
+    "error": MediaRecorderErrorEvent;
     "pause": Event;
     "resume": Event;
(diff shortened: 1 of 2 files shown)


saschanaz:
👍 LGTM

PR #1142, opened 2021-09-13
Exclude features with `version_added: preview`

saschanaz:
See https://github.com/mdn/browser-compat-data/issues/12344 for the reasoning.


orta:
Picking an arbitrary example, seems reasonable to me:

Screen Shot 2021-09-13 at 2 07 10 PM

orta (approved):

saschanaz:
LGTM

saschanaz:
They are added just two days ago, so shouldn't be too breaking: https://github.com/microsoft/TypeScript-DOM-lib-generator/releases/tag/%40types%2Fweb%400.0.31

orta:
great, that's not breaking IMO

PR #1144, opened 2021-09-13
Split out types and values

orta:
This _should_ allow something like Node / React Native to be able to write `/// <reference lib="dom.types" />` and have access to all of the global types, but none of the values. Re: https://github.com/microsoft/TypeScript/issues/43972

The concept is reasonably simple, we create two separate printers one focused on values and another on types. When we don't care to split they are the same printer and so no changes. Then the printer is DI'd down from whatever decides if it's printing a type or value.


saschanaz:
Hmm, as I said in Discord, I think this:

* doesn't solve the collision issue on the static part as we've seen in #1093
* forces including every type when consumers need only a part of the types

So my current idea is that we just duplicate the whole types into the module, while keeping the current lib as-is.

```ts
export interface AbortSignal {
  // ...
};

export var AbortSignal: {
  // ...
};
```

```ts
// on Node.js:
import { AbortSignal } from "typescript/dom";
```

Bashamega:
Hello @orta, @saschanaz 
Do you have any plans to update this PR and merge it?

orta:
No, not from me, so I'll close it 👍🏻

PR #1147, opened 2021-09-16
Use null for RequestInit.window

### LINKED ISSUE #1145: `RequestInit.window` should be null
Web IDL marks it as `any`, but [left a comment](https://fetch.spec.whatwg.org/#dom-requestinit-window) saying that it can only be `null`.

```webidl
dictionary RequestInit {
  // ...
  any window; // can only be set to null
}
```

And in the later [constructor steps](https://fetch.spec.whatwg.org/#dom-request), a `TypeError` is thrown if it is non-null.

> 10. If _init_["`window`"] exists and is non-null, then throw a `TypeError`.

Hence the type to be `null` in TS.

Thank you.

PaperStrike:
Fixes #1145


saschanaz:
Thanks!

LGTM

PR #1157, opened 2021-09-20
Update bim and refer to `legacyNamespace`

### LINKED ISSUE #1155: Add a way to override BCD path for certain types
For example, WebAssembly.compile exists on BCD as `javascript.builtins.WebAssembly.compile` but the code is checking `api.WebAssembly.compile` instead.

saschanaz:
Fixes #1155


saschanaz:
lgtm

PR #1160, opened 2021-09-27
Prepare for 4.5b

orta:
Doesn't look too drastic:

# @types/web

## New interfaces

* `ElementInternals`

## Removed interfaces

* `PaymentAddress` - [Classed as obsolete](https://caniuse.com/mdn-api_paymentaddress)

## Modified

* CSSStyleDeclaration
  * Added: `accentColor`
* Document
  * Removed: `elementFromPoint`, `elementsFromPoint` - these live on `DocumentOrShadowRoot` instead
* HTMLElement
  * Added: `attachInternals`
* HTMLSlotElement
  * Added: `assign`
* Navigator
  * Added: `canShare`
  * Removed: `pointerEnabled` - removed from [spec in 2014](https://webplatform.github.io/docs/dom/Navigator/pointerEnabled/)
* ReadableStream
  * Removed: `forEach` - Doesn't seem to be in [the Readable Stream API](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#methods)
* Text
  * Removed: `assignedSlot` - This might be something we backtrack during the beta, will be easy to re-add if we want. https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/d42f963a7fc80796e7922eaa1cd10df939627cca#r57102539

### Non-value types

* MediaTrackConstraintSet
  * Added: `autoGainControl`, `noiseSuppression`
* MediaTrackSettings
  * Added: `autoGainControl`, `noiseSuppression`
* MediaTrackSupportedConstraints
  * Added: `autoGainControl`, `noiseSuppression`
* DocumentOrShadowRoot
  * Added: `elementFromPoint`, `elementsFromPoint`

## `index.iterable.d.ts`

No changes


orta:
It looks like the migration process doesn't take into account that we have to change the abortcontroller - https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/c7f81377d3ee5157723ef8cb20ef873686b9fed8/deploy/createTypesPackages.js#L198-L204 which means anyone without context is going to keep breaking the migration into TS

orta:
I've added a button to generate a PR to TypeScript re-using @saschanaz's BCD PR work, as the rumor mill is that I got it wrong https://github.com/microsoft/TypeScript/pull/46080#issuecomment-932645573 and that's a good cause to automate

Might have to do a few pushes to main to get it right

orta:
There we go https://github.com/microsoft/TypeScript/pull/46222

.github/workflows/pr-to-typescript.yml
```diff
@@ -0,0 +1,39 @@
...
+      - run: npm run build
+      - run: npm test
+
+      - uses: actions/checkout@v2
+        with:
+          repository: "microsoft/TypeScript"
+          path: "TypeScript"
+
+      - run: |
+          node ./lib/migrate.js ./TypeScript
```
ExE-Boss, line 26:
```suggestion
          node ./deploy/migrate.js ./TypeScript
```

ExE-Boss (changes requested):

PR #1170, opened 2021-10-02
Use lelia/forker instead of gh cli

saschanaz:
I think `read:org` is only required for other features of `gh` and shouldn't be required for forking itself.


saschanaz:
@orta ^^^

orta:
Lets give it a shot!

PR #1173, opened 2021-10-02
Sync `package.json` with `package‑lock.json`

ExE-Boss:
This syncs `package.json` with `package‑lock.json`, ensuring that the dependency versions match CI, thus avoiding a build error like in [run 3702838165](https://github.com/microsoft/TypeScript-DOM-lib-generator/runs/3702838165?check_suite_focus=true), which happens when a package manager other than **npm** is used locally.


saschanaz:
Unfortunately https://github.com/OSS-Docs-Tools/code-owner-self-merge does not support partial ownership of package.json 😬

PR #1182, opened 2021-10-20
Use MessageEvent in EventSource#addEventListener

### LINKED ISSUE #736: EventSource events are always provided a MessageEvent
Right now, `EventSource.addEventListener` has the following `eventName => eventType` mapping:

```ts
interface EventSourceEventMap {
    "error": Event;
    "message": MessageEvent;
    "open": Event;
}
```

And all other events are given the `Event` type. However, *any* event can be given a `MessageEvent`, e.g. given the following EventSource response:

```
event: foobar
data: I am foo

event: open
data: Why hello there

event: error
data: Now isn't this fun
```

And the following code:

```js
var source = new EventSource('https://wherever');
source.addEventListener('foobar', console.log);
source.addEventListener('open', console.log);
source.addEventListener('error', console.log);
```

We'll get the following console output:

```
open { target: EventSource, isTrusted: true, srcElement: EventSource, … }

foobar { target: EventSource, isTrusted: true, data: "I am foo", … }

open { target: EventSource, isTrusted: true, data: "Why hello there", … }

error { target: EventSource, isTrusted: true, data: "Now isn't this fun", … }
```

The trivial solution would be to have the signature `addEventListener(type: string, listener: (this: EventSource, ev: MessageEvent) => any, ...)`

(omitted for brevity that the listener can also be an `EventListenerObject` receiving a `MessageEvent`).

I can send a PR if this seems like the right solution to you, I'm however a ts novice so maybe I'm missing a better solution?

Thanks.

Alexendoo:
Fixes #736

The event listeners are removed and then recreated to avoid having the
existing add/removeEventListeners in addition to the new ones

### DIFF
--- a/unittests/files/eventsource.ts
+++ b/unittests/files/eventsource.ts
@@ -1,4 +1,7 @@
 const source = new EventSource("/");
 
+source.addEventListener("custom", (event) => {
+  event.data; // MessageEvent specific
+});
 source.addEventListener("custom", (event: MessageEvent) => {});
 source.addEventListener("with-handle-event", { handleEvent(event: Event) {} });


Alexendoo:
Thanks, made that change. That is nicer indeed

Issue #1151, opened 2021-09-18
`Node.parentElement` should be `Element | null`, not `HTMLElement | null`


trusktr:
Hello! I'm new to this. How exactly would I change that?

For example, a `<g>` element somewhere inside an `<svg>` element will have a `.parentElement` that is an `SVGElement`, which is not an `HTMLElement`.

saschanaz:
This is true and was done by #885, but then reverted by #970. Seems there were tons of breakages. @orta

I doubt we can do this if that was the case, but it should at least need some comment about it.

trusktr:
I left some comments. Plus imagine when something useful like this comes out in the future:

https://github.com/WICG/webcomponents/issues/942

The only proper type for these DOM APIs is `Element`.

HolgerJeromin:
@trusktr 
Technically you are right in moving to `Element`. But this would break many applications. Most do only deal with HTML and would need manual adjustments. 
Developers should be happy and not annoyed by forcing them to touch all DOM touching code with compiled for years.
SVG users are right now forced to type assert anyway.

ref https://github.com/microsoft/TypeScript/pull/42067#issuecomment-754860862

orta:
Yeah, this has been well covered  and discussed. I don't think I've seen anything new in the comments to warrant keeping this issue around. If you'd like you, can fork this repo to ship your own dom.d.ts build with this in 4.5, but it's unlikely to come in by default given the trade-offs.

PR #1198, opened 2021-11-10
Adjust `.parentElement` type to include `SVGElement`

Andarist:
In reality, this also includes `| null` but this has been removed in https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/2cd8ad3974b63affd454164028960821e77a2e17 so I guess you want to omit that for pragmatic reasons.

Originally this was typed as `Element | null` (and that probably was the most "correct" type) but that was changed to `HTMLELement | null` here: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/560 . However, `SVGElement` can appear here in a lot of scenarios. 

An HTML element can be a parent of SVG (the simplest `<div><svg/></div>`), SVG element can be a parent of another SVG element (`<svg><rect/></svg>`) but also the SVG element can be a parent of an HTML element (`<svg><foreignObject><div></div></foreignObject></svg>`)

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -766,7 +766,7 @@
                             "overrideType": "ParentNode"
                         },
                         "parentElement": {
-                            "overrideType": "HTMLElement"
+                            "overrideType": "HTMLElement | SVGElement"
                         },
                         "childNodes": {
                             "overrideType": "NodeListOf<ChildNode>"


HolgerJeromin:
Build has failed because you missed to include the generated files.

But did you considered the feedback from #1151 ?

Andarist:
Sorry, I've missed that discussion. I see the points raised there but having to cast this through unknown/any first and then to `SVGElement` doesn't seem like a good DX. I understand that people might deal with SVG less often than with HTML but there should be a clear way of working with SVGs too. Not everyone might figure out how to "properly" cast this as it's not intuitive.

It also seems that there is no good list of "expectations" about the DOM types. A lot of this is gray area and the APIs don't provide full type safety. It would be great if things like this would be documented somewhere.

orta:
Happy to have a FAQ of some sort, maybe this could be the first one?

orta:
In here I think, TypeScript's FAQ is already very very long

PR #1203, opened 2021-11-13
Use actions/setup-node@2 with `lts/*`

saschanaz:
`@2` has been there for a year, and `cache` is actually not supported in `@1`.


orta:
👍🏻

PR #1206, opened 2021-11-23
Add more specific types for globalCompositeOperation

dhaig:
Add types for `CanvasRenderingContext2D.globalCompositeOperation`, which is currently a `string` but should only accept certain values. The acceptable values come from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation).

Reported as a bug in TypeScript's main repo: https://github.com/microsoft/TypeScript/issues/34673

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -280,6 +280,37 @@
                 "value": [
                     "inprogress"
                 ]
+            },
+            "GlobalCompositeOperation": {
+                "name": "GlobalCompositeOperation",
+                "value": [
+                    "source-over",
+                    "source-in",
+                    "source-out",
+                    "source-atop",
+                    "destination-over",
+                    "destination-in",
+                    "destination-out",
+                    "destination-atop",
+                    "lighter",
+                    "copy",
+                    "xor",
+                    "multiply",
+                    "screen",
+                    "overlay",
+                    "darken",
+                    "lighten",
+                    "color-dodge",
+                    "color-burn",
+                    "hard-light",
+                    "soft-light",
+                    "difference",
+                    "exclusion",
+                    "hue",
+                    "saturation",
+                    "color",
+                    "luminosity"          
+                ]
             }
         }
     },
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3245,7 +3245,7 @@ declare var CacheStorage: {
 
 interface CanvasCompositing {
     globalAlpha: number;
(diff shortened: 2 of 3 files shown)


inputfiles/addedTypes.jsonc
```diff
@@ -280,6 +280,37 @@
...
+                    "color-dodge",
+                    "color-burn",
+                    "hard-light",
+                    "soft-light",
+                    "difference",
+                    "exclusion",
+                    "hue",
+                    "saturation",
+                    "color",
+                    "luminosity"
```
saschanaz, line 312: I strongly suspect MDN is wrong, [the actual spec](https://drafts.fxtf.org/compositing/#canvascompositingandblending) has much shorter list:

><composite-mode> = clear | copy | source-over | destination-over | source-in |    
destination-in | source-out | destination-out | source-atop |    
destination-atop | xor | lighter | plus-darker | plus-lighter

It's also needed to check whether each item here is actually supported on browsers.

saschanaz: Oh sorry, the spec actually defines more items when including `<blend-mode>`. Checking implementations...

dhaig: The spec is confusing for this property. The extra enum items come from `<blend-mode>` but that also includes `normal` which browsers don't seem to support and which MDN doesn't list. I didn't include `normal` even though the spec technically would, I think that's in line with TypeScript's philosophy? Either way, the supported value `source-over` is what most people would assume `normal` does.

saschanaz: Gecko: https://searchfox.org/mozilla-central/rev/9a5f36b0ddb9cb8ae556fc5b45f8ccea0f0da6f8/dom/canvas/CanvasRenderingContext2D.cpp#4756-4782
WebKit: https://webkit-search.igalia.com/webkit/rev/13bfcede837af469200178969bd5c08bc0bc4518/Source/WebCore/platform/graphics/GraphicsTypes.cpp#38-74
Blink: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/graphics/graphics_types.cc;l=34-51;drc=de68be3f18ba99cc01d75903e167ca09bade253c

Seems all the values there are supported!

saschanaz (approved):
I wonder there should be two separate enums `BlendMode` and `CompositeMode` but for now, lgtm.

saschanaz:
lgtm, @github-actions?

PR #1209, opened 2021-11-30
remove unnecessary type dependency

joematune:
When installing dependencies, the following warning comes from `@types/node-fetch`
```bash
npm WARN deprecated @types/node-fetch@3.0.3: This is a stub types definition. node-fetch provides its own type definitions, so you do not need this installed.
```

Included types for `node-fetch` can be seen in the node fetch repository - https://github.com/node-fetch/node-fetch/tree/main/%40types

The `node-fetch` npm package also shows that type declarations are included - https://www.npmjs.com/package/node-fetch

Screen Shot 2021-11-30 at 6 34 10 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "@octokit/rest": "^18.12.0",
     "@types/jsdom": "^16.2.13",
     "@types/node": "^16.10.5",
-    "@types/node-fetch": "^3.0.2",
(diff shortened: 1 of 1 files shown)


orta:
Ace 👍🏻

saschanaz:
Yay node-fetch!

PR #1220, opened 2021-12-03
Improve loading attribute for HTMLImageElement

1000ch:
[`loading` attribute for `` element is going to be supported also by Safari](https://webkit.org/blog/12040/release-notes-for-safari-technology-preview-135/), [HTML element: img: loading](https://caniuse.com/mdn-html_elements_img_loading).

- Add a description of [`loading` attribuets for `` element](https://html.spec.whatwg.org/#the-img-element).
- Use [`'eager' | 'lazy'` instead of `string` as a type for `loading` attributes](https://html.spec.whatwg.org/#lazy-loading-attributes)

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -356,7 +356,6 @@
                             "overrideType": "\"async\" | \"sync\" | \"auto\""
                         },
                         "loading": {
-                            "name": "loading",
                             "overrideType": "\"eager\" | \"lazy\""
                         }
                     }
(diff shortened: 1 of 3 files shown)


inputfiles/addedTypes.jsonc
```diff
@@ -354,6 +354,10 @@
                         "decoding": {
                             "name": "decoding",
                             "overrideType": "\"async\" | \"sync\" | \"auto\""
+                        },
+                        "loading": {
+                            "name": "loading",
```
saschanaz, line 359:
```suggestion
```

`name` is actually not needed here, same for above items (you don't need to fix them here)

saschanaz (commented):
The title is misleading, it's not actually adding an attribute. Maybe change it to `Improve`?

saschanaz (approved):
Lgtm

saschanaz:
LGTM?

PR #1246, opened 2022-01-14
 DOM lib: Add support for Trusted Types API

tosmolka:
This PR adds support for Trusted Types APIs and Sinks in dom and worker libraries. These APIs are so far only supported by Blink. In https://github.com/microsoft/TypeScript/issues/30024 we seem to be getting an agreement to include this change anyway.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -8727,11 +8727,6 @@ interface InnerHTML {
     innerHTML: string | TrustedHTML;
 }
 
-declare var InnerHTML: {
-    prototype: InnerHTML;
-    new(): InnerHTML;
-};
-
 interface InputDeviceInfo extends MediaDeviceInfo {
 }
 
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3447,13 +3447,13 @@ interface ChildNode extends Node {
      *
      * Throws a "HierarchyRequestError" DOMException if the constraints of the node tree are violated.
      */
-    after(...nodes: (Node | string)[]): void;
+    after(...nodes: (Node | string | TrustedScript)[]): void;
     /**
      * Inserts nodes just before node, while replacing strings in nodes with equivalent Text nodes.
      *
      * Throws a "HierarchyRequestError" DOMException if the constraints of the node tree are violated.
      */
-    before(...nodes: (Node | string)[]): void;
+    before(...nodes: (Node | string | TrustedScript)[]): void;
     /** Removes node. */
     remove(): void;
     /**
@@ -3461,7 +3461,7 @@ interface ChildNode extends Node {
      *
      * Throws a "HierarchyRequestError" DOMException if the constraints of the node tree are violated.
      */
-    replaceWith(...nodes: (Node | string)[]): void;
+    replaceWith(...nodes: (Node | string | TrustedScript)[]): void;
 }
 
 /** @deprecated */
@@ -3843,7 +3843,7 @@ interface DOMParser {
      *
      * Values other than the above for type will cause a TypeError exception to be thrown.
      */
-    parseFromString(string: string, type: DOMParserSupportedType): Document;
+    parseFromString(string: string | TrustedHTML, type: DOMParserSupportedType): Document;
 }
 
 declare var DOMParser: {
@@ -4459,7 +4459,7 @@ interface Document extends Node, DocumentAndElementEventHandlers, DocumentOrShad
      * @param value Value to assign.
      * @deprecated
      */
-    execCommand(commandId: string, showUI?: boolean, value?: string): boolean;
+    execCommand(commandId: string, showUI?: boolean, value?: string | TrustedHTML): boolean;
     /** Stops document's fullscreen element from being displayed fullscreen and resolves promise when done. */
     exitFullscreen(): Promise<void>;
     exitPictureInPicture(): Promise<void>;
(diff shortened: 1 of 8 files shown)


saschanaz:
This needs multiple implementor interest.

shicks:
My understanding is that there are polyfill options for other browsers, making appropriate typings still pretty important.

orta:
You're welcome to use DefinitelyTyped if you'd like to ship types which only work on a single engine 👍🏻

shicks:
Either way, this needs to take advantage of the new differently-typed getter/setter feature in order to be properly backwards-compatible.  For example,

```
interface Element {
  get outerHTML(): string;
  set outerHTML(html: string|TrustedHTML);
}
```

tosmolka:
@shicks , can you please take another look? Do you think this would be backwards-compatible? I ended up introducing `setterType` as I could not find a way to do what we wanted with current implementation of emitProperty. Happy to choose different approach if needed.

shicks:
This looks much better.  IIUC, the TypeScript team has a corpus of real world code that they can compile against - presumably they'd want to test this against that corpus to make sure it doesn't break anything significant.

The main breakage I anticipate at this point is if anybody is _implementing_ the DOM API (i.e. like a NodeJS virtual DOM) - but at this point anybody who's just _using_ the DOM should be fine.

lgarron:
We're prototyping the use of `TrustedHTML` at GitHub, and would really love to see this land!

baselines/dom.generated.d.ts
```diff
@@ -14190,6 +14209,66 @@ declare var TreeWalker: {
     new(): TreeWalker;
 };
 
+interface TrustedHTML {
+    toJSON(): string;
+    toString(): string;
+}
```
DanielRosenwasser, line 14215: Can you give me some more insight into TrustedHTML nodes? Is it really this simple of an interface, or is there something "more" to them?

The reason I ask is that in a structural language like TypeScript, you could accidentally assign lots of stuff to `TrustedHTML`. Is there some internal runtime bit set, or is this really anything with a `toJSON` and a `toString`?

lgarron: `TrustedHTML` is interesting, in that it's both an interface and also the name of a global object.

So it's sort of like a class (you can do `value instanceof TrustedHTML`), but the main point of the API is that it *can't* be constructed without the Trusted Types API. Sort of like you have to do `document.createElement("div")` instead of `new HTMLDivElement()`.

Each TrustedHTML instance is also associated with a specific string policy. So conceptually it might make sense to have something like `TrustedHTML<"dompurify">`. I don't know if it makes sense to include that in TypeScript, though. The `"dompurify"` policy in that case is something that would be enforced by CSP for e.g. `innerHTML` assignment, which seems a bit outside the scope of type checking.

https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML#methods

weswigham: Maybe consider emitting a class with a `private` constructor instead of an interface with these very-widespread-methods then? That'd be allowable in `instanceof` and a private member should enable nominal checking (and thus forbid structural matches).

lgarron: How would that interact with something like https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60417 ?

Element classes like `HTMLDivElement` are declared using the same `interface`+`var` pattern as this PR. Is there a particular reason to diverge from that entirely?

For what it's worth, `new HTMLDivElement()` also doesn't error in TypeScript today.

lgarron:
Is there any chance of moving this forward in the near future?

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/61160 was making some progress last month, but my understanding is that it will still be hard to use if the `Element.innerHTML` setter still only allows `string` assignment in `lib.dom.d.ts`. This makes it rather tricky to use trusted types in practice, without working around lots of errors. 😔

saschanaz:
So far no implementer interest outside of Google, so a progress can happen when something happens there.

lgarron:
> So far no implementer interest outside of Google, so a progress can happen when something happens there.

🙋 We are actively working on exploring trusted types on the `github.com` site, and are specifically running into challenges with the concrete implementation due to this.

saschanaz:
>By implementer, I'm guessing you might mean just browsers, though?

Yes.



> Is there any way to make the `innerHTML` setter type at least _compatible_ with `TrustedHTML` from `@types/trusted-types`? That would make a big difference.

That's really a Microsoft decision, but I think TS should support overloading the setter to make it compatible.

koto:
> So far no implementer interest outside of Google, so a progress can happen when something happens there.

There is now multi implementor interest: https://github.com/mozilla/standards-positions/issues/20#issuecomment-1853427823, with a Gecko implementation [progressing](https://bugzilla.mozilla.org/show_bug.cgi?id=1508286). Implementation in WebKit is also [ongoing](https://bugs.webkit.org/show_bug.cgi?id=266630). What are the next best steps?

saschanaz:
When the feature ships, MDN data will be updated and that will be automatically reflected here.

koto:
> When the feature ships, MDN data will be updated and that will be automatically reflected here.

Do you mean browser compat data on MDN? I think browser compat data on MDN is manually updated, at least [that was the case](https://github.com/mdn/browser-compat-data/pull/6263) when the Trusted Types shipped in Chrome. Once that happens (which, I guess, is after shipping in Gecko or WebKit), what is actually happening automatically in this repo? Trusted Types APIs are already largely integrated in HTML, DOM and other specs (e.g. https://dom.spec.whatwg.org/#parentnode). 

Sorry for those basic questions, I'm just trying to figure out what needs manual work and when is it best to do that work. It seems like the 'when' part is after 2nd implementation part ships, but I don't know yet the 'what' part.

HolgerJeromin:
> Do you mean browser compat data on MDN? I think browser compat data on MDN is manually updated, at least that was the case when the Trusted Types shipped in Chrome. 

Yes. After that a friendly bot will come along and create here a PR like this:  #1726

PR #1251, opened 2022-01-26
Add discriminants to FileSystemHandle subtypes

sandersn:
This makes them almost as usable as the existing shims at @types/wicg-file-system-access.

To make up the rest of the difference, I tried adding an alias

```ts
type X = FileSystemFileHandle | FileSystemDirectoryHandle
```

But
1. I couldn't come up with a good name for X
2. I couldn't easily figure out how to generate the type alias without it being used somewhere.

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -1060,6 +1060,30 @@
                         }
                     ]
                 }
+            },
+            "FileSystemFileHandle": {
+                "name": "FileSystemFileHandle",
+                "properties": {
+                    "property": {
+                        "kind": {
+                            "name": "kind",
+                            "readonly": true,
+                            "overrideType": "\"file\""
+                        }
+                    }
+                }
+            },
+            "FileSystemDirectoryHandle": {
+                "name": "FileSystemDirectoryHandle",
+                "properties": {
+                    "property": {
+                        "kind": {
+                            "name": "kind",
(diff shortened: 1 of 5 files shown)


inputfiles/addedTypes.jsonc
```diff
@@ -1060,6 +1060,30 @@
                         }
                     ]
                 }
+            },
+            "FileSystemFileHandle": {
+                "name": "FileSystemFileHandle",
```
saschanaz, line 1065: `name` field is redundant, it's only needed when this is a new addition.

inputfiles/addedTypes.jsonc
```diff
@@ -1060,6 +1060,30 @@
...
+                "properties": {
+                    "property": {
+                        "kind": {
+                            "name": "kind",
```
saschanaz, line 1069: Same here.

saschanaz (approved):
LGTM but with nits:

saschanaz:
Oh no I didn't intend the merge, but I guess it shouldn't harm 😅

saschanaz:
@orta could the command be more explicit, like `@github-actions merge`?

sandersn:
I'll follow up with fixes for the nits.

sandersn:
https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1252

PR #1252, opened 2022-01-26
Make FileSystem*Handle additions smaller

sandersn:
`name` is not needed when the interface is already defined.

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -1062,7 +1062,6 @@
                 }
             },
             "FileSystemFileHandle": {
-                "name": "FileSystemFileHandle",
                 "properties": {
                     "property": {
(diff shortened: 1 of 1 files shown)


inputfiles/addedTypes.jsonc
```diff
@@ -1062,7 +1062,6 @@
                 }
             },
             "FileSystemFileHandle": {
-                "name": "FileSystemFileHandle",
                 "properties": {
                     "property": {
                         "kind": {
```
saschanaz, line 1067: (Same for the properties)

saschanaz: Oh those are actually new, so they are needed, okay.

saschanaz (approved):
LGTMindeed this time

PR #1259, opened 2022-02-02
Restore HTMLDialogElement exposure

saschanaz:
#1258 restored the interface, and this restores the global object.


saschanaz:
lgtm

PR #1263, opened 2022-02-15
Allow use of readonly number arrays with WebGL methods.

jameswilddev:
This PR changes the `Float32List` type to accept readonly arrays as well as mutable arrays.

The `Float32List` type is used as an argument to the following methods (all of which only read from the value, they do not mutate its contents):

- `clearBufferfv`
- `uniformMatrix2x3fv`
- `uniformMatrix2x4fv`
- `uniformMatrix3x2fv`
- `uniformMatrix3x4fv`
- `uniformMatrix4x2fv`
- `uniformMatrix4x3fv`
- `uniform1fv`
- `uniform2fv`
- `uniform3fv`
- `uniform4fv`
- `uniformMatrix2fv`
- `uniformMatrix3fv`
- `uniformMatrix4fv`
- `vertexAttrib1fv`
- `vertexAttrib2fv`
- `vertexAttrib3fv`
- `vertexAttrib4fv`
- `uniform1fv`
- `uniform2fv`
- `uniform3fv`
- `uniform4fv`
- `uniformMatrix2fv`
- `uniformMatrix3fv`
- `uniformMatrix4fv`


saschanaz:
This also sounds fair but I wonder how libraries are currently using them. Perhaps this one is *right* enough to cause some breaking change? Pinging @orta.

PR #1266, opened 2022-02-17
Override `Blob.stream()` return type to `ReadableStream<Uint8Array>`

yume-chan:
The type is defined at https://w3c.github.io/FileAPI/#blob-get-stream

> [...]
> 
> 2. Let _chunk_ be a new Uint8Array wrapping an ArrayBuffer containing _bytes_. If creating the ArrayBuffer throws an exception, then error _stream_ with that exception and abort these steps.
> 3. Enqueue _chunk_ in stream.


saschanaz (approved):
lgtm

saschanaz:
lgtm

PR #1267, opened 2022-02-19
Add DOMContentLoaded event to Window and Document

LachlanArthur:
The DOMContentLoaded event was missing from both the Document and Window interfaces.

https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event
https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -439,6 +439,10 @@
                         {
                             "name": "gamepaddisconnected",
                             "type": "GamepadEvent"
+                        },
+                        {
+                            "name": "DOMContentLoaded",
+                            "type": "Event"
                         }
                     ]
                 }
@@ -679,6 +683,14 @@
                             "overrideType": "null"
                         }
                     }
+                },
+                "events": {
+                    "event": [
+                        {
+                            "name": "DOMContentLoaded",
+                            "type": "Event"
+                        }
+                    ]
                 }
             },
             // This is used in the React d.ts files, and not including
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4196,6 +4196,7 @@ declare var DeviceOrientationEvent: {
 };
 
 interface DocumentEventMap extends DocumentAndElementEventHandlersEventMap, GlobalEventHandlersEventMap {
+    "DOMContentLoaded": Event;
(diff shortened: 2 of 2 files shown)


inputfiles/addedTypes.jsonc
```diff
@@ -679,6 +683,14 @@
                             "overrideType": "null"
                         }
                     }
+                },
+                "events": {
+                    "event": [
+                        {
+                            "name": "DOMContentLoaded",
```
saschanaz, line 690: Could you move this to `WindowEventHandlers`, thanks!

LachlanArthur: Moving the event from `Document` to `WindowEventHandlers` will remove it from document and add it to `<body>`, `<frameset>`, and `<svg>`, which I believe is incorrect.

The `DOMContentLoaded` event target is always the document object, so the only objects that will ever receive it are document and window.

saschanaz: Oh, I thought this was a Window one.

>The DOMContentLoaded event target is always the document object, so the only objects that will ever receive it are document and window.

Makes sense.

But then I wonder why there are `Event` and `Events` redundancy now...

LachlanArthur: From what I can tell reading the build code:

1. The `Event` interface is now getting included in the `webidl` variable because it's being referenced.
2. It then gets included in the `distinctETypeList` variable because its name ends with the string "Event".

I don't know if this is expected behaviour or a bug, but the resulting definitions aren't exactly _wrong_, as running `document.createEvent( "Event" )` and `document.createEvent( "Events" )` in a browser works fine.

saschanaz: It's not wrong, right, but I don't see anything that includes "Events" in the code nor in the webref dependency 🤔

I'd like to investigate before merging this...

LachlanArthur: The plural is being generated by the function `emitCreateEventOverloads`, which contains a list of events that have plurals.

saschanaz: Ah right. That function generates `createEvent` signature based on `type` and there currently is no `"type": "Event"`, so the addition here.

saschanaz:
lgtm

PR #1268, opened 2022-02-22
#1261: NodeListOf#item

### LINKED ISSUE #1261: NodeListOf#item misses return type "null"
Interface `NodeList` is correctly typed
```typescript
interface NodeList {
    item(index: number): Node | null;
}
```
but interface `NodeListOf` is missing the `null` type:
```typescript
interface NodeListOf<TNode extends Node> extends NodeList {
    item(index: number): TNode;
}
```
therefore TSC does not warn about missing null checks:
```typescript
const node = document.childNodes.item(999999).firstChild;
```

danielrentz:
fixes #1261
NodeListOf#item() misses `null` in return type

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -9932,7 +9932,7 @@ declare var NodeList: {
 };
 
 interface NodeListOf<TNode extends Node> extends NodeList {
-    item(index: number): TNode;
+    item(index: number): TNode | null;
     /**
      * Performs the specified action for each node in an list.
      * @param callbackfn  A function that accepts up to three arguments. forEach calls the callbackfn function one time for each element in the list.


HolgerJeromin:
Thanks for helping.
You manipulated a generated file. 
The [correct place](https://github.com/Microsoft/TSJS-lib-generator#input-files) would be `overridingTypes.jsonc`

danielrentz:
Oh I see, thanks for the pointer. Will take care.

saschanaz:
I think this is intentionally done to follow the current array indexing behavior. Also pinging @orta

danielrentz:
I understand that this will not be changed for array index due to the generic behaviour of TS. But it should be done "right" for explicit methods. At least it should be *consistent* across the interfaces `NodeList` and `NodeListOf`. BTW, how will the new method `Array#at` behave in TS?

orta:
Hrm, yeah, looks like I was a bit gung ho in https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1261

Yeah, `.at` will/should/does (I've not checked, but that's what we agreed on) have the undefined added in the types but we shouldn't really be switching such a popular DOM API I think (just like we didn't do it to arrays) - those sort of changes aren't really possible at the ecosystem size TS is at nowdays

Issue #1016, opened 2021-06-03
type RequestInfo should include URL


aslilac:
Currently `RequestInfo` is `Request | string`, but should (minimally) be `Request | URL | string`.

`fetch(new URL('/some_api', 'https://example.com'))` is valid, and works as you'd expect.

Technically it should be `Request | { toString(): string } | string` according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch), but I'm not sure if there are intricacies around that.

Paril:
I think the intricacy for the latter is that it's not very type-safe and more error-prone. They use the same thing for dataset/css style values, as well as other writeables, even if the spec allows and/or encourages it (for instance writing to `window.location` can be done with a string, and is most commonly done as such, but currently it's only writeable with a URL)

partheseas:
Yeah, it feels like `{ toString(): string }` falls into almost all of the same pits as `{}` in that almost anything could be cast to it.

orta:
Agree, I think we probably don't want this

orta:
Related: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1003

dnalborczyk:
> Agree, I think we probably don't want this

I would think _we do want_ this.

https://developer.mozilla.org/en-US/docs/Web/API/fetch#parameters

> A string or any other object with a stringifier — including a URL object — that provides the URL of the resource you want to fetch.

dnalborczyk:
speaking of which, this is pretty much the same problem, but has been added:

https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/331
https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/332

dnalborczyk:
and this as well: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/600

`fetch` falls pretty much into the same category.

PR #1269, opened 2022-02-24
Add URL to RequestInfo typedef

LiviaMedeiros:
Fixes: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1016

In most methods urlstring params are declared as `string | URL`, which ensures perfect balance between safe code and implicit casts allowed for `USVString`.

This PR aligns `RequestInfo` with them.
In particular, it allows to pass `URL` objects in `fetch`.


saschanaz:
This sounds fair, but I wonder this could be breaking as some other libraries may be using `RequestInfo` but not the implicit URL stringification behavior. @orta, thoughts?

orta:
Yeah, I trust past me to have done enough checking to make the call that we shouldn't do this last year, https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1016#issuecomment-869889376 (likely because of ecosystem breakage) and I don't think anything's changed sinced then

LiviaMedeiros:
Please clarify: you're against allowing any random objects like `{ toString(): string }` from the issue (or even anything stringable, as `USVString`-related specs says), which is obviously not typesafe; or against `URL` instances in particular, which are allowed everywhere in TypeScript if param name in docs matches `/url/i`?

saschanaz:
Accepting URL for Web APIs is okay, what's not okay is to change the typedef and thus affect third-parties. I think changing the signatures of the APIs using `RequestInfo` is better.

src/build/emitter.ts
```diff
@@ -672,7 +672,10 @@ export function emitWebIdl(
   function paramsToString(ps: Browser.Param[]) {
     function paramToString(p: Browser.Param) {
       p = resolvePromise(p);
-      if (p.name.toLowerCase().includes("url") && p.type === "USVString") {
+      if (
+        (p.name.toLowerCase().includes("url") && p.type === "USVString") ||
+        p.type === "RequestInfo"
```
saschanaz, line 677: This is now a little bit long to be in a if clause, could you add a function for this and call it here?

LiviaMedeiros: Sure.

saschanaz (approved):
Looks good. Orta, wdyt?

(I'd prefer something like `acceptsUrl` btw)

orta (approved):
Yeah, I agree, this is a good solution - good creative thinking folks

baselines/dom.generated.d.ts
```diff
@@ -3232,13 +3232,13 @@ declare var CSSTransition: {
  * Available only in secure contexts.
  */
 interface Cache {
-    add(request: RequestInfo): Promise<void>;
+    add(request: RequestInfo | URL): Promise<void>;
     addAll(requests: RequestInfo[]): Promise<void>;
```
saschanaz, line 3236: Not sure how to treat this one 🤔 but okay for now.

LiviaMedeiros: I'm not sure how to properly implement this in generator either.
But in my opinion, even something like `RequestInfo[] | URL[]` would be an okay trade-off here: mixing those in userland code should be rarely needed.

Another option is to refactor this later by introducing some `urlstring` and `RequestInfoOrUrl` typedefs. Dunno about downsides.

These should be affected as well:
https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/d060ef578ae9193a7606833261196977e446731b/baselines/dom.iterable.generated.d.ts#L25-L27

saschanaz: Hmmmmmm. Let's merge this and tackle this in a separate patch.

saschanaz:
lgtm

PR #1280, opened 2022-03-02
Convert `performance.measure()` to overloads

### LINKED ISSUE #1239: performance.measure() overloads instead of optional arguments
Hey there

according to [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure#syntax), the `performance.measure()` fn has 4 different argument sets (overloads) - the current typing suggests only one and marks some of the args as optional.

Perhaps it'd be better to have the 4 overloads listed explicitly? i.e. instead of:
```
measure(measureName: string, startOrMeasureOptions?: string | PerformanceMeasureOptions, endMark?: string): PerformanceMeasure;
```

have sth like:
```
measure(measureName: string): PerformanceMeasure
measure(measureName: string, MeasureOptions: PerformanceMeasureOptions): PerformanceMeasure
measure(measureName: string, startMark: string): PerformanceMeasure
measure(measureName: string, startMark?: string, endMark: string): PerformanceMeasure
```

WDYT?
If it's OK for you - what would be the best way to submit these changes?

yume-chan:
Fixes #1239

Technically the third one includes the first one (when both arguments are omitted), but if the first one is removed I feel like it's hard to discover it.

The `measureOptions` can't be used with `endMark` is defined at https://w3c.github.io/user-timing/#dom-performance-measure:

> 1. If _startOrMeasureOptions_ is a non-[empty](https://infra.spec.whatwg.org/#map-is-empty) [`PerformanceMeasureOptions`](https://w3c.github.io/user-timing/#dom-performancemeasureoptions) object, run the following checks:
>     1. If _endMark_ is given, [throw](https://webidl.spec.whatwg.org/#dfn-throw) a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -10275,7 +10275,9 @@ interface Performance extends EventTarget {
     getEntriesByName(name: string, type?: string): PerformanceEntryList;
     getEntriesByType(type: string): PerformanceEntryList;
     mark(markName: string, markOptions?: PerformanceMarkOptions): PerformanceMark;
-    measure(measureName: string, startOrMeasureOptions?: string | PerformanceMeasureOptions, endMark?: string): PerformanceMeasure;
+    measure(measureName: string): PerformanceMeasure;
+    measure(measureName: string, measureOptions: PerformanceMeasureOptions): PerformanceMeasure;
+    measure(measureName: string, startMark?: string, endMark?: string): PerformanceMeasure;
     now(): DOMHighResTimeStamp;
     setResourceTimingBufferSize(maxSize: number): void;
     toJSON(): any;
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -2323,7 +2323,9 @@ interface Performance extends EventTarget {
     getEntriesByName(name: string, type?: string): PerformanceEntryList;
     getEntriesByType(type: string): PerformanceEntryList;
     mark(markName: string, markOptions?: PerformanceMarkOptions): PerformanceMark;
-    measure(measureName: string, startOrMeasureOptions?: string | PerformanceMeasureOptions, endMark?: string): PerformanceMeasure;
+    measure(measureName: string): PerformanceMeasure;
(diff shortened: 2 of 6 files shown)


saschanaz:
Sounds like the spec should be changed too 🤔

But for TypeScript I think it prefers unions over overloads to pass union arguments. @orta, has there been any changes over union arguments?

```ts
declare function foo(arg: number): any;
declare function foo(arg: string): any;

let arg: number | string;
foo(arg); // This fails 😬
```

yume-chan:
In this case the two overloads has different argument length. `performance.measure('name', { start: 'start' }, 'end')` is allowed in current type definition but will result a runtime error.

The WebIDL spec reads: https://webidl.spec.whatwg.org/#idl-overloading-vs-union

> When the operation has significantly different semantics for different argument types or lengths, [overloading](https://webidl.spec.whatwg.org/#dfn-overloaded) is preferred.

In this case maybe the difference is not "significant".

But the spec continues:

> Even though the IDL is shorter in the second version _[when using optional arguments]_, two distinctively different concepts are conflated in the first argument _[`startOrMeasureOptions`]_. Without [overloads](https://webidl.spec.whatwg.org/#dfn-overloaded), the question "is _[startMark]_ or _[options]_ paired with _[endMark]?"_ is much more difficult to answer without reading the [method steps](https://webidl.spec.whatwg.org/#method-steps) of the operation. This makes the second version remarkably less readable than the first.

IMO this case can fall in this situation.

I will try to make a pr upstream, but please leave this pr open.

yume-chan:
I merged first two overloads like `createImageBitmap` does:

```ts
    createImageBitmap(image: ImageBitmapSource, options?: ImageBitmapOptions): Promise<ImageBitmap>;
    createImageBitmap(image: ImageBitmapSource, sx: number, sy: number, sw: number, sh: number, options?: ImageBitmapOptions): Promise<ImageBitmap>;
```

Although `performance.measure('name')` now satisfies both overloads, it seems TypeScript doesn't complain.

Issue #1282, opened 2022-03-02
Ping @saschanaz whenever `Update core dependencies` action fails


saschanaz:
Something should ping me since I'm not the owner.

saschanaz:
Hi @saschanaz, kindly pinging you as the 'Update core dependencies' job failed today.

saschanaz:
Hi @saschanaz, kindly pinging you as the 'Update core dependencies' job failed today.

sandersn:
https://github.com/microsoft/TypeScript-DOM-lib-generator/actions/runs/11908769649/job/33184821427 shows that the error is "Unknown DOM type: GPUError":

```
Error: Unknown DOM type: GPUError
    at convertDomTypeToTsTypeSimple (file:///home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/lib/build/emitter.js:322:[15](https://github.com/microsoft/TypeScript-DOM-lib-generator/actions/runs/11908769649/job/33184821427#step:9:16))
    at convertBaseType (file:///home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/lib/build/emitter.js:266:24)
    at convertDomTypeToTsTypeWorker (file:///home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/lib/build/emitter.js:278:22)
    at convertDomTypeToTsType (file:///home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/lib/build/emitter.js:247:22)
    at Array.map (<anonymous>)
    at convertDomTypeToTsTypeWorker (file:///home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/lib/build/emitter.js:280:14)
    at convertDomTypeToTsType (file:///home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/lib/build/emitter.js:247:22)
    at convertDomTypeToTsReturnType (file:///home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/lib/build/emitter.js:251:[22](https://github.com/microsoft/TypeScript-DOM-lib-generator/actions/runs/11908769649/job/33184821427#step:9:23))
[...]

PR #1283, opened 2022-03-02
Update to bcd@4.1.7

saschanaz:
Manually updating because of #1282. (Not updating to the latest one since it needs more investigation)


saschanaz:
lgtm

PR #1292, opened 2022-03-17
feat: add serviceworkerRegistration.navigationPreload

mathe42:
We now get support in FF Nightly and Safari TP so I think this can be added now. (Not 100% shure as we only have it stable in chromium)

see https://developer.mozilla.org/en-US/docs/Web/API/NavigationPreloadManager#browser_compatibility

feel free to close this (or don't merge it) until we get 2 stable implementations...


saschanaz:
This was removed by 3575d1b7c760e2bbf44a5b62e23e146055a00bbc when BCD didn't have the relevant item, but it now does and is supported by all three browser (except Firefox where it's still in beta stage https://github.com/mdn/browser-compat-data/issues/15414). It meets the standard to be added, so:

lgtm

PR #1296, opened 2022-03-22
fix: audioworklet registerProcessor should expect a constructor not a function

### LINKED ISSUE #1295: AudioWorkletProcessor can't be constructed like in the examples
See https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletProcessor/AudioWorkletProcessor

When I copy the example code I get the following error:
```
Argument of type 'typeof TestProcessor' is not assignable to parameter of type 'AudioWorkletProcessorConstructor'.  Type 'typeof TestProcessor' provides no match for the signature '(options: any): AudioWorkletProcessor'.
```

Example provided:
```
// test-processor.js
class TestProcessor extends AudioWorkletProcessor {
  constructor (options) {
    super()
    console.log(options.numberOfInputs)
    console.log(options.processorOptions.someUsefulVariable)
  }
  process (inputs, outputs, parameters) {
    return true
  }
}

registerProcessor('test-processor', TestProcessor)
```

mathe42:
fixes #1295

see also https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletProcessor/AudioWorkletProcessor

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -890,7 +890,7 @@ declare namespace WebAssembly {
 }
 
 interface AudioWorkletProcessorConstructor {
-    (options: any): AudioWorkletProcessor;
+    new (options: any): AudioWorkletProcessor;
 }
 
 interface PerformanceObserverCallback {
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -366,6 +366,11 @@
                 "overrideSignatures": [
                     "new (...params: any[]): HTMLElement"
                 ]
+            },
+            "AudioWorkletProcessorConstructor": {
+                "overrideSignatures": [
+                    "new (options: any): AudioWorkletProcessor"
+                ]
             }
         }
     },


saschanaz:
Could you add a unit test too?

And I wonder a new type is needed here:

```ts
interface AudioWorkletProcessorImpl extends AudioWorkletProcessor {
  process(...);
}
```

mathe42:
Will look into it...

unittests/files/audioworklet/usage.ts
```diff
@@ -0,0 +1,8 @@
+registerProcessor(
```
mathe42, line 1: Added test in extra folder as other files has to be included. If other setup is needed say how to change it.

baselines/audioworklet.generated.d.ts
```diff
@@ -216,6 +216,15 @@ declare var AudioWorkletProcessor: {
     new(): AudioWorkletProcessor;
 };
 
+interface AudioWorkletProcessorImpl extends AudioWorkletProcessor {
+    process(inputs: Float32Array[][], outputs: Float32Array[][], parameters: {[name: string]: Float32Array}): boolean;
```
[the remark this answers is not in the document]
saschanaz: You can use `Record<string, Float32Array>` there.

baselines/audioworklet.generated.d.ts
```diff
@@ -216,6 +216,15 @@ declare var AudioWorkletProcessor: {
     new(): AudioWorkletProcessor;
 };
 
+interface AudioWorkletProcessorImpl extends AudioWorkletProcessor {
+    process(inputs: Float32Array[][], outputs: Float32Array[][], parameters: Record<string, Float32Array>): boolean;
+}
+
+declare var AudioWorkletProcessorImpl: {
```
yume-chan, line 223: It doesn't really exist right?

mathe42: yes how do I prevent this?

yume-chan: Not sure. Try removing `"exposed": "AudioWorklet"` in override?

saschanaz: noInterfaceObject will do the work.

saschanaz:
lgtm

Thank you!

mathe42:
@saschanaz did not get merged!?

PR #1299, opened 2022-03-23
Manually update to bcd@4.1.13

saschanaz:


saschanaz:
lgtm

saschanaz:
What, it doesn't even check the result?

orta:
It does check the results: https://github.com/OSS-Docs-Tools/code-owner-self-merge/blob/8714b4990278162db0fb87717dd4a1e56fbd7a36/index.js#L179-L191

Maybe there's a case missing?

PR #1300, opened 2022-03-30
fix: Add missing "preservesPitch" field to HTMLMediaElement

huytd:
This PR adds the missing `preservesPitch` field to `HTMLMediaElement`.

Fixes the issue https://github.com/microsoft/TypeScript/issues/48444

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -7216,6 +7216,8 @@ interface HTMLMediaElement extends HTMLElement {
     readonly played: TimeRanges;
     /** Gets or sets a value indicating what data should be preloaded, if any. */
     preload: "none" | "metadata" | "auto" | "";
+    /** Gets or sets a flag that indicate whether or not the browser should adjust the pitch of the audio to compensate for changes to the playback rate. */
+    preservesPitch: boolean;
     readonly readyState: number;
     readonly remote: RemotePlayback;
     /** Returns a TimeRanges object that represents the ranges of the current media resource that can be seeked. */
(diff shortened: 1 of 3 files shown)


orta:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/preservesPitch#browser_compatibility - looks reasonable to me, unsure why it's not in there by default tbh

saschanaz:
This is fixed by https://github.com/mdn/browser-compat-data/pull/15474 and will be automatically added when a new release happens. I'd just wait rather than merging this and then rolling back.

huytd:
Got it. I saw that PR merged but not realized that it wasn't released. Thank you so much for the feedback! So I think we can close this PR 😅

saschanaz:
Hmm, oh wait, actually that PR only adds spec_url which shouldn't matter here. What matters is that the attribute is only available in standard form and is prefixed everywhere else.

huytd:
I see, look like there is a ticket to unprefixing the `preservesPitch` in other browsers:

- https://bugs.webkit.org/show_bug.cgi?id=214922
- https://bugs.webkit.org/show_bug.cgi?id=214922

From my understanding, once they got fixed, this issue could also be fixed by the generator. Should we just wait for them?

saschanaz:
Yup, browsers are willing to implement it afaict, we can wait 👍

orta:
ah yeah, gotta learn how to read these new MDN tables - nice spot

huytd:
Cool. Let me close this PR and make a comment on the original ticket. Thank you so much @saschanaz @orta !

PR #1301, opened 2022-03-31
Add the missing "disabled" property to HTMLStyleElement

huytd:
Fixes https://github.com/microsoft/TypeScript/issues/42266

Look like the `HTMLStyleElement.disabled` property is defined in [the specs](https://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-16428977), also available on every browser, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLStyleElement#browser_compatibility).

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -1128,7 +1128,6 @@
                 }
             },
             "HTMLStyleElement": {
-                "name": "HTMLStyleElement",
                 "properties": {
                     "property": {
                         "disabled": {
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -1126,6 +1126,17 @@
                         }
                     }
                 }
+            },
+            "HTMLStyleElement": {
+                "name": "HTMLStyleElement",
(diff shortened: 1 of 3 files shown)


inputfiles/addedTypes.jsonc
```diff
@@ -1126,6 +1126,17 @@
                         }
                     }
                 }
+            },
+            "HTMLStyleElement": {
+                "name": "HTMLStyleElement",
```
saschanaz, line 1131: This line is redundant, please remove it.

saschanaz (commented):
First, you are looking at a legacy spec, the correct one doesn't have it: https://html.spec.whatwg.org/multipage/semantics.html#htmlstyleelement

But that said, it's true that every browser has it and the spec wants to define it: https://github.com/whatwg/html/issues/5590

So yeah, let's do this.

inputfiles/addedTypes.jsonc
huytd: Thank you, I updated the code to remove it

yositripAction (commented):
approved

PR #1319, opened 2022-04-29
Move more deprecated HTML tags to HTMLElementDeprecatedTagNameMap

saschanaz:


saschanaz:
Huh, it requires signing again anyway.

PR #1326, opened 2022-05-04
fix: `waitUntil` and `transaction` type

susiwen8:
- close https://github.com/microsoft/TypeScript/issues/48601
- close https://github.com/microsoft/TypeScript/issues/48553

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -570,6 +570,10 @@ interface IDBObjectStoreParameters {
     keyPath?: string | string[] | null;
 }
 
+interface IDBTransactionOptions {
+    durability?: IDBTransactionDurability;
+}
+
 interface IDBVersionChangeEventInit extends EventInit {
     newVersion?: number | null;
     oldVersion?: number;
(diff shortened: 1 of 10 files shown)


HolgerJeromin:
Thanks for helping.
You manipulated a generated file. 
The [correct place](https://github.com/Microsoft/TSJS-lib-generator#input-files) would be `overridingTypes.jsonc`

susiwen8:
> Thanks for helping. You manipulated a generated file. The [correct place](https://github.com/Microsoft/TSJS-lib-generator#input-files) would be `overridingTypes.jsonc`

Thanks!

saschanaz:
There is a reason that `ExtendableEvent.waitUntil` accepts `any`: It basically accept any object and wraps it with `Promise.resolve()` if it's not a promise.

But strictly requiring a promise seems to make a sense in this case.

susiwen8:
> There is a reason that `ExtendableEvent.waitUntil` accepts `any`: It basically accept any object and wraps it with `Promise.resolve()` if it's not a promise.
> 
> But strictly requiring a promise seems to make a sense in this case.

Yes, I saw you [commit](https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/620ea42f857f32716e0e3b0a7a78e69b6c58b043) 3 years ago. 
I read [doc](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil#parameters) in `MDN`. Telling that `waitUntil` accept a `promise`.

susiwen8:
So I should change `waitUntil` back?

saschanaz:
You can keep it, but you need to follow what @HolgerJeromin said: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1326#issuecomment-1117415511

susiwen8:
@saschanaz Tests passed. Please take a look.

saschanaz:
It's still running 😉

saschanaz:
lgtm

(Checking if this repo supports automerge)

PR #1330, opened 2022-05-10
Disallow clearing timers without no argument at all

Andarist:
This aligns the signatures with the recent change in `@types/node` ([here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59914)).

My rationale behind this change is outlined [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59914#issuecomment-1122085478), but in a short, I don't think those calls make any sense:
```js
clearTimeout() // what do we clear?
clearInterval() // what do we clear?
```


inputfiles/overridingTypes.jsonc
```diff
@@ -218,6 +218,22 @@
                         }
                     }
                 }
+            },
+            "WindowOrWorkerGlobalScope": {
+                "methods": {
+                    "method": {
+                        "clearInterval": {
+                            "overrideSignatures": [
+                                "clearInterval(id: number | undefined): void"
```
saschanaz, line 227: `overrideSignatures` should be used as a last resort. Could you do the `signature` way as it's done below in OnErrorEventHandlerNonNull etc.?

inputfiles/overridingTypes.jsonc
```diff
@@ -218,6 +218,38 @@
...
+                            }
+                        },
+                        "clearTimeout": {
+                            "signature": {
+                                "0": {
+                                    "param": [
+                                        {
+                                            "name": "id",
+                                            "optional": "false",
+                                            "overrideType": "number | undefined"
```
saschanaz, line 245: `overrideType` should be also a last resort, perhaps `additionalTypes: ["undefined"]`?

Andarist:
@saschanaz thanks for the suggestions. Sorry for a few rapid pushes - I forgot that I had to accept baselines on top of rerunning the build script and didn't know at first how I should verify my changes locally 😅 I think this build will finally pass and it comes with your suggestions applied.

saschanaz:
Thanks!

lgtm

Issue #1174, opened 2021-10-02
Inconsistency between input event and beforeinput event


sapphi-red:
Now beforeinput event is `InputEvent` (because it is defined with addedTypes [here](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/5706098631f053f8000a2844f2819f574756ffeb/inputfiles/addedTypes.jsonc#L64-L65)).
On the other hand, input event is `Event` (this is from [HTML Standard](https://html.spec.whatwg.org/multipage/indices.html#event-input-input)).

[`beforeinput` event is defined in UIEvents](https://www.w3.org/TR/uievents/#event-type-beforeinput) and here it is defined that [`input` is also `InputEvent`](https://www.w3.org/TR/uievents/#event-type-input).

I think input event should be `InputEvent`.

saschanaz:
This is because `input` does not always use `InputEvent`. Copy below to the address bar and see what happen on the console.

`data:text/html,<input type=range oninput="console.log(event.constructor.name)"><input type=text oninput="console.log(event.constructor.name)">`

On `input[type=range]` the event uses `Event` and on `input[type=text]` it uses `InputEvent`. To cover both it should be `Event` rather than `InputEvent`.

sapphi-red:
I see.
Thank you!

L1Q:
Is there a way to specify it can be both? I'd expect `GlobalEventHandlersEventMap` in `lib.dom.d.ts` to state 
``
    "input": Event | InputEvent;
``

saschanaz:
That's no different than just Event, though...

L1Q:
Union types TS open the way for type guards and optional chaining. The current way only calls for unsafe type casts.

saschanaz:
Event is a strict superset of InputEvent and thus unioning them just gives Event, with no additional type gaurd.

L1Q:
You *can* provide your own guard though.

saschanaz:
How?

L1Q:
Basically, this works perfectly fine, correctness-wise
```ts
const input = document.createElement('input')
input.setAttribute('type', 'text')
input.addEventListener('input', (ev: Event) => {
    if(ev instanceof InputEvent) {
        console.log('inputting text using input type:', ev.inputType)
        input.setAttribute('type', 'range')
    } else {
        console.log('manipulating slider')
        input.setAttribute('type', 'text')
    }
})
document.body.appendChild(input)
```

[...]

PR #1332, opened 2022-05-15
Fix "input" event type

L1Q:
As per the spec, the Event interface for the "input" event is InputEvent, the same as "beforeinput"
https://www.w3.org/TR/2021/WD-uievents-20211020/#event-type-input

```ts
const input_element = document.createElement('input')

// currently OK
input_element.addEventListener('beforeinput', (ev: InputEvent) => {})
// currently fails, expects ev to be of 'Event' type
input_element.addEventListener('input', (ev: InputEvent) => {})
```
[>> playground](https://www.typescriptlang.org/play?ts=4.8.0-dev.20220514#code/MYewdgzgLgBAlmADgVygfQKYBsMFsNiwC8MAJiMMvoQHTABOGAhlBgKI7VQAUA5Aiii8AlAFgAUBIGpMnAlBpNSpNgDd5AGTjQCGenwBGGAGYhG0oQBoY3DKoBcMAJJJUa+cJhEAfDADeAL5i4hayePKKyu6EWjpgenwWvNa2Ds6uUNFQnj7+QUA)

Related MDN and W3C:
[MDN beforeinput](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event) | [W3C beforeinput](https://www.w3.org/TR/2021/WD-uievents-20211020/#event-type-beforeinput)
[MDN input](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) | [W3C input](https://www.w3.org/TR/2021/WD-uievents-20211020/#event-type-input)

note that MDN page for "input" event states the interface can be ``InputEvent or Event``.

In practice, both events are instanceof InputEvent in Chrome 101 and Firefox 100.

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -133,6 +133,10 @@
                             "type": "FormDataEvent"
                         },
                         {
+                            "name": "input",
+                            "type": "InputEvent"
+                        },
+                        {
                             "name": "keydown",
                             "type": "KeyboardEvent"
                         },


saschanaz:
Not all input events use InputEvent, MDN is right here. See also https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1174#issuecomment-933042088

@github-actions close

L1Q:
I see, thanks!

PR #1334, opened 2022-05-16
override HeadersInit to replace `string[][]` to `[string, string][]`

ShuiRuTian:
A runtime error will be thrown if the array is not two-arity
```
new Headers([[1]]) // error, Failed to construct 'Headers': Invalid value
new Headers([[1,2,3]]) // error, Failed to construct 'Headers': Invalid value
```


inputfiles/addedTypes.jsonc
```diff
@@ -1212,7 +1212,7 @@
                 "name": "InsertPosition"
             },
             {
-                "additionalTypes": ["Headers"],
```
saschanaz, line 1215: `overrideType` should be a last resort especially when the original type is complex. This works:

```json
                "type": {
                    "0": {
                        "subtype": {
                            "overrideType": "[string, string]"
                        }
                    }
                },
                "additionalTypes": ["Headers"],
```

ShuiRuTian: done

saschanaz:
Cool, lgtm

Issue #1097, opened 2021-08-08
Decide what to do with comments


saschanaz:
Currently there are two ways to fetch comments
1. From MDN
2. From specs

Both are not being actively maintained. I think either one should be dropped and replaced by another one.

MattiasBuelens:
FYI: It looks like the `fetch-comment` and `fetch-mdn` scripts are now *both* broken.
* `fetch-comment` [requires](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/c86018dfd7edcf7a8959623c20abb1ae8d54b2dc/src/fetch-comment.ts#L20) `inputfiles/idlSources.json`, which was removed in #1044.
* `fetch-mdn` [requires](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/c86018dfd7edcf7a8959623c20abb1ae8d54b2dc/src/fetch-mdn.ts#L19) `inputfiles/browser.webidl.preprocessed.json`, which was removed in #1034.
* `fetch-mdn` tries to fetch JSON files from MDN using the URL template `https://developer.mozilla.org/en-US/docs/Web/API/${name}$json` (where `${name}` is the name of an IDL interface), but this no longer works. For example, `https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream$json` returns HTTP 404. There doesn't seem to be an official machine-friendly API anymore... 😕 The best we can do would be processing the Markdown source files from https://github.com/mdn/content, such as https://github.com/mdn/content/blob/main/files/en-us/web/api/readablestream/index.md.

saschanaz:
Safari DevTools seemingly imports comments for CSS properties from somewhere, not sure it's MDN or not.

Bashamega:
Hello @saschanaz 
I think this should be closed, because I have updated the comments in #1940

saschanaz:
Good point, closing.

PR #1349, opened 2022-06-25
Remove comment fetchers

saschanaz:
(See also https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1097#issuecomment-1004099360)

Both fetch-comment and fetch-mdn are not being maintained and are currently broken.

Theoretically markdown text can still be fetched from https://github.com/mdn/content/tree/main/files/en-us/web/api/ but that needs some effort. (It uses some MDN-specific syntax, etc.) I currently don't have enough interest to do that.

This fixes a build failure caused by `@types/jsdom` (https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/60969).


sandersn (approved):
This sounds reasonable since I tried to use these this year and never got them to work.
The original intent was to run them manually, right? Do the currently checked-in apiDescriptions.json and *.commentmap.json still get correctly applied?

saschanaz:
>The original intent was to run them manually, right?

Yup.

>Do the currently checked-in apiDescriptions.json and *.commentmap.json still get correctly applied?

I think they are, as there haven't been structure changes related to comments.

saschanaz:
Lgtm

PR #1350, opened 2022-06-30
fix SubtleCrypto encrypt/decrypt return

panva:
The return of encrypt/decrypt is currently `Promise<any>`, this fixes the issue, i think?

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -1415,7 +1415,9 @@
                                             "additionalTypes": ["RsaOaepParams", "AesCtrParams", "AesCbcParams", "AesGcmParams"]
                                         }
                                     ],
-                                    "overrideType": "Promise<ArrayBuffer>"
+                                    "subtype": {
+                                      "type": "ArrayBuffer"
+                                    }
                                 }
                             }
                         },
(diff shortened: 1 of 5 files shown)


inputfiles/overridingTypes.jsonc
```diff
@@ -1414,7 +1414,8 @@
                                             "name": "algorithm",
                                             "additionalTypes": ["RsaOaepParams", "AesCtrParams", "AesCbcParams", "AesGcmParams"]
                                         }
-                                    ]
+                                    ],
+                                    "overrideType": "Promise<ArrayBuffer>"
```
saschanaz, line 1418:
```suggestion
                                    "type": {
                                      "subtype": "ArrayBuffer"
                                    }
```

I know you just copied from the existing examples, but `overrideType` should be used only as a last resort.

inputfiles/overridingTypes.jsonc
```diff
@@ -1463,7 +1464,8 @@
...
```
saschanaz, line 1468:
```suggestion
                                    "subtype": {
                                      "type": "ArrayBuffer"
                                    }
```

saschanaz:
lgtm

See also: https://github.com/w3c/webcrypto/issues/293

PR #1362, opened 2022-07-19
Add readable byte streams

MattiasBuelens:
[With Firefox 102 adding support for readable byte streams](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/102), we now have two browser engines (Chromium and Gecko) shipping this feature. This means we can finally add official type definitions! 😁

Most of work was already done in #890 but then those types had to be removed. This PR brings them back, and hopefully they'll stay this time around. 😅

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -154,6 +154,21 @@ interface Transformer<I = any, O = any> {
     writableType?: undefined;
 }
 
+interface UnderlyingByteSource {
+    autoAllocateChunkSize?: number;
+    cancel?: UnderlyingSourceCancelCallback;
+    pull?: (controller: ReadableByteStreamController) => void | PromiseLike<void>;
+    start?: (controller: ReadableByteStreamController) => any;
+    type: "bytes";
+}
+
+interface UnderlyingDefaultSource<R = any> {
+    cancel?: UnderlyingSourceCancelCallback;
+    pull?: (controller: ReadableStreamDefaultController<R>) => void | PromiseLike<void>;
+    start?: (controller: ReadableStreamDefaultController<R>) => any;
+    type?: undefined;
+}
+
 interface UnderlyingSink<W = any> {
     abort?: UnderlyingSinkAbortCallback;
     close?: UnderlyingSinkCloseCallback;
@@ -549,6 +564,8 @@ interface ReadableStream<R = any> {
 
 declare var ReadableStream: {
     prototype: ReadableStream;
+    new(underlyingSource: UnderlyingByteSource, strategy?: { highWaterMark?: number }): ReadableStream<Uint8Array>;
+    new<R = any>(underlyingSource: UnderlyingDefaultSource<R>, strategy?: QueuingStrategy<R>): ReadableStream<R>;
     new<R = any>(underlyingSource?: UnderlyingSource<R>, strategy?: QueuingStrategy<R>): ReadableStream<R>;
 };
 
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -75,9 +75,18 @@ interface QueuingStrategyInit {
     highWaterMark: number;
 }
 
-interface ReadableStreamReadDoneResult {
+interface ReadableStreamGetReaderOptions {
+    /**
+     * Creates a ReadableStreamBYOBReader and locks the stream to the new reader.
+     *
+     * This call behaves the same way as the no-argument variant, except that it only works on readable byte streams, i.e. streams which were constructed specifically with the ability to handle "bring your own buffer" reading. The returned BYOB reader provides the ability to directly read individual chunks from the stream via its read() method, into developer-supplied buffers, allowing more precise control over allocation.
+     */
+    mode?: ReadableStreamReaderMode;
+}
+
+interface ReadableStreamReadDoneResult<T> {
     done: true;
-    value?: undefined;
+    value?: T;
 }
 
 interface ReadableStreamReadValueResult<T> {
@@ -154,10 +163,11 @@ interface UnderlyingSink<W = any> {
 }
 
 interface UnderlyingSource<R = any> {
+    autoAllocateChunkSize?: number;
     cancel?: UnderlyingSourceCancelCallback;
     pull?: UnderlyingSourcePullCallback<R>;
     start?: UnderlyingSourceStartCallback<R>;
-    type?: undefined;
+    type?: ReadableStreamType;
 }
 
 /** A controller object that allows you to abort one or more DOM requests as and when desired. */
@@ -529,7 +539,9 @@ declare var ReadableByteStreamController: {
 interface ReadableStream<R = any> {
     readonly locked: boolean;
     cancel(reason?: any): Promise<void>;
+    getReader(options: { mode: "byob" }): ReadableStreamBYOBReader;
     getReader(): ReadableStreamDefaultReader<R>;
+    getReader(options?: ReadableStreamGetReaderOptions): ReadableStreamReader<R>;
     pipeThrough<T>(transform: ReadableWritablePair<T, R>, options?: StreamPipeOptions): ReadableStream<T>;
     pipeTo(destination: WritableStream<R>, options?: StreamPipeOptions): Promise<void>;
     tee(): [ReadableStream<R>, ReadableStream<R>];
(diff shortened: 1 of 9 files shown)


baselines/dom.generated.d.ts
```diff
@@ -1485,9 +1485,18 @@ interface RTCTransportStats extends RTCStats {
...
+     * This call behaves the same way as the no-argument variant, except that it only works on readable byte streams, i.e. streams which were constructed specifically with the ability to handle "bring your own buffer" reading. The returned BYOB reader provides the ability to directly read individual chunks from the stream via its read() method, into developer-supplied buffers, allowing more precise control over allocation.
+     */
+    mode?: ReadableStreamReaderMode;
+}
+
+interface ReadableStreamReadDoneResult<T> {
     done: true;
-    value?: undefined;
+    value?: T;
 }
```
MattiasBuelens, line 1500: This is needed for BYOB readers, which return an empty view on the original buffer when the stream is closed.

I'm not sure if we still want separate types for `done: false` and `done: true`? On the one hand, it's nice that you can write:
```javascript
while (true) {
  const result = await reader.read();
  if (result.done) {
    break;
  }
  const chunk = result.value; // narrowed to T, cannot be undefined
}
```
On the other hand, I don't know if this a sufficiently convincing argument to keep this in TypeScript's DOM types. The stream specification itself only has [`ReadableStreamReadResult`](https://streams.spec.whatwg.org/#dictdef-readablestreamreadresult), which does not distinguish between the two cases.

baselines/dom.generated.d.ts
```diff
@@ -11377,7 +11389,7 @@ declare var ReadableStream: {
 };
 
 interface ReadableStreamBYOBReader extends ReadableStreamGenericReader {
-    read(view: ArrayBufferView): Promise<ReadableStreamReadResult<ArrayBufferView>>;
+    read<T extends ArrayBufferView>(view: T): Promise<ReadableStreamReadResult<T>>;
```
MattiasBuelens, line 11392: When you pass a `Uint8Array`, you're guaranteed to receive a `Uint8Array` back in the result. This type reflects that. 🙂

baselines/webworker.generated.d.ts
```diff
@@ -5876,9 +5888,9 @@ type NamedCurve = string;
 type OnErrorEventHandler = OnErrorEventHandlerNonNull | null;
 type PerformanceEntryList = PerformanceEntry[];
 type PushMessageDataInit = BufferSource | string;
-type ReadableStreamController<T> = ReadableStreamDefaultController<T>;
-type ReadableStreamReadResult<T> = ReadableStreamReadValueResult<T> | ReadableStreamReadDoneResult;
-type ReadableStreamReader<T> = ReadableStreamDefaultReader<T>;
+type ReadableStreamController<T> = ReadableStreamDefaultController<T> | ReadableByteStreamController;
```
MattiasBuelens, line 5891: This could break existing code, for example:
```javascript
new ReadableStream({
  start(controller) {
    controller.enqueue("a");
  }
});
```
We don't yet recognize that this is a non-byte stream, so we leave `controller` as `ReadableStreamDefaultController<T> | ReadableByteStreamController`. But that means you can **only** pass byte chunks to `enqueue()`.

I could investigate if we can change the `ReadableStream` constructor to have separate overloads for constructing byte streams and non-byte streams?

inputfiles/overridingTypes.jsonc
```diff
@@ -2380,11 +2380,23 @@
...
                             "signature": {
                                 "0": {
-                                    "overrideType": "ReadableStreamDefaultReader<R>"
+                                    "overrideType": "ReadableStreamReader<R>"
                                 }
-                            }
+                            },
+                            "force-references": [
+                                {
+                                    "type": "ReadableStreamDefaultReader"
```
saschanaz, line 2394: I think only dictionaries/typedefs need this, and these are interfaces.

baselines/audioworklet.generated.d.ts
```diff
@@ -529,7 +539,9 @@ declare var ReadableByteStreamController: {
 interface ReadableStream<R = any> {
     readonly locked: boolean;
     cancel(reason?: any): Promise<void>;
+    getReader(options: { mode: "byob" }): ReadableStreamBYOBReader;
```
saschanaz, line 542: Could you add a short unit test to make sure this works?

baselines/dom.generated.d.ts
saschanaz: Also a good candidate for a unit test.

saschanaz (commented):
Cool, thanks! Some nits.

baselines/webworker.generated.d.ts
```diff
@@ -2691,19 +2716,23 @@ declare var ReadableByteStreamController: {
...
     pipeThrough<T>(transform: ReadableWritablePair<T, R>, options?: StreamPipeOptions): ReadableStream<T>;
     pipeTo(destination: WritableStream<R>, options?: StreamPipeOptions): Promise<void>;
     tee(): [ReadableStream<R>, ReadableStream<R>];
 }
 
 declare var ReadableStream: {
     prototype: ReadableStream;
+    new(underlyingSource: UnderlyingByteSource, strategy?: { highWaterMark?: number }): ReadableStream<Uint8Array>;
+    new<R = any>(underlyingSource: UnderlyingDefaultSource<R>, strategy?: QueuingStrategy<R>): ReadableStream<R>;
     new<R = any>(underlyingSource?: UnderlyingSource<R>, strategy?: QueuingStrategy<R>): ReadableStream<R>;
```
MattiasBuelens, line 2731: These additional constructor signatures make these examples work:
```javascript
new ReadableStream({
  start(controller) {
    controller.enqueue("a");
  }
});

new ReadableStream({
  type: "bytes",
  start(controller) {
    controller.byobRequest?.respond(1);
  }
});
```
(And yes, there are unit tests. 😉)

How do we feel about this? I think it's worth the extra complexity, since these are literally the textbook examples on how developers can construct a readable stream.

baselines/webworker.generated.d.ts
MattiasBuelens: I gave it a go! 😁

MattiasBuelens (commented):
Unit tests are added.

I also added additional constructor signatures for `ReadableStream`, to distinguish between byte streams and non-byte streams. (This is a bit complex, so please take a look!)

inputfiles/addedTypes.jsonc
```diff
@@ -1195,6 +1195,63 @@
...
+                "members": {
+                    "member": {
+                        "type": {
+                            "name": "type",
+                            "required": true,
+                            "overrideType": "\"bytes\""
+                        },
+                        "autoAllocateChunkSize": {
+                            "name": "autoAllocateChunkSize",
+                            "overrideType": "number"
```
saschanaz, line 1210:
```suggestion
                            "type": "unsigned long long"
```

Kinda weird but I'd prefer `type` here whenever it works.

inputfiles/overridingTypes.jsonc
```diff
@@ -2349,6 +2349,20 @@
                 "constructor": {
                     "signature": {
                         "0": {
+                            "param": [
+                                {
+                                    "name": "underlyingSource",
+                                    "optional": false,
+                                    "overrideType": "UnderlyingByteSource"
```
saschanaz, line 2356:
```suggestion
                                    "type": "UnderlyingByteSource"
```

(I believe this works)

saschanaz (commented):
Thank you!

(Sorry for responding late, I was off and playing games 😄. My vacation does not really apply here, but whatever.)

unittests/files/streams.ts
```diff
@@ -1,16 +1,25 @@
...
     controller.desiredSize;
     controller.enqueue("a");
   },
   pull(controller) {
+    assertType<ReadableStreamDefaultController<string>>(controller);
     controller.enqueue("b");
     controller.close();
   },
+  cancel(reason) {
+    assertType<any>(reason);
```
saschanaz, line 15: Does this really do anything? 😄 (But probably good to be explicit)

saschanaz (commented):
Okay everything's cool, thank you for additional assertions!

saschanaz:
LGTM

unittests/files/streams.ts
MattiasBuelens: ¯\\\_(ツ)_/¯

PR #1363, opened 2022-07-19
Return CryptoKeyPair in SubtleCrypto#importKey

### LINKED ISSUE #1361: SubtleCrypto#import missing jwk CryptoKeyPair result
For now SubtleCrypto#import only allows jwk importing results to be a CryptoKey while JWK [allows storing key pair](https://stackoverflow.com/questions/72151096/how-to-derive-public-key-from-private-key-using-webcryptoapi/72153942#72153942)

BANOnotIT:
fixes #1361


mathe42:
In the spec https://w3c.github.io/webcrypto/#SubtleCrypto-method-importKey  it is statet in step 8 that the result is allways of type CryptoKey so I'm not sure if this change is valid.

I think the stackoverflow link you provided is not what you think. Yes we can import a key-pair but as i understand the private key is just droped and only the public is provided...

PR #1369, opened 2022-07-26
feat: AudioParamMap extends ReadonlyMap<string, AudioParam>

### LINKED ISSUE #1358: [AudioWorklet] Missing properties in AudioParamMap interface
According to official documentation:

https://webaudio.github.io/web-audio-api/

https://www.w3.org/TR/webaudio/#audioparammap

This interface has "entries", "forEach", "get", "has", "keys", "values", @@iterator methods and a "size" getter brought by readonly maplike.

Captura de pantalla 2022-07-05 a las 19 36 47

but in node_modules/typescript/lib/lib.dom.d.ts (https://www.npmjs.com/package/@types/audioworklet) we only have

Captura de pantalla 2022-07-05 a las 19 37 40

Is there any timeline for when this interface will be updated?

Thanks!

mathe42:
fixes #1358

Note that the forEach method could be removed as it exists on the ReadonlyMap Interface. I tried to remove it by adding it to removedTypes.json but failed.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2276,7 +2276,7 @@ declare var AudioParam: {
     new(): AudioParam;
 };
 
-interface AudioParamMap {
+interface AudioParamMap extends ReadonlyMap<string, AudioParam> {
     forEach(callbackfn: (value: AudioParam, key: string, parent: AudioParamMap) => void, thisArg?: any): void;
 }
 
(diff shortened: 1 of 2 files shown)


mathe42:
The test fails... Why is ReadonlyMap not defined?

saschanaz:
This is not needed, see https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1358#issuecomment-1195989733

mathe42:
Yes - missed that. Sorry.

PR #1378, opened 2022-08-14
add automation for Transferable

### LINKED ISSUE #1197: Autogenerate `type Transferable`
It's essentially the list of interfaces with `[Transferable]` extended attribute. It already lacks stream objects which are transferable, it should be autogenerated to prevent such omission rather than manually defined in overrideTypes.json.

mathe42:
This is a try to fix #1197. This might not the optimal solution! 

I simply checked for all Interfaces if they have the `Transferable` attribute and added it to the `addedItems` object where it was defined before.

This is my first code-contribution here so there might be things that I missed.


ExE-Boss:
I think that this should use a similar implementation to `ArrayBufferLike`[^1]:
```ts
/**
 * Allowed ArrayBuffer types for the buffer of an ArrayBufferView and related Typed Arrays.
 */
interface ArrayBufferTypes {
	ArrayBuffer: ArrayBuffer;
}
type ArrayBufferLike = ArrayBufferTypes[keyof ArrayBufferTypes];
```

[^1]: <https://github.com/microsoft/TypeScript/blob/03b12a6a45ca0dc043896160fd81060a980352d9/lib/lib.es5.d.ts#L1586-L1592>

mathe42:
That is a good solution for ArrayBufferLike as this can be extended (as it is a ES-Syntax-Class) by the Runntime (Browser, Node, Deno) with for example SharedArrayBuffer but Transferable is a Web-Feature that doesn't need to be extended. (Yes NodeJS and Deno adopt Transferable but they have theire own types independent from @types/web).

---

TL;DR: As this is nothing a user / runtime can extend. I prefere to keep it as it currently is.

saschanaz:
This is an interesting approach! I thought I'd do it in emitter but this looks better than playing manually with strings.

LGTM

HolgerJeromin:
Should this list be sorted to have a stable output?

PR #1381, opened 2022-08-18
Fix LICENSE of published packages

jakebailey:
The license on this repo is Apache 2.0, and we also prepend an Apache 2.0 copyright header onto these files when distributed with TypeScript. But, when the packages like `@types/web` were introduced, they were instead shipped with the MIT license, which isn't the right one for this code.


DanielRosenwasser (approved):
For all intents and purposes, I don't think it matters to most consumers. This was likely an oversight, but for consistency we should align with whatever `lib.d.ts` already ships as within TypeScript. Otherwise this will cause more issues down the line.

Issue #1384, opened 2022-08-29
according to DOM specs cancelBubble is a legacy alias of .stopPropagation()


Ahmed-Hakeem:
# lib Update Request

## Missing / Incorrect Definition

According to DOM spec cancelBubble is a legacy alias of .stopPropagation(), and after exploring library code this issue happened because of where the property value came from, It is read from the property deprecated value  from data.json (cancelBubble.__compact.status.deprecated) which is exported by @mdn/browser-compat-data v5.0.0, this value is read inside mapInterfaceLike function in src/build/bcd/mapper.ts

This issue has been resolved in @mdn/browser-compat-data starting from v5.1.9 inside 
commit: 36ffef821fafb429430f0286009787505a3ff000 

so I think it will be solved by only updating @mdn/browser-compat-data 

## Documentation Link

https://dom.spec.whatwg.org/#interface-event

## Reference 
https://github.com/microsoft/TypeScript/issues/50517

saschanaz:
Fixed in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1394, thanks!

@github-actions close

Ahmed-Hakeem:
> Fixed in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1394, thanks!
> 
> 
> 
> @github-actions close

It's my pleasure 😊❤️

PR #1386, opened 2022-08-30
Update @mdn/browser-compat-data to fetch updated mdn API data

Ahmed-Hakeem:
this update will solve this issue  #1384

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -298,6 +298,7 @@ declare var ErrorEvent: {
 interface Event {
     /** Returns true or false depending on how event was initialized. True if event goes through its target's ancestors in reverse tree order, and false otherwise. */
     readonly bubbles: boolean;
+    /** @deprecated */
     cancelBubble: boolean;
     /** Returns true or false depending on how event was initialized. Its return value does not always carry meaning, but true can indicate that part of the operation during which event was dispatched, can be canceled by invoking the preventDefault() method. */
     readonly cancelable: boolean;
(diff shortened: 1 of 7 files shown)


mathe42:
You have to run `npm i` `npm run build` and `npm run baseline-accept` localy and commit results!

saschanaz:
https://github.com/mdn/browser-compat-data/issues/17127 is the major blocker here. We'll have to workaround it somehow, probably in bcd-idl-mapper.

saschanaz:
Fixed in #1394, thanks!

@github-actions close

Ahmed-Hakeem:
> Fixed in #1394, thanks!
> 
> 
> 
> @github-actions close

It's my pleasure 😊❤️, I think this will solve this issue but it won't solve your concern here mdn/browser-compat-data#17127 you will have to care about it in every build until you make a workaround inside bcd-idl-mapper.

I hoped you made these changes inside this pull request, but anyway the most important thing is that the issue has been solved, thanks ❤️.

saschanaz:
>I think this will solve this issue but it won't solve your concern here https://github.com/mdn/browser-compat-data/issues/17127 you will have to care about it in every build until you make a workaround inside bcd-idl-mapper.

I worked around it in https://github.com/saschanaz/bcd-idl-mapper/blob/7c0d88d755a5836b9d6e4550f850be2f2e9f8a4b/lib/mixin/search.js#L33, which this generator depends on. For now it should be sufficient for us.

Ahmed-Hakeem:
> >I think this will solve this issue but it won't solve your concern here https://github.com/mdn/browser-compat-data/issues/17127 you will have to care about it in every build until you make a workaround inside bcd-idl-mapper.
> 
> 
> 
> I worked around it in https://github.com/saschanaz/bcd-idl-mapper/blob/7c0d88d755a5836b9d6e4550f850be2f2e9f8a4b/lib/mixin/search.js#L33. For now it should be sufficient for us.

Great, thank you ❤️

PR #1391, opened 2022-09-04
Don't drop "conditional" from CredentialMediationRequirement.

agl:
This removal was added in c06ba3140cf9a60621747f5760d270a8008a502d
without a specific commit message. The comment says that this enum value
depends on `isConditionalMediationAvailable`, which is true, but a
number of APIs depend on feature detection yet are not removed.

This enum value is supported in Safari (as of iOS 16 & macOS 13) and
Chromium and, without this change, developers will have to cast
"conditional" whenever used.


agl:
(Never mind. Can't sign CLA so hopefully someone else will run into this same issue and fix it.)

saschanaz:
That "depends on" comment is not about feature detection but about the support status of the corresponding IDL attribute. In other words, it should exist only if the attribute exists. Unfortunately I haven't really automated this so far :(

saschanaz:
(Filed https://github.com/mdn/browser-compat-data/issues/17827)

christianb-cc:
> (Filed [mdn/browser-compat-data#17827](https://github.com/mdn/browser-compat-data/issues/17827))

Looks like it is supported by Chrome, and Safari now. Are we able to bump this? Trying to use TypeScript with a webauthn implementation :)

@saschanaz

saschanaz:
https://github.com/mdn/browser-compat-data/blob/e0f7feac0db49548fa953beb605f03c3814d4a82/api/Credential.json#L79 suggests it's not in Chrome, can you file a bug there if Chrome indeed supports it?

agl:
> suggests it's not in Chrome

As the manager of Chrome's WebAuthn team I can confirm that Chrome supports it for public-key credentials.

saschanaz:
In that case please file an issue on BCD and it will be half-automatically added here.

christianb-cc:
> Actually it seems Chrome does have `PublicKeyCredential.isConditionalMediationAvailable()` but not `Credential.isConditionalMediationAvailable()`, is it somehow slipped? Anyway I'll go merge #1454.

Excellent, thank you, also sorry for commenting on the wrong thread. I had multiple PRs open and got confused. I intended to comment on #1454.

saschanaz:
#1454 is blocked on CLA too, it'll be helpful if you can file another PR.

PR #1400, opened 2022-09-17
Manually update to webref/idl@3.17.0

### LINKED ISSUE #1370: [WebRTC] A lot of metrics are missing
If you look at RTCStats and its subdirectory you'll notice the metrics are very much not up to date.

For example [RTCOutboundRtpStreamStats in dom.generated.d.ts](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/main/baselines/dom.generated.d.ts#L1343) appears to be missing 20-30 or so metrics compared to [the spec](https://w3c.github.io/webrtc-stats/#outboundrtpstats-dict*). This is not a case of "[not implemented by more than one browser](https://github.com/microsoft/TypeScript-DOM-lib-generator#why-is-my-fancy-api-still-not-available-here)", as we can see in the [Web Platform Tests dashboard](https://wpt.fyi/results/webrtc-stats/supported-stats.html?label=experimental&label=master&aligned) there are a lot of metrics implemented by multiple browsers that are not present here.

I don't know how the generator works, but it's worth noting that WebRTC stats do not have a WebIDL definition in the Chromium code base since these metrics objects are JS dictionaries that get converted from lower layer [C++ classes](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/stats/rtcstats_objects.h) so this is automatic.

### LINKED ISSUE #1390: `Selection.modify()` is missing.
See [this playground example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBBCmAbewoEtwwLwwCYmAFcBbeMKAOgHN4oBlJFdcACgEoBuAKDQDMYWAQgTJUGMGxhQAFgCcQAdxhh4SgKKz5slgCIAciDiMx4HZy5cRTcRWIhcfAJ674ADyhlcOgDQwdyXigfP2BpAENZMNR4WTMOIA).

```js
const selection = document.getSelection();
if (!selection) throw new Error("No selection");

selection.modify("extend", "left", "character");
//        ^^^^^^---- Property 'modify' does not exist on type 'Selection'.
```

`Selection.modify()` seems to be supported in [all major browsers](https://developer.mozilla.org/en-US/docs/Web/API/Selection/modify#browser_compatibility).

saschanaz:
The mass removal of RTC items in removeTypes reflects https://github.com/w3c/webrtc-stats/pull/622 but also the much better implementation status.

Fixes #1370
Fixes #1390

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -38,24 +38,6 @@ interface MessageEventInit<T = any> extends EventInit {
     source?: MessageEventSource | null;
 }
 
-interface PerformanceMarkOptions {
-    detail?: any;
-    startTime?: DOMHighResTimeStamp;
-}
-
-interface PerformanceMeasureOptions {
-    detail?: any;
-    duration?: DOMHighResTimeStamp;
-    end?: string | DOMHighResTimeStamp;
-    start?: string | DOMHighResTimeStamp;
-}
-
-interface PerformanceObserverInit {
-    buffered?: boolean;
-    entryTypes?: string[];
-    type?: string;
-}
-
 interface PromiseRejectionEventInit extends EventInit {
     promise: Promise<any>;
     reason?: any;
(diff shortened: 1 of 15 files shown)


saschanaz:
lgtm

Issue #1393, opened 2022-09-08
incomplete FileSystemHandle types


jozefchutka:
Please update types for File System Access API:

[FileSystemDirectoryHandle](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/85672cd2b4eeb2852897f05af1cfa8a5faa24974/baselines/sharedworker.generated.d.ts#L1449) is missing [entries/values iterators](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle)

[FileSystemFileHandle](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/85672cd2b4eeb2852897f05af1cfa8a5faa24974/baselines/sharedworker.generated.d.ts#L1463) missing [createWritable](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle)

[FileSystemHandle](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/85672cd2b4eeb2852897f05af1cfa8a5faa24974/baselines/sharedworker.generated.d.ts#L1474) is missing
[queryPermission and requestPermission](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle)

mathe42:
createWritable, queryPermission and requestPermission are not missing they are only supported in chromium so they are not added see for example https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createWritable#browser_compatibility

For [FileSystemDirectoryHandle](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/85672cd2b4eeb2852897f05af1cfa8a5faa24974/baselines/sharedworker.generated.d.ts#L1449) is missing [entries/values iterators](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle) on first I think you are right not 100% sure maybe that would just need a bump of the BCD version to be fixed...

saschanaz:
> [FileSystemDirectoryHandle](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/85672cd2b4eeb2852897f05af1cfa8a5faa24974/baselines/sharedworker.generated.d.ts#L1449) is missing [entries/values iterators](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle)

[...]

PR #1403, opened 2022-09-18
Add async iterable flavors

MattiasBuelens:
Add support for `async iterable` declarations [per Web IDL](https://webidl.spec.whatwg.org/#idl-async-iterable), and emit them as `.asynciterable.d.ts` flavors. This is a first step towards microsoft/TypeScript#29867.

At the time of writing, only `FileSystemDirectoryHandle`'s async iterator is supported by two browser engines (Chrome and Safari). `ReadableStream`'s async iterator is not implemented anywhere yet, and gets correctly filtered out by `getRemovalData()`.

Fixes part of #1393.

### DIFF
--- a/baselines/audioworklet.asynciterable.generated.d.ts
+++ b/baselines/audioworklet.asynciterable.generated.d.ts
@@ -1,0 +1,8 @@
+/////////////////////////////
+/// AudioWorklet Async Iterable APIs
+/////////////////////////////
+
+interface ReadableStream<R = any> {
+    [Symbol.asyncIterator](options?: ReadableStreamIteratorOptions): AsyncIterableIterator<any>;
+    values(options?: ReadableStreamIteratorOptions): AsyncIterableIterator<any>;
+}
--- a/baselines/audioworklet.asynciterable.generated.d.ts
+++ b/baselines/audioworklet.asynciterable.generated.d.ts
@@ -3,6 +3,6 @@
 /////////////////////////////
 
 interface ReadableStream<R = any> {
-    [Symbol.asyncIterator](options?: ReadableStreamIteratorOptions): AsyncIterableIterator<any>;
-    values(options?: ReadableStreamIteratorOptions): AsyncIterableIterator<any>;
+    [Symbol.asyncIterator](options?: ReadableStreamIteratorOptions): AsyncIterableIterator<R>;
+    values(options?: ReadableStreamIteratorOptions): AsyncIterableIterator<R>;
 }
--- a/baselines/audioworklet.asynciterable.generated.d.ts
+++ b/baselines/audioworklet.asynciterable.generated.d.ts
@@ -1,8 +1,3 @@
 /////////////////////////////
 /// AudioWorklet Async Iterable APIs
 /////////////////////////////
-
-interface ReadableStream<R = any> {
-    [Symbol.asyncIterator](options?: ReadableStreamIteratorOptions): AsyncIterableIterator<R>;
-    values(options?: ReadableStreamIteratorOptions): AsyncIterableIterator<R>;
-}
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -2479,6 +2479,13 @@
                             }
                         }
                     }
+                },
+                "iterator": {
+                    "type": {
+                        "0": {
+                            "overrideType": "R"
+                        }
+                    }
                 }
             },
             "ReadableStreamBYOBReader": {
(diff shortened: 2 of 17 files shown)


src/build/bcd/mapper.ts
```diff
@@ -84,6 +84,23 @@ function mapInterfaceLike(
   };
   const methods = filterMapRecord(i.methods?.method, recordMapper);
   const properties = filterMapRecord(i.properties?.property, recordMapper);
+
+  if (i.iterator) {
+    const iteratorKey = i.iterator.async ? "@@asyncIterator" : "@@iterator";
+    const iteratorCompat = mergeCompatStatements(
+      data[iteratorKey] ?? data["values"]
```
MattiasBuelens, line 91: In practice, pretty much all iterable interfaces have a `values` entry, but only some have `@@iterator` (e.g. [`MediaKeyStatusMap`](https://developer.mozilla.org/en-US/docs/Web/API/MediaKeyStatusMap#browser_compatibility)). So we *could* get away with always using `data["values"]`, but I don't know if that's very clear.

baselines/audioworklet.asynciterable.generated.d.ts
```diff
@@ -0,0 +1,3 @@
+/////////////////////////////
+/// AudioWorklet Async Iterable APIs
+/////////////////////////////
```
MattiasBuelens, line 3: There aren't any async iterable APIs inside audio worklets yet. Do we want to emit an empty file, or remove this file instead?

saschanaz: I'm okay with keeping this.

src/build/bcd/mapper.ts
saschanaz: Hmmm. I think it's worth filing an issue on BCD. For now let's live with this with a comment about the situation, and perhaps a link to a BCD issue.

inputfiles/overridingTypes.jsonc
```diff
@@ -2479,6 +2479,13 @@
                             }
                         }
                     }
+                },
+                "iterator": {
+                    "type": {
+                        "0": {
+                            "overrideType": "R"
```
saschanaz, line 2486: This seems redundant here.

src/build.ts
```diff
@@ -42,17 +42,26 @@ async function emitFlavor(
...
+  const iterators = emitWebIdl(exposed, options.global[0], "sync");
   await fs.writeFile(
     new URL(`${options.name}.iterable.generated.d.ts`, options.outputFolder),
     iterators
   );
+
+  const asyncIterators = emitWebIdl(exposed, options.global[0], "async");
+  await fs.writeFile(
+    new URL(
+      `${options.name}.asynciterable.generated.d.ts`,
```
saschanaz, line 60: I'm not a fan of having extra files, but I guess we need it 🥲

src/build/emitter.ts
```diff
@@ -1705,6 +1712,83 @@ export function emitWebIdl(
...
+      const comments = i.iterator!.comments?.comment;
+
+      methods.forEach((m) => {
+        emitComments({ comment: comments?.[m.name] }, printer.printLine);
+        printer.printLine(`${m.name}(${paramsString}): ${m.definition};`);
+      });
+    }
+
+    const subtypes = getAsyncIteratorSubtypes();
+    if (subtypes) {
```
saschanaz, line 1768: Can this return early instead?

saschanaz (commented):
That was quick, very cool! Thank you very much. Some nits:

inputfiles/overridingTypes.jsonc
MattiasBuelens: Fair enough. I'll need to remember to add it back once browsers *actually* add support for this async iterator, and it actually ends up in the generated types. 😛

saschanaz: Hmmmmm. That said, this is the only needed manual data and we know it will eventually come, so maybe we can keep this 👀

src/build/bcd/mapper.ts
MattiasBuelens: Looks like there's already such an issue: https://github.com/mdn/browser-compat-data/issues/6367.

I'll link to it in a code comment.

src/build/emitter.ts
MattiasBuelens: I tried to be consistent with [`emitIterator()`](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/fce78f39b24bd096483b96c446e6fc26d65ca086/src/build/emitter.ts#L1681-L1705), which doesn't use an early return either... 🤷‍♂️

saschanaz: I think it's equally readable? It's like "hey, no subtypes and no method with sequences? We are done here then"

saschanaz:
Thanks! LGTM.

PR #1416, opened 2022-10-03
add crypto.subtle.derivebits EcdhKeyDeriveParams length null overload

panva:
As per https://github.com/w3c/webcrypto/pull/345


saschanaz:
This doesn't make sense, the algorithm cannot get `length` as null when the IDL doesn't allow it. I'll file a bug there.

saschanaz:
Oh, you already did: https://github.com/w3c/webcrypto/issues/322

I'll wait for the response in that case.

PR #1425, opened 2022-10-30
fix: InsertPosition

HomyeeKing:
spec: https://dom.spec.whatwg.org/#dom-element-insertadjacentelement


saschanaz:
This removed something, perhaps `npm i` again before build?

saschanaz:
Thank you! LGTM.

PR #1431, opened 2022-11-10
Add CodeQL scanning

sandersn:
Copied from TS repo.


sandersn:
The scanner found 3 medium alerts, but they were all false positives: the recursive object merge already worked the recommended way, it just couldn't tell.

Also this code is run manually to generate files which are later inspected in a diff, so there's no way that Object.prototype pollution could cause a security problem.

PR #1432, opened 2022-11-19
More specific crypto.randomUUID return type

HarryAllen1:
Modifies the `crypto.randomUUID()` return type to 5 hyphen-separated strings.


HarryAllen1:

inputfiles/overridingTypes.jsonc
```diff
@@ -140,13 +140,13 @@
                 }
             },
             "HTMLOrSVGElement": {
-              "properties": {
-                "property": {
-                  "nonce": {
-                    "name": "nonce?"
-                  }
+                "properties": {
```
saschanaz, line 143: I think we'll once need to format this file correctly, but for now can you revert the formatting so that the PR can have a single purpose?

saschanaz (changes requested):

inputfiles/overridingTypes.jsonc
```diff
@@ -3426,4 +3433,4 @@
             }
         }
     ]
-}
+}
```
saschanaz, line 3436: EOF is accidentally removed 👀

HarryAllen1: No idea how that happened... should be fixed now.

saschanaz:
LGTM, thanks!

PR #1443, opened 2022-12-07
better constant definitions for Websocket, EventSource, and FileReader

k-yle:
The websocket spec defines the values of constants like `WebSocket.OPEN`, so they're never going to change. Defining them as literal numbers allows us to write more type-safe code. The same as true for `EventSource` and `FileReader`.

This [same change was made to the `@types/ws` package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ws/index.d.ts#L55-L62), which has been very helpful in the NodeJS world. 

I think it's unlikley that this will be a breaking change (?) since `0` and `1` etc. are a subset of `number`, but I could be wrong here. 

There was a discussion at https://github.com/microsoft/TypeScript/issues/25090 where an impossible solution to this problem was proposed (const enums).

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -1805,23 +1805,15 @@
                 "constants": {
                     "constant": {
                         "CONNECTING": {
-                            "name": "CONNECTING",
-                            "readonly": true,
                             "overrideType": "0"
                         },
                         "OPEN": {
-                            "name": "OPEN",
-                            "readonly": true,
                             "overrideType": "1"
                         },
                         "CLOSING": {
-                            "name": "CLOSING",
-                            "readonly": true,
                             "overrideType": "2"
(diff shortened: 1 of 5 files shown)


inputfiles/overridingTypes.jsonc
```diff
@@ -1801,6 +1801,51 @@
...
+                            "overrideType": "3"
+                        }
+                    }
+                }
+            },
+            "EventSource": {
+                "constants": {
+                    "constant": {
+                        "CONNECTING": {
+                            "name": "CONNECTING",
```
saschanaz, line 1834: Thank you! One thing, you don't need to repeat this name and readonly field, the overrideType should be sufficient.

k-yle: thanks, I've removed these fields

PR #1448, opened 2022-12-19
Emit constants with literal types

saschanaz:
Replacing #1443.

Undoing sort-by-name since it's more confusing this way. A caveat is that mixins may make incorrect order, but then mixins rarely have any constants.


saschanaz:
LGTM

Issue #1444, opened 2022-12-07
Add method `static Response.json(data, init)`


regseb:
Hi,

A new method has been added: [**`static Response.json(data, init)`**](https://fetch.spec.whatwg.org/#dom-response-json) to create a response that contains JSON (not to be confused with [`Response.prototype.json()`](https://fetch.spec.whatwg.org/#dom-body-json)).

- https://github.com/whatwg/fetch/issues/1389
- https://github.com/whatwg/fetch/pull/1392

```typescript
declare var Response: {
    // ...
    json(data: any, init?: ResponseInit): Response;
};
```
Support:

- :heavy_check_mark: Chromium [105](https://chromestatus.com/feature/5197912798658560)
- :x: Firefox: [_Implement `Response.json` static method_](https://bugzil.la/1758943) / [_D171217_](https://phabricator.services.mozilla.com/D171217)
- :x: Safari: [_Implement `Response.json` static method_](https://webkit.org/b/240375) / https://github.com/WebKit/WebKit/pull/10593
- :heavy_check_mark: Node via [Undici v5.3.0](https://github.com/nodejs/undici/releases/tag/v5.3.0)
  - node-fetch [v3.3.0](https://github.com/node-fetch/node-fetch/releases/tag/v3.3.0)
- :heavy_check_mark: Deno [1.22](https://deno.com/blog/v1.22#new-responsejson-static-method)
- :heavy_check_mark: Bun [0.0.74](https://github.com/oven-sh/bun/commit/394ec8dcbc862eaab7239c8592613c99f3eb87cd)

---

I found that in this project:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/513aae273e24603cdc14216e2d4111be96c8e564/inputfiles/removedTypes.jsonc#L83-L95

karlhorky:
@regseb Now that #1462 has been closed and [the `Response.json()` types will show up in TS 5.2](https://github.com/microsoft/TypeScript/issues/52841#issuecomment-1574002759), maybe this can be closed?

PR #1462, opened 2023-01-03
introduces Response.json() static method

feugy:
### 📖 What's in there?

A new method has been added: [static Response.json(data, init)](https://fetch.spec.whatwg.org/#dom-response-json) to create a response that contains JSON (not to be confused with [Response.prototype.json()](https://fetch.spec.whatwg.org/#dom-body-json)).

All the details are in #1444


feugy:
@microsoft-github-policy-service agree [company="Vercel"]

feugy:
@microsoft-github-policy-service agree company="Vercel"

inputfiles/removedTypes.jsonc
```diff
@@ -83,19 +83,6 @@
                     }
                 }
             },
-            "Response": {
-                "methods": {
-                    "method": {
-                        // BCD's api.Response.json is not about the static method but about Body#json.
```
HolgerJeromin, line 89: Only chrome/blink has this new API AFAIS. 
So it should not be part of typescript right now as we need two supporting engines.

feugy: Hello!.
It's not just Chrome: Node.js, Cloudflare, Deno, Vercel's Edge runtime also support this static method.
These built-in types can not be extended (Response is a variable, it can not be extended), so it's a big blocker here.

HolgerJeromin: https://github.com/microsoft/TypeScript-DOM-lib-generator#why-is-my-fancy-api-still-not-available-here

feugy: This is fair. But because the there is no "constructor" type to be extended, it is impossible to "augment" the constructor in user land (it's only possible to add instance methods, not static methods).

This is making it impossible for users the work around the limitation <picture data-single-emoji=":sad:" title=":sad:">:sad:</picture> 
Do you have any suggestion I could implement?

regseb: I suggested adding Node and Deno to the list of vendors: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1476. If this proposal is accepted, the addition of the `static Response.json(data, init)` method will be allowed.

karlhorky: Yeah, seems like since responses are more commonly generated on the backend, having multiple backend runtimes supporting this seems like it should be enough.

karlhorky: @feugy in case this doesn't get accepted because of this (unusual) restriction about two **browser** engines implementing this, I also tagged you in the Mozilla / Gecko Bugzilla tracking issue over here (because you apparently have a local version of Firefox with this implemented):

https://bugzilla.mozilla.org/show_bug.cgi?id=1758943#c1

baselines/dom.generated.d.ts
```diff
@@ -18736,6 +18736,8 @@ declare var Response: {
     new(body?: BodyInit | null, init?: ResponseInit): Response;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Response/error_static) */
     error(): Response;
+    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Response/json) */
```
regseb, line 18739: The https://developer.mozilla.org/docs/Web/API/Response/json page describes the `Response.prototype.json()` method. The `static Response.json(data, init)` method isn't yet documented in MDN. https://github.com/mdn/content/issues/21550


The URL will probably be: `https://developer.mozilla.org/docs/Web/API/Response/json_static`

feugy: Hello and thanks for the review. 
You're absolutely right, the right URL does not exist yet.
Unfortunately, using it (or removing the comment) fails the test suite.

feugy:
@HolgerJeromin, I've re-based this PR since this static method [has been implemented on Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1758943), and is planned in release 115.

Despite the fact that is it not implemented in the MDN documentation yet, it is fully specified [in the fetch spec](https://fetch.spec.whatwg.org/#dom-response-json).

I hope we can now consider adding it to Typescript types.

feugy:
Thanks a mil!

baselines/dom.generated.d.ts
hax: `https://developer.mozilla.org/docs/Web/API/Response/json_static` has been created last week.

Issue #71, opened 2016-02-25
Suggestion: Add String literal types support


YuichiNukiyama:
**TypeScript Version:** 

1.9.0-dev.20160225

**Code**

```ts

{
        "kind": "property",
        "interface": "HTMLTextAreaElement",
        "name": "autocomplete",
        "type": "\"on\" | \"off\""
}

```

**Expected behavior:**

generate following code.

```ts

interface HTMLTextAreaElement {
        autocomplete: "on" | "of";
}

```

**Actual behavior:** 

generate nothing.

DanielRosenwasser:
You're bitwise-ORing two strings. Did you try just writing this?

```ts
{
        "kind": "property",
        "interface": "HTMLTextAreaElement",
        "name": "autocomplete",
        "type": '"on" | "off"'
}
```

YuichiNukiyama:
Yes, I did. But, resuilt was same.

zhengbli:
I tried your code snippet, and it worked find for me. Did you add the snippet in `addedTypes.json`?

YuichiNukiyama:
Thank you good guys.

PR #1467, opened 2023-01-09
 Add more precise typing for autocomplete HTML attribute

yanndinendal:
This will allow to autocomplete… the `autoComplete` attribute in vscode for example.

autoComplete autocompletion in vscode

The list is quite long and it's not easy remembering the spelling of each one of them (first_name or given-name? password or current-password?), and it would improve the UX of sites if they were more appropriately used (for example, some people may think that the input's `type` and/or `name` attributes may be enough for fields like email or password, but knowing to use `autoComplete="new-password"` or `one-time-code` can be really useful).

---

~~I tried declaring a `type AutoComplete` to avoid repetition but didn't find how to do that:
[microsoft/TypeScript#52169](https://github.com/microsoft/TypeScript/pull/52169/files).~~

See updated proposal at https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1467#issuecomment-1378524560

**Fixes microsoft/TypeScript#52168.**
See also #71.


yanndinendal:

yanndinendal:

saschanaz:
First, please define an enum (which then generates a union type) and use that.

Second, are we sure every item here is well supported? Poorly supported items shouldn't be added.

HolgerJeromin:
Typescript team should be confident that they want `(string & {})` in this repo which uses a side effect of current typescript intellisense code.
If yes, this should perhaps be done for all attributes instead of one (the data should be available afaik).

yanndinendal:
@saschanaz: thanks, fixed (I didn't know how to do it with these json files, at first).

I've re-read the spec and declared an `AutoFill` enum that matches the spec, and I've tested most of them and confirmed they seem indeed supported in at least Chrome or Firefox.
I couldn't find definite info on Can I Use or MDN except for some of them like `new-password` which is largely supported:
https://caniuse.com/mdn-html_global_attributes_autocomplete_new-password

I've also fixed the possible values for `autocomplete` on a form element, which can't act as an autofill element, and confirmed it's supported on textareas and selects:
- https://caniuse.com/mdn-api_htmltextareaelement_autocomplete
- https://caniuse.com/mdn-api_htmlselectelement_autocomplete

---

I added `webauthn`, which is in the spec (linked in the comments) and can be used in combination with `username` or `current-password` values (already available in stable releases of Chrome):
https://web.dev/passkey-form-autofill/

---

Unfortunately, I don't know how to add typescript auto-completion for a list of strings (`space-separated tokens`, like `"username webauthn"` or `"shipping street-address"`), but it should already cover most use-cases (and more advanced ones are handled with `(string & {})`).  
@HolgerJeromin: So it's necessary to keep it instead of a strict list to follow the spec (it also prevents this to be a breaking change).

Also note that this syntax/workaround is already used and tested in the TypeScript repo,   
such as with `tests/baselines/reference/specialIntersectionsInMappedTypes.types`'s  
`type Alignment = (string & {}) | "left" | "center" | "right";`   
or `tests/lib/react18/react18.d.ts`'s `AriaRole`.

saschanaz:
What does `(string & {})` do, why not just `string`?

It's also currently broken, please use `AutoFill | string` instead for `overrideType`.

yanndinendal:
@saschanaz:

#### With `| (string & {})`:
autoComplete autocompletion in vscode

#### Without:
no editor auto-completion.

---

See microsoft/TypeScript#33471, and [microsoft/TypeScript#29729](https://github.com/microsoft/TypeScript/issues/29729#issuecomment-567871939) for an explanation of the issue.

I think I originally stumbled on this method in React's typedefs.

yanndinendal:
> It's also currently broken, please use `AutoFill | string` instead for `overrideType`.

@saschanaz: Sorry I don't understand that last part, the tests pass and it seems to work.  
By broken, do you mean that it should generate something else? Or that some cases (not handled by the tests) will not work?

I can't use `| string` or we will lose autocompletion (it would have type `string` only).

saschanaz:
Look at the output: `type AutoFill = "(string & {})"`. You don't want this 😛

yanndinendal:
Do you see the right commit? I see the full union:

image

yanndinendal:
Sorry I got it, the rest of the union is ok but `"(string & {})"` should not be quoted :o

I don't know how to do that :/

RyanCavanaugh:
Do not add `string & {}` to this file.

If the set of values is truly unconstrained, `string` is the appropriate type, and is what should be written here.

saschanaz:
Yup, that's also what other properties do.

yanndinendal:
>If the set of values is truly unconstrained, string is the appropriate type, and is what should be written here.

It's not unconstrained. Writing `string` or any union ` | string` wold lose any auto-completion, so we might as well close this PR if that's not something you want. Can I ask why?

I could try writing it as a [string template literal type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) instead, based on [this table from the spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#determine-a-field's-category), to allow 1, 2, 3 or 4 tokens from the union, in the correct order, but it would take me some time to do it right so we may want to first discuss if this could be investigated/accepted, or if it wouldn't be accepted either.

yanndinendal:
> Typescript team should be confident that they want `(string & {})` in this repo which uses a side effect of current typescript intellisense code.

just to clarify that this is to work around a bug in typestring; I would be happy to use `| string` instead once it's fixed, but in the meanwhile it would be no better than closing this PR and losing type suggestions.

See issues linked in [#issuecomment-1377304118](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1467#issuecomment-1378404272) for explanations.

yanndinendal:
Re [#issuecomment-1378404272](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1467#issuecomment-1378404272)
> I could try writing it as a [string template literal type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) instead, based on [this table from the spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#determine-a-field's-category), to allow 1, 2, 3 or 4 tokens from the union, [in the correct order](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-detail-tokens)

Here are the full constraints, in the right order, for the [set of space-separated tokens](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#set-of-space-separated-tokens):

```md
- `section-${string}`
- "shipping" | "billing"
- AutoFill // (the union I added to this PR, excluding "webauthn")
- "webauthn"
```

yanndinendal:
Ok, here is how it would look like, I just need to list all autorized permutations, and declare these types in json files but I made it work locally with a standard .d.ts file and it's quite impressive what it can suggest!

```ts
  type AutoFillSection = `section-${string}`;
  type AutoFillAddressKind = 'shipping' | 'billing';
  type AutoFillNormalField =
    | 'name'
    | 'honorific-prefix'
    | 'given-name'
    | 'additional-name'
    | 'family-name'
    | 'honorific-suffix'
    | 'nickname'
    | 'username'
    | 'new-password'
    | 'current-password'
    | 'one-time-code'
    | 'organization-title'
    | 'organization'
    | 'street-address'
    | 'address-line1'
    | 'address-line2'
    | 'address-line3'
    | 'address-level4'
    | 'address-level3'
    | 'address-level2'
    | 'address-level1'
    | 'country'
    | 'country-name'
    | 'postal-code'
    | 'cc-name'
    | 'cc-given-name'
    | 'cc-additional-name'
    | 'cc-family-name'
    | 'cc-number'
    | 'cc-exp'
    | 'cc-exp-month'
    | 'cc-exp-year'
    | 'cc-csc'
    | 'cc-type'
    | 'transaction-currency'
    | 'transaction-amount'
    | 'language'
    | 'bday'
    | 'bday-day'
    | 'bday-month'
    | 'bday-year'
    | 'sex'
    | 'url'
    | 'photo';
  type AutoFillContactKind = 'home' | 'work' | 'mobile' | 'fax' | 'pager';
  type AutoFillContactField =
    | 'tel'
    | 'tel-country-code'
    | 'tel-national'
    | 'tel-area-code'
    | 'tel-local'
    | 'tel-local-prefix'
    | 'tel-local-suffix'
    | 'tel-extension'
    | 'email'
    | 'impp';
  type AutoFillField =
    | AutoFillNormalField
    | AutoFillContactField
    | `${AutoFillContactKind} ${AutoFillContactField}`;
  type AutoFillCredentialField = 'webauthn';

  type AutoFill =
    | 'off'
    | 'on'
    | AutoFillField
    | `${AutoFillSection} ${AutoFillField}`
    | `${AutoFillAddressKind} ${AutoFillField}`
    | `${AutoFillSection} ${AutoFillAddressKind} ${AutoFillField}`
    | `${AutoFillSection} ${AutoFillAddressKind} ${AutoFillField} ${AutoFillCredentialField}`
    | `${AutoFillAddressKind} ${AutoFillField} ${AutoFillCredentialField}`
    | `${AutoFillSection} ${AutoFillField} ${AutoFillCredentialField}`
    | `${AutoFillField} ${AutoFillCredentialField}`;
```

HolgerJeromin:
> Also note that this syntax/workaround is already used and tested in the TypeScript repo,
such as with tests/baselines/reference/specialIntersectionsInMappedTypes.types's
type Alignment = (string & {}) | "left" | "center" | "right";
or tests/lib/react18/react18.d.ts's AriaRole.

Internal tests are different from code which is shipped to many developers and shapes the way they work.

yanndinendal:
Hi @RyanCavanaugh, @saschanaz: I pushed a new version without `| (string & {})` that is now fully spec-compliant. I would appreciate any feedback. :)

See https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1467#issuecomment-1378524560 for screenshots and context.

yanndinendal:
Up? 🙏

saschanaz:
Again, only the well supported features should be added. Please check the implementation status. Probably can be checked in:

* Gecko: https://searchfox.org/mozilla-central/source/dom/base/AutocompleteFieldList.h
* WebKit: https://searchfox.org/wubkat/source/Source/WebCore/html/Autofill.cpp
* Blink: https://source.chromium.org/chromium/chromium/src/+/main:components/autofill/core/common/autocomplete_parsing_util.cc;l=1?q=autocomplete_parsing_util.cc&sq=

baselines/dom.generated.d.ts
```diff
@@ -18550,6 +18553,12 @@ type AudioContextLatencyCategory = "balanced" | "interactive" | "playback";
 type AudioContextState = "closed" | "running" | "suspended";
 type AuthenticatorAttachment = "cross-platform" | "platform";
 type AuthenticatorTransport = "ble" | "hybrid" | "internal" | "nfc" | "usb";
+type AutoFillAddressKind = "billing" | "shipping";
+type AutoFillBase = "off" | "on";
+type AutoFillContactField = "email" | "impp" | "tel" | "tel-area-code" | "tel-country-code" | "tel-extension" | "tel-local" | "tel-local-prefix" | "tel-local-suffix" | "tel-national";
```
yanndinendal: `"tel-extension"` is unsupported in Gecko but supported in the two others; kept.

baselines/dom.generated.d.ts
```diff
@@ -18550,6 +18553,12 @@ type AudioContextLatencyCategory = "balanced" | "interactive" | "playback";
...
+type AutoFillContactKind = "fax" | "home" | "mobile" | "pager" | "work";
+type AutoFillCredentialField = "webauthn";
```
yanndinendal, line 18560: Not yet supported in Gecko but supported in all other engines, and critical for Passkeys support; kept.

baselines/dom.generated.d.ts
```diff
@@ -18550,6 +18553,12 @@ type AudioContextLatencyCategory = "balanced" | "interactive" | "playback";
...
+type AutoFillNormalField = "additional-name" | "address-level1" | "address-level2" | "address-level3" | "address-level4" | "address-line1" | "address-line2" | "address-line3" | "bday" | "bday-day" | "bday-month" | "bday-year" | "cc-additional-name" | "cc-csc" | "cc-exp" | "cc-exp-month" | "cc-exp-year" | "cc-family-name" | "cc-given-name" | "cc-name" | "cc-number" | "cc-type" | "country" | "country-name" | "current-password" | "family-name" | "given-name" | "honorific-prefix" | "honorific-suffix" | "language" | "name" | "new-password" | "nickname" | "one-time-code" | "organization" | "organization-title" | "photo" | "postal-code" | "sex" | "street-address" | "transaction-amount" | "transaction-currency" | "url" | "username";
```
yanndinendal, line 18561: `additional-name` is unsupported in Gecko only; kept.

yanndinendal: `address-level4` is unsupported in Blink only; kept.

yanndinendal: `bday-day`,  `bday-month`, `bday-year` is unsupported in Gecko only; kept.

yanndinendal, line 18561: `bday` (full date) is only supported in Webkit: keep or remove?

yanndinendal: `cc-additional-name` is only supported in Webkit: keep or remove?

yanndinendal: `cc-given-name`, `cc-additional-name`, `cc-family-name`, `cc-csc`, `cc-type`
are unsupported in Gecko only; kept.

yanndinendal, line 18561: `one-time-code` seems to be only supported in Chrome, according to code search,  
but I highly doubt it wouldn't work in Safari iOS.  
I must have missed it in the code, I've seen it used and have seen articles [like this about it](https://medium.com/@lamalamaMark/one-time-security-code-in-ios-safari-27869e16085f).

Kept.

yanndinendal: Similarly, I don't see `current-password` or `new-password` [in Blink's code](https://source.chromium.org/chromium/chromium/src/+/main:components/autofill/core/common/autocomplete_parsing_util.cc;l=1?q=autocomplete_parsing_util.cc), but it does work in Chrome without a doubt.

Kept.

yanndinendal, line 18561: There is a bug / typo in Gecko, where `honorific-*` are spelled `honorifix-*` like Asterix characters. :)
But both are meant to be supported in Gecko. I will open a bugzilla issue.

`honorific-prefix` is supported in all other engines.

`honorific-suffix` is missing in Blink.

Kept both.

yanndinendal: `language` seems to be only supported in Webkit: keep or remove?

yanndinendal: `nickname` is only supported in Webkit: keep or remove?

yanndinendal: `organization-title` is only supported in Webkit: keep or remove?

yanndinendal: `transaction-amount` and `transaction-currency` are unsupported in Gecko only; kept.

yanndinendal (commented):
@saschanaz: thank you very much for the links; here is everything I found;
let me know if I should remove some other tokens that are only supported in one browser engine.

saschanaz:
Great research, I think things with only one engine support shouldn't be included.

saschanaz:
Also, this kind of complex type can take some benefit from unit tests.

baselines/dom.generated.d.ts
```diff
@@ -18550,6 +18553,12 @@ type AudioContextLatencyCategory = "balanced" | "interactive" | "playback";
 type AudioContextState = "closed" | "running" | "suspended";
 type AuthenticatorAttachment = "cross-platform" | "platform";
 type AuthenticatorTransport = "ble" | "hybrid" | "internal" | "nfc" | "usb";
+type AutoFillAddressKind = "billing" | "shipping";
+type AutoFillBase = "off" | "on";
+type AutoFillContactField = "email" | "impp" | "tel" | "tel-area-code" | "tel-country-code" | "tel-extension" | "tel-local" | "tel-local-prefix" | "tel-local-suffix" | "tel-national";
+type AutoFillContactKind = "fax" | "home" | "mobile" | "pager" | "work";
```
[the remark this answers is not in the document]
yanndinendal: To remove

baselines/dom.generated.d.ts
yanndinendal: Remove all these

yanndinendal: https://bugzilla.mozilla.org/show_bug.cgi?id=1814735

yanndinendal:
@RyanCavanaugh: all good? :)

yanndinendal:
@saschanaz: Let's merge this?

saschanaz:
I need some time to verify whether those indeed match the requirement. Will do it soonish...

inputfiles/addedTypes.jsonc
```diff
@@ -1374,6 +1463,21 @@
...
+            {
+                "name": "AutoFillField",
+                "overrideType": "AutoFillNormalField | AutoFillContactField | `${AutoFillContactKind} ${AutoFillContactField}`"
+            },
+            {
+                // See the full list of possible autofill values for the `autocomplete` attribute:
+                // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#determine-a-field's-category
+                // Full spec at https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill.
+                "name": "AutoFill",
+                "overrideType": "AutoFillBase | AutoFillField | `${AutoFillSection} ${AutoFillField}` | `${AutoFillAddressKind} ${AutoFillField}` | `${AutoFillSection} ${AutoFillAddressKind} ${AutoFillField}` | `${AutoFillSection} ${AutoFillAddressKind} ${AutoFillField} ${AutoFillCredentialField}` | `${AutoFillAddressKind} ${AutoFillField} ${AutoFillCredentialField}` | `${AutoFillSection} ${AutoFillField} ${AutoFillCredentialField}` | `${AutoFillField} ${AutoFillCredentialField}`"
```
saschanaz, line 1480: Hmm, so per the spec it's something like:

`AutofillSection? AutofillAddressKind? (AutoFillNormalField | (AutoFillContactKind? AutoFillContactField)) AutoFillCredentialField?`

And this is the result of spreading it to every possible combination.

It's hard to read and maintain, I wonder this could be better as the type system is very versatile nowadays. 🤔

saschanaz: How about this?

```ts
type OptionalPrefixToken<T extends string> = `${T} ` | ""
type OptionalPostfixToken<T extends string> = ` ${T}` | ""

// (AutoFillNormalField | (AutoFillContactKind? AutoFillContactField))
type AutoFillField = AutoFillNormalField | `${OptionalPrefixToken<AutoFillContactKind>}${AutoFillContactField}`;

// AutofillSection? AutofillAddressKind? AutoFillField AutoFillCredentialField?
type AutoFill = `${OptionalPrefixToken<AutoFillSection>}${OptionalPrefixToken<AutoFillAddressKind>}${AutoFillField}${OptionalPostfixToken<AutoFillCredentialField>}`
```

inputfiles/addedTypes.jsonc
```diff
@@ -1374,6 +1463,21 @@
             {
                 "name": "EventListenerOrEventListenerObject",
                 "overrideType": "EventListener | EventListenerObject"
+            },
+            {
+                "name": "AutoFillSection",
+                "overrideType": "`section-${string}`"
```
saschanaz, line 1469: This is problematic as it allows space: `section-foo bar baz` is allowed as `AutoFillSection`.

wszydlak:
Which specification do you use to get enum values for `autocomplete`? I'm asking cause https://html.spec.whatwg.org/multipage/input.html#htmlinputelement does not define enum values for autocomplete, instead it defines `string`.

saschanaz:
> Which specification do you use to get enum values for `autocomplete`? I'm asking cause https://html.spec.whatwg.org/multipage/input.html#htmlinputelement does not define enum values for autocomplete, instead it defines `string`.

It [gives more details](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete) if you scroll up and click `autocomplete — Hint for form autofill feature` under "Content attributes".

>When wearing the [autofill expectation mantle](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-expectation-mantle), the [autocomplete](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete) attribute, if specified, must have a value that is an ordered [set of space-separated tokens](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#set-of-space-separated-tokens) consisting of either a single token that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "[off](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-off)", or a single token that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "[on](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-on)", or [autofill detail tokens](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-detail-tokens).

wszydlak:
@saschanaz 
Thanks :)

yanndinendal:
Thanks for the detailed review, I'm looking into it. :)

inputfiles/addedTypes.jsonc
yanndinendal: Honestly it doesn't seem trivial, I don't think I'm fluent enough in typescript to do this (or if it's possible at all).

I will add tests that show the expected and unexpected cases anyway.

I'll continue looking into it but I would argue that the current type (before this PR) allows much more invalid values, so maybe we could iterate on it? Or do you feel like it would be a blocker?

yanndinendal: I can see how to do it with a generic, but not inside a template literal type:

[playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggrsA9gMQJYBt0DlECcC2AhumhOgCZQC8UARIeeasKogHbEC0H+EtUAHzoNyuCAGdxndBABuZAIz8h9RmMnS5ZAEzLhaiVJnz0AZj2rRhzSYAsFkeqOo2EJYP1WN6FxF0fLJ2lfcwCAI3JCEE5IkAsIqM58dmAAC3jYzhAIQlwLAGN8znzxfIKiiAAPMHLOKrAklPSAwrrqrJy8lqKAM0J8DGiePm7OAHNUeTZufpGVVuHatjh8MIgu+aLQSALEODZgXDiWvYOjmd4CuFwxA84wQkkAdzxyCz6B9CHZiwmpi7mdFS7DwqB6qCKYDE4MqFmBbFB4KK4jgPRhFkWAVcT3uj3EL1wbwC7AgnBYvGKiHIgNoeDGhDYqAAXoQWOwLGBEOJgFx8lSadyxBBgJxHIYLIcGeJCPk2dN+qdgBLcFKZXLitdbvljio4OJ1osANwAWAAUNtoPAkGhMABhFJqgDSLkoNFowMuAWSYQwNIJAGtaCbzeBLQgUBh0PaDmrSBRqHQIEQMBKyKn0KKxIRKdT05T9odonzcwFgGQ2mW2OJWGw8+hEPliHWG1woRB0aXy-XGxmUWjULDOxmOHKm8GLbBwzb0DADJJnWxXXQfZgXGMLOJUqgwGA10GzROrRG7WJqQdUMQ40vaE8IGFCAhUrXg2azQB6N9QTjf1+mg+hqAcAAZQefIICAiBZRrAAeAAVKAqkrchxCgQU1wAPgTeDEIgRcUIAA31KD2E4AASABvFwenWKAACUJGAABffCzSgNioAAfjohiEMqJCCIotC2DGRioEEw412Y1j2K41x5FwaS2IALigWDFKgFS5PWF9TVkXIoEIFTgNA8DILlaCAHIiPVHpEEQe9cAszCaCssya04Wz7NyCzgw-KAAAFgCkeozLqG48DNPTcCgMIjMQECZVM4i2Es6z3M8mLvOcqBXOSjy7MyxydMPKdIwg5KE2MxLyvM-DxNwST8PQnSoqgfIVKPacaprBNcpsuyHJ8s18nfT9AuC6pQvWXAIt0-TyA60rMG69herSkiMsG4NyF-Pzv04X8JwAeTAUd0AABWhAdYMQf1cLg3j+NQiThOyuryNg0T8ICWh-0gKATrO86uWAGEbrulLsL43DkOehrXoTb6KM+76VF+v9TT8gAKTrIxwAhL1QMhKCEHGlqjB1ZQXcguNxu1KeAK8AEomb+sNrUjK8Ezp7A8CIEgifjIR3sBmtiEu9trtu+6eejHkqZddDGIo2WGavZidL8o9wWWtz2Fp8MdZnOdxGpg2OcwLnZdPXCWEJ4mOLZycLfQRGKNF9hxauypwfuqqwJWlKecD9Clfd06xYu73faD8nZy8U3FeV8iefV8OgZBsHpdjl3bRt897YoJWWIx0avx-UvMbGoK2kgWUwpmhS5uiuBFpdta9emTb9L1A1ZiG004Ei-SQDb49XZc9au4K3vcGGCyzRAX9WsQMfp162f57L8ba6m8Km9asA18jDf9Wi+egA)

yanndinendal: Yes, pretty sure it's not currently possible. Anyway as I said, still much less false positives than before. :)

inputfiles/addedTypes.jsonc
saschanaz: That seems not problematic for me. Someone else from Microsoft may disagree, though.

inputfiles/addedTypes.jsonc
saschanaz: >Anyway as I said, still much less false positives than before. :)

While I admit, it may give false impression that any match would be legal when it's not, which may be more harmful.

```ts
// Has typos but no error, because AutoFillSection greedily matches the typos :(
var illegal: AutoFill = 'section-typescript billin hom tel-local'
```

Perhaps it's okay to have some helper generics? I want to hear from Microsoft people.

inputfiles/addedTypes.jsonc
sandersn: I think it's fine to show the union instead of the alias.

sandersn:
Note: I can approve this PR for Ryan once @saschanaz also approves it.

inputfiles/addedTypes.jsonc
saschanaz: @sandersn This one is the main reason I'm not approving, what do you think?

sandersn: I think some overgeneration is OK as long as it's not rejecting values that are legal.
I do wonder about performance though. Let me look again at the generated size of the type.

inputfiles/addedTypes.jsonc
yanndinendal: Thanks, I'll try to commit this when I get some time (I've broken my arm ^^).

saschanaz: Oh no, take care 🤕

yanndinendal: @saschanaz: done. :)

Thanks, I agree it's much more readable and maintainable like that.

inputfiles/addedTypes.jsonc
yanndinendal: > This is problematic as it allows space: `section-foo bar baz` is allowed as `AutoFillSection`.

Note for other readers that it is not a full catch-all, you still need to add valid tokens to your attribute, so for example, `section-foo bar baz username` would pass but not `section-foo bar baz` alone, nor `section-wrong off` as shown in the tests.

yanndinendal: > I do wonder about performance though. Let me look again at the generated size of the type.

@sandersn: Hi, I've pushed the final version, are you ok with the generated types?

https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1467/files#diff-d57c51d024219b9f37e6b42df127db25f700acbaade23ce1edcecda2853b995c

yanndinendal:
Conflicts fixed

yanndinendal:
@sandersn, @saschanaz: LGTY?

saschanaz:
I was mostly waiting for https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1467#discussion_r1139389699 and also myself to check the support status again. Something like https://mdn-bcd-collector.appspot.com/ but for autofill would be useful.

sandersn:
I turned the types into values and got 424:

```js
var OptionalPostfixToken = _ => 2
var OptionalPrefixToken = _ => 2
var AutoFillSection = 1
var AutoFillAddressKind = 2
var AutoFillBase = 2
var AutoFillContactField = 9
var AutoFillContactKind = 3
var AutoFillCredentialField = 1
var AutoFillNormalField = 36
var AutoFillField = AutoFillNormalField + OptionalPrefixToken(AutoFillContactKind) * AutoFillContactField;
var AutoFill = AutoFillBase + OptionalPrefixToken(AutoFillSection)*OptionalPrefixToken(AutoFillAddressKind)*AutoFillField*OptionalPostfixToken(AutoFillCredentialField)
console.log(AutoFill)
```

I think that's reasonable; any sizeable addition to the DOM adds that many types.

sandersn (approved):

inputfiles/addedTypes.jsonc
saschanaz: Can you add a TODO comment about the over-match of `section-{string}`?

unittests/files/autocomplete.ts
```diff
@@ -0,0 +1,59 @@
+// @ts-expect-error should be a string
+document.body.getElementsByTagName("input")[0].autocomplete = false;
+// @ts-expect-error should not be empty
+document.body.getElementsByTagName("input")[0].autocomplete = "";
```
saschanaz, line 4: Looks like it can be empty: https://github.com/web-platform-tests/wpt/blob/606530b23a2ed89e965403475c6f40a98afb0900/html/semantics/forms/the-form-element/form-autocomplete.html#L63-L67

And also per the step 32 of the spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-processing-model

>Default: Let the element's [IDL-exposed autofill value](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#idl-exposed-autofill-value) be the empty string, and its [autofill hint set](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-hint-set) and [autofill scope](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-scope) be empty.

inputfiles/addedTypes.jsonc
```diff
@@ -352,6 +352,95 @@
...
+                    "honorific-prefix",
+                    "given-name",
+                    "additional-name",
+                    "family-name",
+                    "honorific-suffix",
+
+                    "username",
+                    "new-password",
+                    "current-password",
+                    "one-time-code",
```
saschanaz, line 385: A comment about iOS or mentioning Apple doc would be helpful to understand why this is here although WPT says no.

saschanaz (commented):
Sorry for the long delay, and thank you for maintaining this so far!

Looks good with some nits. It seems a bit more fields are supported now than this was written, but asking that change here would be very unfair, that can be done separately.

saschanaz:
I'm assuming that having the Optional*Token helper types here is okay, as @sandersn hasn't complained about that.

yanndinendal:
@saschanaz: pushed, thanks! :)

saschanaz (approved):

saschanaz:
Let's see I can merge this even if there's a change request, because that's solved now.

LGTM

yanndinendal:
@sandersn, @saschanaz: Hi, do you when it will be included in https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts and released? 😄

saschanaz:
(I'm not really allowed to know such info, but the roadmap implies the new release is coming soonish, as the release happens roughly every three months and the last release was three months ago.) https://github.com/microsoft/TypeScript/wiki/Roadmap)

saschanaz:
If you need it today, consider using https://www.npmjs.com/package/@types/web.

... except it looks like the release step is broken since 0.0.99.

sandersn:
> I'm assuming that having the Optional*Token helper types here is okay, as @sandersn hasn't complained about that.
Basically yes. I'll check the performance numbers when I copy the types into TS 5.2 beta tomorrow but I think it'll be OK.

I'll check out the publication failure and try to fix it.

yanndinendal:
Unfortunately, this does not get picked up by `@types/react` / `@types/react-dom` JSX `autoComplete` prop. `¯\_(ツ)_/¯`

Asking for instructions here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66982

gavinwahl:
I think this type is too strict. All the tokens should be matched case-insensitively, so `autoComplete="NeW-PASSworD"` needs to be allowed. 

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill-detail-tokens
> a token that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for one of the following strings:
> ...

HolgerJeromin:
> I think this type is too strict. All the tokens should be matched case-insensitively, so autoComplete="NeW-PASSworD" needs to be allowed.

In general typescript does not allow all valid and spec'ed code but try to give hints to best practice while still being useful.
Most setters designed for strings in browser APIs allows for example any content which will be converted to string internally. 
Still TS forces the string datatype.

I think this is a similar (but less strong) decision.

Beside the point that typescript IMO does not have the concept for string types needed here.

saschanaz:
Indeed, there's no case-insensitive string type.

gavinwahl:
I think `string` is the only reasonable option here.

PR #1470, opened 2023-01-10
Change input event from Event to InputEvent

swimauger:
`input` event emits an `InputEvent`, this is more specific

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5766,7 +5766,7 @@ interface GlobalEventHandlersEventMap {
     "focusout": FocusEvent;
     "formdata": FormDataEvent;
     "gotpointercapture": PointerEvent;
-    "input": Event;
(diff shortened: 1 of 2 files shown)


swimauger:
> 

saschanaz:
This is intentional. See also: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1174#issuecomment-933042088

swimauger:
I see, in that case can we imply the type based on `HTMLInputElement` `type`, something like this? Although, this maybe more trouble than it's worth, since we can just type cast 😄 

Screen Shot 2023-01-21 at 5 10 05 PM

saschanaz:
You can define one yourself, but I can't imagine we have it here since `.type` can freely change and thus the assumption will be fragile.

saschanaz:
I think it's explained enough...

@github-actions close

PR #1474, opened 2023-01-14
Add overloads for specific context types to OffscreenCanvas.getContext

StenAL:
Each context type can only return the relevant context or null.

There are already similar overloads in place for HTMLCanvasElement.getContext.

Relevant part of the HTML spec: https://html.spec.whatwg.org/multipage/canvas.html#dom-offscreencanvas-getcontext


StenAL:

saschanaz:
LGTM, thanks!

chanya2534:
> Merging because @saschanaz is a code-owner of all the changes - thanks!

PR #1475, opened 2023-01-15
FileReader.readyState: use FileReader state constants as possible values

StenAL:
The spec doesn't allow any other values besides these constants for readyState.

See https://w3c.github.io/FileAPI/#dom-filereader-readystate

Not sure if this is the preferred way to narrow the type, I couldn't find any other examples using this pattern. Since this declaration uses constants instead of enums, it shouldn't break any consuming code. The following still compiles:
```ts
fileReader.readyState === 2;
declare const someNumber: number;
fileReader.readyState === someNumber;
```

While using an invalid value here no longer does:
```ts
fileReader.readyState === 5 // TS2367: This comparison appears to be unintentional because the types '0 | 1 | 2' and '5' have no overlap.
```


saschanaz:
LGTM

Indeed this is new, but I think it's good to have. A bit verbose but it gives reference to the constants.

PR #1484, opened 2023-01-25
Add support for `WEBGL_compressed_texture_pvrtc`

SuperSodaSea:
In mdn/browser-compat-data#18692 and mdn/browser-compat-data#18704 the BCD for `WEBGL_compressed_texture_pvrtc` is updated. With the latest BCD there are two major browser engines (Blink & WebKit) that support `WEBGL_compressed_texture_pvrtc`, so we can add interface `WEBGL_compressed_texture_pvrtc` and `getExtension(extensionName: "WEBGL_compressed_texture_pvrtc")` now.


SuperSodaSea:

SuperSodaSea:
Seems that some problems need to be solve in #1481 and mdn/browser-compat-data#18710. I will wait for them to be solved.

saschanaz:
BCD is updated, please merge.

SuperSodaSea:
Done!

saschanaz:
Please don't update package.json here as doing so requires another review which happens not so frequently.

SuperSodaSea:
Okay, changes in package.json are reverted.

saschanaz:
Thanks! LGTM

Issue #1237, opened 2022-01-01
Support for `structuredClone`


cawa-93:
New DOM api `structuredClone` should be added.

MDN: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

saschanaz:
Thanks for the ping. It will be automatically added when it gets multiple implementations.

yume-chan:
Note: `@mdn/browser-compat-data@4.1.10` now has enough browser implementations for `structuredClone` (but it breaks other parts)

saschanaz:
Yeah, #1282. 😬

HolgerJeromin:
fixed with #1283

saschanaz:
>but it need an override to become generic:

I guess this one still needs to be done

pragmat1c:
Any progress here?

saschanaz:
PRs welcome.

ehoogeveen-medweb:
FWIW, one issue I ran into with my own generic version is that if the source object is (deeply) `readonly` (I had a complex template object defined with `as const` to get good type information), the result of the structured clone is _also_ considered (deeply) `readonly`.

In my case, the whole point was to clone the template object a few times and then write to the clones, so I had to add a hacky `DeepMutable<T>` to make it work. I don't know whether it _always_ makes sense for `readonly` to be stripped as the result of cloning, but simply specifying that the output type match the input type might not be ideal either.

GabrielDelepine:
Any chance we could get a typed returned value in the future?

Something like `declare function structuredClone(value: T, options?: StructuredSerializeOptions): T;` ?

I know that `structuredClone` is not returning a strictly identical object (functions are silently removed for instance).
Maybe that's the reason to use `: any` here?

GabrielDelepine:
My current "workaround" (let me know what you think)

```typescript
type MyType = {
  ok?: boolean
}

const foo: MyType = {}

const bar =  structuredClone(foo) as typeof foo

bar.ok = true // OK
bar.something = 'else' // Typescript error, because bar belong to MyType
```

cawa-93:
This type declaration works for me:
```ts
type Cloneable<T> = T extends Function | Symbol
  ? never 
  : T extends Record<any, any> 
    ? {-readonly [k in keyof T]: Cloneable<T[k]>}
    : T

declare function structuredClone<T>(value: Cloneable<T>, options?: StructuredSerializeOptions | undefined): Cloneable<T>
[...]

PR #1520, opened 2023-03-12
Make structuredClone generic

ConorLinehan:
Looks to addresse [this issue](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1237).

Creates a generic for structuredClone.

[Playground with examples](https://www.typescriptlang.org/play?ssl=3&ssc=3&pln=1&pc=1#code/C4TwDgpgBAsgrsAhgIwDYQDwBUB8UC8UA3gFBRQC0AThIgCYD2AdqiFANoAKUAlk1AGsIIBgDMoWALoAuCV0kBuEgF8lJPsAhVRiAMbQA6n0YB3YmXKWAzsCpxdwODToBhVM0xYCURExA4ACgA3RFQ4CFksABooBjBgHmYrAH5ZAGVbe0dnNK0eUJ4ALwgAeXjEpisASkilZRISUEgoGnpmVgBRAA9EAFswdG8AJVpGFhAMUnImPoioGyo+AHMlckQluaY4XuQtOpwG3STgKAge-vRZVrHO84HoQimoGd65gHJNGzeoi3W5gGYAAwqNRHSonXTuJgPKAmYwMEwAOgWWScEFcUIgATOfXuVSUAHoCVAAALAKwUM6QByUqhUBhUZ7MKC9BAodAkSEeRF-bwAJgAnKDjiy2WgIG4PN44UxTMjMg40RiPBh4EhxRhru0QN1cegcIEudD8SRWer0JLoTyNvyhQ0iVA0gALBhwVB0U50hmnO7oKyIw4ipjMXKLRCFdkwp6iOBMXSyAJVAh4Ij1eokB1kilUiA0rT0xkAOQYJyseQKEfFnJF+YZlphMrlKMVznrAWDTFD+Ur6HxUAdRlQqE9Bdi-DsTASr3txKwVF8VlEWnRsWQACtc+SZ1AXK1ND5+ABBOmIEAAITgoiXjLhwCdPnmRWgfCgyBAn2r4Nfl+vAEZvNCZjHvO54-loAS-gAbCan42KuG4OP+jwWMgYFULIqFXlov4-Ko271lAd7QAw66blAYJIHwyyEU60CYdeMS+B6tgLtevDALBJwkQhwB8t4zbZOibbcZuOHEIR86VNesjsPR2GSFAygwQ6u60Puvg+CebCiPSvQ0dARornJVCcbwk7-HyABqPAQCYfGEIBUAAJLmXywGngEIkOHyiLGSaGgWdZtl8uwgIKYQAAsfJKGCVgMOgiLuEsAQBVZNl2aFkjKcSWaUl01LALSo50BASC6LRHruaBWEmbFJypUFJhIc8tnOa5VWeaRiG+WhfYDjwQ4jt6zITlOEBAA)

# Addressed from issue

### Typed return value
Types return value now instead of any. (Taken from test)
```typescript
function assertType<T>(_x: T) {}

const toBeCloned = {
  name: "abc",
  address: "test",
  age: 30,
  info: {
    url: "https://example.com",
  },
} as const;

const nonMatchingType = { foo: "bar" } as const;
const clone = structuredClone(toBeCloned);

assertType<typeof toBeCloned>(clone);
// @ts-expect-error non matching type
assertType<typeof nonMatchingType>(clone);
```

### Make readonly types mutable
```typescript
type Mutable<T> = {
  -readonly [P in keyof T]: T[P];
};

type readonlyExample = Readonly<{
  name: string;
  age: number;
}>

const example: readonlyExample = {
  name: 'test',
  age: 30
};

const clone = window.structuredClone(example);
// @ts-expect-error non mutable
clone.age = 29;

const mutableClone = window.structuredClone<Mutable<readonlyExample>>(clone);
mutableClone.age = 29;
```

instead of being within the generic function can pass through mutable helper.

# Not addressed from issue

### Error on non structured-cloneable types
From [mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#exceptions) will throw exception if not [structuredCloneable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#javascript_types) type.  Could add a generic constraint to a `StructuredCloneable` type to prevent these errors in this pr?

Example
```typescript
const nonSeriazable = {
  func: () => {}
}

// @ts-expect-error Not serializable
const errorClone = window.structuredClone(nonSeriazable); // Will error on runtime
```

### Errors on transferring an object.
From [docs](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#transferring_an_object)

When an object is transferred using it in the original throws an exception, which isn't captured on typing.

Example
```
// Create an ArrayBuffer with a size in bytes
const buffer1 = new ArrayBuffer(16);

const object1 = {
  buffer: buffer1,
};

// Clone the object containing the buffer, and transfer it
const object2 = structuredClone(object1, { transfer: [buffer1] });

// Create an array from the cloned buffer
const int32View2 = new Int32Array(object2.buffer);
int32View2[0] = 42;
console.log(int32View2[0]);

// @ts-expect-error detached ArrayBuffer
const int32View1 = new Int32Array(object1.buffer); // Will error on runtime
```

Not sure on handling this one `buffer1` is still an `ArrayBuffer` after being cloned, it'll just throw an error if used. Maybe doesn't need to be handled here.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -17776,7 +17776,7 @@ interface WindowOrWorkerGlobalScope {
     reportError(e: any): void;
     setInterval(handler: TimerHandler, timeout?: number, ...arguments: any[]): number;
     setTimeout(handler: TimerHandler, timeout?: number, ...arguments: any[]): number;
-    structuredClone(value: any, options?: StructuredSerializeOptions): any;
+    structuredClone<T = any>(value: T, options?: StructuredSerializeOptions): T;
 }
 
 interface WindowSessionStorage {
@@ -19193,7 +19193,7 @@ declare function queueMicrotask(callback: VoidFunction): void;
 declare function reportError(e: any): void;
 declare function setInterval(handler: TimerHandler, timeout?: number, ...arguments: any[]): number;
 declare function setTimeout(handler: TimerHandler, timeout?: number, ...arguments: any[]): number;
-declare function structuredClone(value: any, options?: StructuredSerializeOptions): any;
+declare function structuredClone<T = any>(value: T, options?: StructuredSerializeOptions): T;
 declare var sessionStorage: Storage;
 declare function addEventListener<K extends keyof WindowEventMap>(type: K, listener: (this: Window, ev: WindowEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
 declare function addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -5751,7 +5751,7 @@ interface WindowOrWorkerGlobalScope {
     reportError(e: any): void;
     setInterval(handler: TimerHandler, timeout?: number, ...arguments: any[]): number;
     setTimeout(handler: TimerHandler, timeout?: number, ...arguments: any[]): number;
-    structuredClone(value: any, options?: StructuredSerializeOptions): any;
+    structuredClone<T = any>(value: T, options?: StructuredSerializeOptions): T;
 }
 
 interface WorkerGlobalScopeEventMap {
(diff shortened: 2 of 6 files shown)


ConorLinehan:

saschanaz:
I guess correctly limiting the type would be a hard work if not impossible. This is already better than before, so I'll merge it. Thanks!

LGTM

ConorLinehan:
@saschanaz I'll have a look to see can it work in a followup pr, think an `extends` on a mapped type with Primitives and Transfers should handle it. Think your right and we'll be able to catch the transferred over case though

PR #1525, opened 2023-03-18
Fix CredentialsContainer.store() return type

lukewarlow:
The existing return type is wrong. This function returns an empty promise.

As can be seen in the specification and the MDN documentation.

https://w3c.github.io/webappsec-credential-management/#abstract-opdef-store-a-credential

https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/store

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -4312,7 +4312,7 @@ interface CredentialsContainer {
     create(options?: CredentialCreationOptions): Promise<Credential | null>;
     get(options?: CredentialRequestOptions): Promise<Credential | null>;
     preventSilentAccess(): Promise<void>;
-    store(credential: Credential): Promise<Credential>;
+    store(credential: Credential): Promise<void>;
 }
 
 declare var CredentialsContainer: {


saschanaz:
MDN is wrong:

* Spec: https://w3c.github.io/webappsec-credential-management/#framework-credential-management
* Gecko implementation: https://searchfox.org/mozilla-central/rev/751e4466a7f4d3354727298d20abeecf39b33c40/dom/webidl/CredentialManagement.webidl#27
* WebKit: https://searchfox.org/wubkat/rev/aea708ebe4cafe961fc0f0c192d99eda3fe95c8b/Source/WebCore/Modules/credentialmanagement/CredentialsContainer.idl#34
* Blink: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/credentialmanagement/credentials_container.idl;l=10;drc=047c7dc4ee1ce908d7fea38ca063fa2f80f92c77

lukewarlow:
Having ran it in the browser it also seems to return undefined?

lukewarlow:
I will close this for now and file an issue on the spec itself because I don't think it's types are correct. Chrome is the only browser I know of that actually implements this API in a meaningful way so the browser type declarations are probably just matching the spec IDL which is wrong.

saschanaz:
Sounds like that's a bug in Blink, maybe file a bug in https://bugs.chromium.org/p/chromium/ ?

lukewarlow:
@saschanaz idk if it requires manual work or if it will update automatically at some point but I've fixed this in the spec (not sure if it's published yet). It now correctly shows it returns an empty promise.

saschanaz:
It's not being recognized because of https://github.com/w3c/webappsec-credential-management/issues/213.

PR #1526, opened 2023-03-18
Generate links to MDN docs

### LINKED ISSUE #1464: Link to MDN docs
I was thinking it would be nice to be able to get a link to the MDN docs for a builtin JavaScript function while hovering over it in VS Code, so I started looking into the possibility of modifying this tool to emit docstrings like the following:

```typescript
/** [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clearRect) */
clearRect(x: number, y: number, w: number, h: number): void;
```

Would this be a welcome change? If so then I could work on the actual implementation and open a pull request, but I wanted to check first in case this wouldn't be wanted (since I don't see any links in docstrings currently).

samestep:
Resolves #1464. I only observed `mdnUrl`s being emitted for `Interface`, `Method`, and `Property`, so those were the only ones to which I added an `mdnUrl` property in `src/build/types.d.ts`.


saschanaz:
AMAZING, the script change is so short and easy! Thank you!

LGTM

Issue #1421, opened 2022-10-26
The definition of FormData.append allows errors if a fileName is specified


m-gallesio:
`FormData.append` is currently typed as:
```typescript
append(name: string, value: string | Blob, fileName?: string): void;
```

However, calling it with both a `string` value and a `fileName` causes an error in both Edge (106.0.1370.52) and Firefox (106.0.1) (it does NOT in Internet Explorer).

Unless this causes other problems, I think it would be better to separate it into two overloads:

```typescript
append(name: string, value: string): void;
append(name: string, value: Blob, fileName: string): void;
```

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) `value` in the first case is automatically converted to a string, to it might even by typed as `any` or a generic.

The same considerations should be also valid for `FormData.set`.

PR #1534, opened 2023-03-31
Prevent runtime errors on FormData set and append

ConorLinehan:
Addresses [this issue](https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1421)

Adds two overloads to `append` and `set` to prevent runtime errors.

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -1240,17 +1240,15 @@
                         "append": {
                             "name": "append",
                             "flavor": "Web",
-                            "overrideSignatures": [
-                                "append(name: string, value: string): void",
-                                "append(name: string, value: Blob, fileName?: string): void"
+                            "additionalSignatures": [
+                                "append(name: string, value: string | Blob): void"
                             ]
                         },
                         "set": {
                             "name": "set",
                             "flavor": "Web",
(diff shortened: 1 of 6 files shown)


inputfiles/overridingTypes.jsonc
```diff
@@ -1241,14 +1241,16 @@
                             "name": "append",
                             "flavor": "Web",
                             "overrideSignatures": [
-                                "append(name: string, value: string | Blob, fileName?: string): void"
+                                "append(name: string, value: string): void",
+                                "append(name: string, value: Blob, fileName?: string): void"
```
saschanaz, line 1245: `additionalSignatures` is probably better here. The original signatures were already looked like this and the overriding signature was intended to support passing argument with type `string | Blob`.

So please do this instead:

```json
"additionalSignatures": [
    "append(name: string, value: string | Blob): void;"
]
```

Same for `set()`. It will end up with three signatures and that should be fine.

unittests/files/formData.ts
saschanaz, line 1: Please add a test where the second argument is a union of string and Blob.

saschanaz (changes requested):

saschanaz:
Cool, thanks! LGTM.

PR #1535, opened 2023-04-01
Add CustomEvent to EventListener

mkcode:
Type definition for EventListener and EventListenerObject is overly narrow and does not allow for CustomEvent.

Submitting this PR on behalf of issue: https://github.com/microsoft/TypeScript/issues/28357

MDN and WhatWG seem to specify that this is allowed:

 - https://dom.spec.whatwg.org/#defining-event-interfaces
 - https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -558,11 +558,11 @@ declare var Event: {
 };
 
 interface EventListener {
-    (evt: Event): void;
+    (evt: Event | CustomEvent): void;
 }
 
 interface EventListenerObject {
-    handleEvent(object: Event): void;
+    handleEvent(object: Event | CustomEvent): void;
 }
 
 /**
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -8060,11 +8060,11 @@ declare var EventCounts: {
 };
 
 interface EventListener {
-    (evt: Event): void;
+    (evt: Event | CustomEvent): void;
 }
 
 interface EventListenerObject {
-    handleEvent(object: Event): void;
+    handleEvent(object: Event | CustomEvent): void;
 }
 
 interface EventSourceEventMap {
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -2293,11 +2293,11 @@ declare var Event: {
 };
 
 interface EventListener {
-    (evt: Event): void;
+    (evt: Event | CustomEvent): void;
 }
 
 interface EventListenerObject {
-    handleEvent(object: Event): void;
+    handleEvent(object: Event | CustomEvent): void;
 }
 
 interface EventSourceEventMap {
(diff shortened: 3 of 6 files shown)


mkcode:
@saschanaz - Following up on your comment: https://github.com/microsoft/TypeScript/issues/28357#issuecomment-1396640257

I believe that extending the EventMap interfaces would not solve the issue for generic EventTargets.

This way does, but may break other things.

saschanaz:
`Event | CustomEvent` is equivalent to `CustomEvent`, and thus this makes every unknown event interfaces as CustomEvents. IMO that's too aggressive, as:

1. Any "custom events" also can use the base Event instead of CustomEvent or any custom subclasses of Event
2. Any new event added in new versions of browsers will collide with the CustomEvent definition (`Event as AnySubclassedEvent` works but `CustomEvent as AnySubclassedEvent` does not)

>I believe that extending the EventMap interfaces would not solve the issue for generic EventTargets.

You should still be able to [add your own addEventListener signature in that case](https://www.typescriptlang.org/play?#code/PTAEFEDcFMDsBcCyBDADqVz72gJ1qAJYBmoAngPYCuoAFsjKALbKxmjQwIDOAsAFADCCPMWQBjaKACa1XFDhI0oAN4DQG0ACJKVeV3haAXKADCVbvApMFCAQF8BQkbjGSZcgJIu3U6AA8cWAATbggDABVkXABzaHhVdVBkYODbeAAZQks4PAAeAGkOQLhQ0ABraEpSWT10lFQAPgAKTkUTAoAaUAAbbKC8E2bm+Fpsk1rcbxxXCWhuzgm5erQAbQKAXQBKUABeRuS2HYAfUFgqHp7uilR4QgpYbgB+EwBBVPSsnNg8AHlb+6PUCnABGFAoPWgrC2JkgFEIwQA3EkUmkDF8BrgRmRUNATJZcMIYt1xMhLiCJOUTJ9+rlcL99IoMXTfiCAFbQcQJU7nS7XAEPZ5vD7o2k-ekCoGg8GQ6Gw+FIhxOfjBTk9aJSSDRIhLPTTURzZH8QgAOlRNO+eGaOmWBi03VakB2+0S-E0HEgJtV8GQhB6RvsWyNAhAoAZoG4hCYqB6ZCMyuEM18Hj1PjmACZikEyukorF4q6NObRZasW0ECYbXU7d0+qWhiMxtxdVM05IFpATOZLNZ0s6DqwyCczhcrqAbndBS9QO80UyxX9JWFpRCobAYaA4QijUWRfPS9jcfj4ITYMTQKTyZTqSXMQyLXf2ZzuSO+eOl9PZw+WUvgaAwaucqbgqAZAA) and follow the pattern of the existing types.

mkcode:
Yep. This is much more complicated than the DX that I want of just specifying the parameter type of the callback function:

```ts
object.addEventListener('my-event', (event: CustomEvent<CustomType>) => { event.detail }))
```

But, I think you are right that this what I wrote is actually just a bad TypeScript pattern. The type definition belongs in the object definition using the methods you described, not in callback handlers.

Thank you. I appreciate your time and detailed response!

PR #1536, opened 2023-04-03
Allows passing iterable to `Blob` and `File`

demensky:
- `Blob`: [mdn](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob#array), [w3c](https://w3c.github.io/FileAPI/#dom-blob-blob)
- `File`: [mdn](https://developer.mozilla.org/en-US/docs/Web/API/File/File#bits), [w3c](https://w3c.github.io/FileAPI/#dom-file-file)

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3073,7 +3073,7 @@ interface Blob {
 
 declare var Blob: {
     prototype: Blob;
-    new(blobParts?: BlobPart[], options?: BlobPropertyBag): Blob;
+    new(blobParts?: Iterable<BlobPart>, options?: BlobPropertyBag): Blob;
 };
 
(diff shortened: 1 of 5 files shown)


demensky:

baselines/dom.generated.d.ts
```diff
@@ -3073,7 +3073,7 @@ interface Blob {
 
 declare var Blob: {
     prototype: Blob;
-    new(blobParts?: BlobPart[], options?: BlobPropertyBag): Blob;
+    new(blobParts?: Iterable<BlobPart>, options?: BlobPropertyBag): Blob;
```
saschanaz, line 3076: You cannot do this as the base lib needs to be ES5 compatible. This kind of thing is done in dom.iterable but the constructor can't be extended in the current design. We are blocked here.

saschanaz (changes requested):

PR #1544, opened 2023-04-24
make location.reload() return never

k-yle:
writing code after `location.reload()` is very likely a mistake, since the code will probably never run. 

So this PR changes `location.reload()` to return `never`, akin to `process.exit()` in node

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -14475,7 +14475,7 @@ interface Location {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Location/reload)
      */
-    reload(): void;
+    reload(): never;
     /**
      * Removes the current page from the session history and navigates to the given URL.
      *
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -584,6 +584,17 @@
                     }
                 }
             },
+            "Location": {
+                "methods": {
+                    "method": {
+                        "reload": {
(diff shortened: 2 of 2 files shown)


saschanaz:
Thanks! LGTM.

HolgerJeromin:
In my test this does not change compile behavior with code after that. 
But no problem with it either.

HolgerJeromin:
Interesting. My code uses the `location` object on `window` which breaks this error also in playground:
```ts
declare interface Location {
  reload(): never;
}
window.location.reload()
console.log("...")

**Comment by turansky** [comment]
What about iframes?

```js
const frame = window[0]
frame.location.reload() // never

console.log("Is it really unreachable?")
```

aerialist7:
And `self.location.reload()`?

HolgerJeromin:
You can use the playground link above to test yourself...
iframe does not work either

aerialist7:
And yes, seems that we can have executable code after `iframe`s location reload

HolgerJeromin:
> And yes, seems that we can have executable code after iframes location reload.

Which is a good thing BTW, because an iframe reload does not break code flow in runtime of the top frame.
But this could result to false negatives which are not solvable by the developer (without many `@ts-ignoreerror`).

saschanaz:
Oh yes, I did not think about the iframe case, I guess I'll have to revert this.

saschanaz:
Reverting this in #1545 to prevent unexpected breakage. Feel free to file an issue to continue the discussion.

PR #1545, opened 2023-04-25
Revert "make location.reload() return never (#1544)"

saschanaz:
This reverts commit b24e2d7a9d73ce657d9eddf1b1c49597812b61ee.


saschanaz:
LGTM

PR #1547, opened 2023-04-29
Make TextEncoderEncodeIntoResult members required

castholm:
`TextEncoderEncodeIntoResult` is defined as a dictionary without `required` members in https://encoding.spec.whatwg.org/#interface-textencoder
```
dictionary TextEncoderEncodeIntoResult {
  unsigned long long read;
  unsigned long long written;
};
```

However, algorithm for `encodeInto()` (which is the only place the type is referenced) effectively ensures that the properties are always set, described in https://encoding.spec.whatwg.org/#dom-textencoder-encodeinto

>1\. Let read be 0.
>2\. Let written be 0.
>...
>7\. Return «[ "read" → read, "written" → written ]».

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -3367,6 +3367,18 @@
                     }
                 }
             },
+            "TextEncoderEncodeIntoResult": {
+                "members": {
+                    "member": {
+                        "read": {
+                            "required": true
+                        },
+                        "written": {
+                            "required": true
+                        }
+                    }
+                }
+            },
             "UnderlyingSource": {
                 "typeParameters": [
                     {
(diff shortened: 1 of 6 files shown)


castholm:

inputfiles/overridingTypes.jsonc
```diff
@@ -3367,6 +3367,18 @@
                     }
                 }
             },
+            "TextEncoderEncodeIntoResult": {
```
saschanaz, line 3370:
```suggestion
            // TextEncoderEncodeIntoResult is used only for return type, always with all fields defined
            "TextEncoderEncodeIntoResult": {
```

saschanaz (commented):
Looks good, could you just add one line of comment?

saschanaz:
Thanks! LGTM.

PR #1569, opened 2023-06-13
Fix `Response.json` mdn link

hax:


HolgerJeromin:
> 27,983 additions, 27,983 deletions

Seems a bit fishy to me :)

PR #1573, opened 2023-06-15
Add augmentable `PermissionNameMap`

Andarist:


baselines/audioworklet.generated.d.ts
```diff
@@ -38,6 +38,15 @@ interface MessageEventInit<T = any> extends EventInit {
     source?: MessageEventSource | null;
 }
 
+interface PermissionNameMap {
```
Andarist, line 41: I'm not sure what's special about this file that this type is being generated here but `PermissionName` is not and i'm not sure how to control this.

saschanaz:
Interesting approach. Do you have any specific permission in mind you want to enable through this, BTW?

Andarist:
> Interesting approach. 

I've seen other `*Map` types here and I thought that this was a somewhat established pattern here. But perhaps those other types had a different purpose after all 😅 

> Do you have any specific permission in mind you want to enable through this, BTW?

A friend of mine was trying to augment the related functions with `'clipboard-write' | 'clipboard-read'`.

saschanaz:
> I've seen other `*Map` types here and I thought that this was a somewhat established pattern here. But perhaps those other types had a different purpose after all 😅

The key difference is that this one is not really a map but a list, while others are actual maps. Maybe rename?

> A friend of mine was trying to augment the related functions with 'clipboard-write' | 'clipboard-read'.

In most cases you don't need those permissions, since user interaction e.g. mouse click and key push will automatically grant those permission for a short time (which is called transient user activation).

PR #1577, opened 2023-06-19
Manually update to idl@3.35.0

saschanaz:


saschanaz:
Oh no, SharedArrayBuffer is ES2017 thing.

inputfiles/overridingTypes.jsonc
```diff
@@ -3553,6 +3553,11 @@
                     }
                 ],
                 "overrideType": "ReadableStreamDefaultReader<T> | ReadableStreamBYOBReader"
+            },
+            {
+                "name": "AllowSharedBufferSource",
+                // No SharedArrayBuffer in ES5, oh no.
```
saschanaz, line 3559: For now I'm just removing SharedArrayBuffer, as I don't have any better solution. cc @sandersn

saschanaz:
lgtm

inputfiles/overridingTypes.jsonc
sandersn: Uhhh, is the problem here that DOM types have to work with any ES target, even ones that are incredibly old, like ES5/ES2015?

If so, this is going to become more of a problem over time. Maybe we could split the DOM types into parts which require newer ES types?

saschanaz: Theoretically we can have `dom.sharedbuffer` like we have `dom.iterable`, yes. That should work for now, until we get any case that function overloads don't work (which is already a problem for `dom.iterable`). Let me file an issue for that...

saschanaz: Filed #1579.

PR #1586, opened 2023-06-29
fix: updated CustomEvent.detail to be null when not defined

### LINKED ISSUE #1585: CustomEvent.detail is null when not defined
CustomEvent.detail is defined as follows:

```ts
interface CustomEvent<T = any> extends Event {
    readonly detail: T;

	// ...
}
```

Source: https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/957eced9e98fc61bf86ab9524deb5eaa1f4d3c24/inputfiles/overridingTypes.jsonc#L538-L540

This is slightly wrong as CustomEvent.detail defaults to `null` when not defined: [Playground](https://www.typescriptlang.org/play?#code/DYUwLgBCELwQdiA7hAwgVwM5gPYFsBRANxHjAB514ATEAMwEtFqA+ACgHI8RNMBDAOYgOASgBQoSHVgJkaLLkIky5IjgatO3XoOHiQAOlpg+DYBH5gGmRjwhVajZgG4xdI+FPnL125ghqGq4AxjjwmDigBsA4AmyGxl6wMHDw6MDAIiFhEVExce6JZsmp6ZmuYmAAngAO0ABKIHzAACKeZgAqtSDkHTJ88FUsMn0gAB5gpNT+AN4AvhAA-BB9AFwIZa7VdRAAgjKNzW0mnd2UNPRMIKyV3RAAQgdNre3AXXWq6jfb0KhPR693j00hkWLcdi1-i8Tm8ztgAE5MAQQAA+9guTmuwzEQA)

The playground contains a fix I suggest: `detail: T extends {} ? T : null`

I'll open a PR referring to this issue

GauBen:
Fixes #1585

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -309,7 +309,7 @@ interface CustomEvent<T = any> extends Event {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/CustomEvent/detail)
      */
-    readonly detail: T;
+    readonly detail: T extends {} ? T : null;
     /**
      * @deprecated
      *
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5892,7 +5892,7 @@ interface CustomEvent<T = any> extends Event {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/CustomEvent/detail)
      */
-    readonly detail: T;
+    readonly detail: T extends {} ? T : null;
     /**
      * @deprecated
      *
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -1710,7 +1710,7 @@ interface CustomEvent<T = any> extends Event {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/CustomEvent/detail)
      */
-    readonly detail: T;
+    readonly detail: T extends {} ? T : null;
     /**
      * @deprecated
      *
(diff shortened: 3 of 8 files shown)


GauBen:
I'll update the tests accordingly if this change is accepted

saschanaz:
Sounds fine to me.

saschanaz:
Can you perhaps also add unit tests for this?

GauBen:
Is `eventlistener.ts` the right place for it?

saschanaz:
I'd prefer a separate file for custom events.

GauBen:
I just noticed that it is possible to create broken CustomEvents:
https://www.typescriptlang.org/play?#code/MYewdgziA2CmB00QHMAUZYHcAEBhArhAC4gC2AogG6xhEA8xATgJZjIB8qA5LNbVwEp4AE1hEAhs2jYAvHOxh80aAIDcAKAD0m7Lr36Dho8aMA9APxA

Do I try to address this too? Would this be a breaking change?

saschanaz:
Looks right, although it starts to be over-complex... I'm okay with that if @sandersn says okay

sandersn:
I don't know CustomEvent, but running the example from the playground link, it looks like `CustomEvent.default` is null if CustomEventInit isn't provided to the constructor. At least, it doesn't appear to be `undefined` the way that `{ detail?: T }` implies. So what about this?

```ts
interface CustomEventInit<T = any> extends EventInit {
    detail: T; // note: you probably need to add the `T extends {} ? T : null` conditional type to map undefined -> null
}

declare var CustomEvent: {
  prototype: CustomEvent;
  new<T>(type: string, eventInitDict: CustomEventInit<T>): CustomEvent<T>
  new(type: string): CustomEvent<null>
};
```

Of course this might break backward compatibility. I don't know how new or widely-used CustomEvent is.

GauBen:
Closing to declutter

PR #1593, opened 2023-07-12
Make the function signature overloads of `SubtleCrypto#exportKey` more flexible

steveluscher:
# Problem

This test case _fails_ with the current state of the `exportKey` overloads ([playground](https://www.typescriptlang.org/play?#code/BQbQ5AVg7g1mA0ACMAnAhlByAOMDGAzgBxZgG4CWYAuomgYgNICmAngGID2KAtmgC4hqASgB0AM24BRNHgAWwSbwGIAvAD5EAbwBQifYjwpW2fp1EEArgCN+AG2ajmAD2zd+LVou59+SLQC+dAwAwsamnJ7CANx6BgD08QbJKQB6APw6ATFAA)).

```ts
(['jwk', 'raw', 'pkcs8', 'spki'] as KeyFormat[]).forEach(format => {
    crypto.subtle.exportKey(
      format, // ERROR
      {} as CryptoKey,
    );
});
```

# Change

This PR makes it so that the `"jwk"` overload works as expected by producing `Promise<JsonWebKey>` as the return type, and the remaining keys in the `KeyFormat` union produce `Promise<ArrayBuffer>` as the return type, _and_ the example code above typechecks correctly ([playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMoFcBGYA2EDCUAngA5gD2yA3gFDL3IQAeJ5UYA0hEQBQxsBbOGABcyAEQArAO4BrcQBpks7mMKkKXIgEoxABSjkBwAM4QAPACkT5EAHUImLQD4A3HQbNW7LX0HCxAFEmBBx0ABMLLQAxfzAlKTlxZyUVIjViMnItXWQDI1MLAEEoKDgiACF0GBhoNw96LzZObj8oIVFkGLjU1WR1LJz9Q2MzcxKyyuraqGdkAB88kcKrG3tHF3cAXxoaHgBtAHIZWUOlQ7LpM+RDklkEEwAOa8OTO+BDgF1kOBNkBFsJjA2gAdPwoIFEAALHi-IggBDIcEdZAAXjmtAYP2kcGAYH+mQoIJMWFwEBBTR8rWRwiUVC2Pz+A003G07ixAHoOViebyGAA9AD8DWQJmkeIQUOQbQ62moIqxCF+KGOckOIgVfLgOLxBI05GJpLwFJYzV8NPi1AZv36hOyrPZfM53KdrqFmt5mCgEDgskdvMi8HQOFEHp52tx+IQdsN2GNlJavAtdOtTLtOX9rq5rrdwqxOy2bKAA)).

### DIFF
--- a/unittests/files/keyusage.ts
+++ b/unittests/files/keyusage.ts
@@ -1,6 +1,7 @@
 function assertType<T>(_x: T) {}
 
 const mockKey = {} as CryptoKey;
+assertType<Promise<ArrayBuffer | JsonWebKey>>(crypto.subtle.exportKey('' as KeyFormat, mockKey));
 assertType<Promise<JsonWebKey>>(crypto.subtle.exportKey('jwk', mockKey));
 assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey('pkcs8', mockKey));
 assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey('raw', mockKey));
--- a/unittests/files/keyusage.ts
+++ b/unittests/files/keyusage.ts
@@ -1,3 +1,5 @@
+function assertType<T>(_x: T) {}
+
 const mockKey = {} as CryptoKey;
 
 assertType<Promise<JsonWebKey>>(crypto.subtle.exportKey("jwk", mockKey));
--- a/unittests/files/keyusage.ts
+++ b/unittests/files/keyusage.ts
@@ -1,24 +1,44 @@
-function assertType<T>(_x: T) {}
-
 const mockKey = {} as CryptoKey;
-assertType<Promise<ArrayBuffer | JsonWebKey>>(crypto.subtle.exportKey('' as KeyFormat, mockKey));
-assertType<Promise<JsonWebKey>>(crypto.subtle.exportKey('jwk', mockKey));
-assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey('pkcs8', mockKey));
-assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey('raw', mockKey));
-assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey('spki', mockKey));
 
-const usageInline = crypto.subtle.generateKey({
-  name: "AES-GCM",
-  length: 256,
-},  true, ['encrypt', 'decrypt'])
+assertType<Promise<JsonWebKey>>(crypto.subtle.exportKey("jwk", mockKey));
+assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey("pkcs8", mockKey));
+assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey("raw", mockKey));
+assertType<Promise<ArrayBuffer>>(crypto.subtle.exportKey("spki", mockKey));
 
-const usageConst = crypto.subtle.generateKey( {
-  name: "AES-GCM",
-  length: 256,
-},  true, ['encrypt', 'decrypt'] as const)
+assertType<Promise<ArrayBuffer | JsonWebKey>>(
+  crypto.subtle
+    .exportKey("" as KeyFormat, mockKey)
+    .then((ambiguousKey) =>
+      ambiguousKey instanceof ArrayBuffer
+        ? (ambiguousKey satisfies ArrayBuffer)
+        : (ambiguousKey satisfies JsonWebKey)
+    )
+);
 
-const keyUsage: ReadonlyArray<KeyUsage> = ['encrypt', 'decrypt']
-const usageAsReadonly = crypto.subtle.generateKey( {
-  name: "AES-GCM",
-  length: 256,
-},  true, keyUsage)
+const usageInline = crypto.subtle.generateKey(
+  {
+    name: "AES-GCM",
+    length: 256,
+  },
+  true,
+  ["encrypt", "decrypt"]
+);
+
+const usageConst = crypto.subtle.generateKey(
+  {
+    name: "AES-GCM",
+    length: 256,
+  },
+  true,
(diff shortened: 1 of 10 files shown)


steveluscher:

steveluscher:
Mmm… hold on. I see why this doesn't work for `importKey` quite the same way. I'll revert that.

saschanaz:
I'm pretty sure this is done to fix the `jwk` incorrectly returning ArrayBuffer. Can you write some unit tests and check the return type?

steveluscher:
Check out the second playground link; `"jwk"` still returns `Promise<JsonWebKey>`!

saschanaz:
I mean: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMoFcBGYA2EDCUAngA5gD2yA3gFDL3IQAeJ5UYA0hEQBQxsBbOGABcyAEQArAO4BrcQBpks7mMKkKXIgEoxABSjkBwAM4QAPACkT5EAHUImLQD4A3HQbNW7LX0HCxLQAxfzAlFSI1YjJyLV1kAyNTCwBBKCg4IgAhdBgYaDcaAF8aGjwwZExc-KgxNIzs6uh3HgBtAHIZWXaldozpHuR2klkEEwAOQfaTEeB2gF1kOBNkBFsTMG0AOn4oAFFEAAseZaIQBGRdoQqAXmdqD3oqvOhkG6XpOGAKhGiKLZMWFwEC2XjYnG4fig1yUVCKSxW6hicXcDGQAHp0cgACqHUzIEyHcjoHAAE0uXxwlXQFW+yFJ5AgKxA5Aq8GAOAAhMVtK4gA

unittests/files/keyusage.ts
```diff
@@ -1,3 +1,11 @@
+function assertType<T>(_x: T) {}
+
+const mockKey = {} as CryptoKey;
+assertType<Promise<JsonWebKey>>(crypto.subtle.exportKey('jwk', mockKey));
```
saschanaz, line 4: Please also check what happens if the input is `'jwk' | 'pkcs8'`.

saschanaz:
I think we are blocked by https://github.com/microsoft/TypeScript/issues/14107 which is still open.

unittests/files/keyusage.ts
steveluscher: 100%. I'm on it.

steveluscher:
There we go! Updated the [playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMoFcBGYA2EDCUAngA5gD2yA3gFDL3IQAeJ5UYA0hEQBQxsBbOGABcyAEQArAO4BrcQBpks7mMKkKXIgEoxABSjkBwAM4QAPACkT5EAHUImLQD4A3HQbNW7LX0HCxAFEmBBx0ABMLLQAxfzAlKTlxZyUVIjViMnItXWQDI1MLAEEoKDgiACF0GBhoNw96LzZObj8oIVFkGLjU1WR1LJz9Q2MzcxKyyuraqGdkAB88kcKrG3tHF3cAXxoaHgBtAHIZWUOlQ7LpM+RDklkEEwAOa8OTO+BDgF1kOBNkBFsJjA2gAdPwoIFEAALHi-IggBDIcEdZAAXjmtAYP2kcGAYH+mQoIJMWFwEBBTR8rWRwiUVC2Pz+A003G07ixAHoOViebyGAA9AD8DWQJmkeIQUOQbQ62moIqxCF+KGOckOIgVfLgOLxBI05GJpLwFJYzV8NPi1AZv36hOyrPZfM53KdrqFmt5mCgEDgskdvMi8HQOFEHp52tx+IQdsN2GNlJavAtdOtTLtOX9rq5rrdwqxOy2bKAA) link in the description.

saschanaz:
Ah, that works better 👍

unittests/files/keyusage.ts
```diff
@@ -1,3 +1,12 @@
+function assertType<T>(_x: T) {}
+
+const mockKey = {} as CryptoKey;
+assertType<Promise<ArrayBuffer | JsonWebKey>>(crypto.subtle.exportKey('' as KeyFormat, mockKey));
```
saschanaz, line 4: IMO this is too loose as either `Promise<ArrayBuffer>` or `Promise<JsonWebKey>` will match the type. But not sure how to make it strict enough 🤔

saschanaz: Maybe we can test whether it works for both `as Promise<ArrayBuffer>` and `as Promise<JsonWebKey>`, like:

```ts
crypto.subtle.exportKey('' as KeyFormat, mockKey) as Promise<ArrayBuffer>; // type includes ArrayBuffer
crypto.subtle.exportKey('' as KeyFormat, mockKey) as Promise<JsonWebKey>; // type also includes JsonWebKey
```

steveluscher: > This is too loose as either Promise<ArrayBuffer> or Promise<JsonWebKey> will match the type

Isn't that exactly the point when `format` is not known to yield one or the other?

How about this?

```ts
assertType<Promise<ArrayBuffer | JsonWebKey>>(
  crypto.subtle
    .exportKey("" as KeyFormat, mockKey)
    .then((ambiguousKey) =>
      ambiguousKey instanceof ArrayBuffer
        ? (ambiguousKey satisfies ArrayBuffer)
        : (ambiguousKey satisfies JsonWebKey)
    )
);
```

saschanaz: > Isn't that exactly the point when `format` is not known to yield one or the other?

Yes but that'll never fail when the return type regresses to either ArrayBuffer or JsonWebKey.

>How about this?

Sounds good!

saschanaz:
This is technically a breaking change but I think it's reasonable.

LGTM

PR #1594, opened 2023-07-12
Add `Ed25519` overload for `SubtleCrypto#generateKey`

steveluscher:
I find myself constantly having to do this:

```ts
const key = await crypto.subtle.generateKey('Ed25519', false, ['sign', 'verify']) as CryptoKeyPair;
```

This PR updates the types so as to indicate that this method always returns a keypair for the algorithm `Ed25519`. It further constrains the type of the `KeyUsages` to just `sign` and `verify` – the only ones allowed by [the spec](https://wicg.github.io/webcrypto-secure-curves/#ed25519).

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -1589,7 +1589,7 @@
                         "generateKey": {
                             "name": "generateKey",
                             "additionalSignatures": [
-                                "generateKey(algorithm: \"Ed25519\", extractable: boolean, keyUsages: ReadonlyArray<Extract<KeyUsage, \"sign\" | \"verify\">>): Promise<CryptoKeyPair>",
+                                "generateKey(algorithm: \"Ed25519\", extractable: boolean, keyUsages: ReadonlyArray<\"sign\" | \"verify\">): Promise<CryptoKeyPair>",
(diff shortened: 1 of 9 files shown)


saschanaz:
Maybe we should have KeyUsagesMap for other algorithms too?

inputfiles/overridingTypes.jsonc
```diff
@@ -1589,6 +1589,7 @@
                         "generateKey": {
                             "name": "generateKey",
                             "additionalSignatures": [
+                                "generateKey(algorithm: \"Ed25519\", extractable: boolean, keyUsages: ReadonlyArray<Extract<KeyUsage, \"sign\" | \"verify\">>): Promise<CryptoKeyPair>",
```
saschanaz, line 1592: Isn't `Extract<KeyUsage, "sign" | "verify">` effectively just `"sign" | "verify"`? Maybe the latter can be concise.

saschanaz:
LGTM per:

* The spec: https://wicg.github.io/webcrypto-secure-curves/#ed25519-operations
* WPT test: https://github.com/web-platform-tests/wpt/blob/0a2fec9de6371ed856208657ee313e4bc7818a41/WebCryptoAPI/generateKey/successes.js#L30
* WPT results: https://wpt.fyi/results/WebCryptoAPI/generateKey/successes_Ed25519.https.any.html?label=experimental&label=master&aligned

But this is Ed25519 specific, we should generalize this for other algorithms.

PR #1609, opened 2023-08-09
Fix argument name in signatures of {add,remove}EventListener

NilsIrl:
It now matches the definition in
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#syntax

This came up following the use of a linter rule that checks for parameter name comments like https://github.com/microsoft/TypeScript/blob/main/scripts/eslint/rules/argument-trivia.cjs does

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -214,9 +214,13 @@ interface AbortSignal extends EventTarget {
     readonly reason: any;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/AbortSignal/throwIfAborted) */
     throwIfAborted(): void;
+    addEventListener<K extends keyof AbortSignalEventMap>(type: K, listener: (this: AbortSignal, ev: AbortSignalEventMap[K]) => any, useCapture: boolean): void;
     addEventListener<K extends keyof AbortSignalEventMap>(type: K, listener: (this: AbortSignal, ev: AbortSignalEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
+    addEventListener(type: string, listener: EventListenerOrEventListenerObject, useCapture: boolean): void;
     addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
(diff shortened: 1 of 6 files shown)


saschanaz:
This will break if the third argument is passed as union type, right?

NilsIrl:
> This will break if the third argument is passed as union type, right?

That sounds correct. What would be the best solution?

```ts
addEventListener(/* ... */ options: AddEventListenerOptions): void;
addEventListener(/* ... */ useCapture: boolean): void;
addEventListener(/* ... */ options?: AddEventListenerOptions | boolean): void;
```

or

```ts
addEventListener(/* ... */ useCapture: boolean): void;
addEventListener(/* ... */ options?: AddEventListenerOptions | boolean): void;
```

or something else?

NilsIrl:
@microsoft-github-policy-service agree company=Google

saschanaz:
The latter sounds fine, except I'm not sure how argument-trivia reports a problem here as I'm not familiar with that rule. What does it say?

NilsIrl:
> The latter sounds fine

Updated the PR.

> Is it something like forcing `/* useCapture */ true`?

Yep. It enforces that certain arguments (`true`, `false`, `null`, `undefined`) are tagged. When there is a tag, makes sure that it matches the name of the argument.

> (Me thinking maybe `{ useCapture: true }` is easier than having a comment)

Probably, but it's easier to change it here to match the MDN documentation than every usage of it.

saschanaz:
Thanks. This looks good, but could you add some comment why the new overload is needed?

saschanaz:
Actually, can we use `capture` instead? Whatever MDN is saying is kinda arbitrary, the actual dictionary member name is `{ capture: true }`, so maybe better with some consistency.

NilsIrl:
`useCapture` comes from previous versions of the DOM Events spec: https://www.w3.org/TR/2009/WD-DOM-Level-3-Events-20090908/#interface-EventTarget

saschanaz:
And that's not used now. I think it would be easier to recognize if we could make the equivalent calls look equivalent:

```ts
foo.addEventListener(name, listener, /* capture */ true);
foo.addEventListener(name, listener, { capture: true });
```

NilsIrl:
Considering the point of this is to not "break" backward compatibility. If this still requires people to change their uses of `useCapture`, I think it would be preferable if `{ capture: true }` was used for new code instead of `/* capture */`.

PR #1628, opened 2023-10-08
Manually update to idl@3.39.0

saschanaz:
Closes #1625 

Mostly about replacing some GLuint uses to number, and adding MouseEvent.layerX/layerY.


saschanaz:
lgtm

PR #1634, opened 2023-10-09
Override SVGSVGElement method getElementById

remcohaszing:
This method may return null, but not according to the current type definitions.


HolgerJeromin:
Perhaps on this svg level this could/should return `svgelement|null` because `<foreignObject>` should be rare enough.

saschanaz:
This is a spec issue, I just filed one. Thanks! https://github.com/w3c/svgwg/issues/923

Bashamega:
Hello @saschanaz 
I think since your PR was merged, we can close this pr

PR #1652, opened 2023-11-27
Narrow types of 'midimessage' and  'statechange' events

fa-sharp:
This correctly narrows the types of the `midimessage` and `statechange` events from the Web MIDI API. The interfaces are correctly laid out in MDN, but currently they show up as a generic Event type in TypeScript. I've listed the relevant docs below.

MIDIInput: [midimessage event](https://developer.mozilla.org/en-US/docs/Web/API/MIDIInput/midimessage_event)
MIDIPort: [statechange event](https://developer.mozilla.org/en-US/docs/Web/API/MIDIPort/statechange_event)


fa-sharp:

PR #1660, opened 2023-12-16
Use `@webref/events` to detect event types

saschanaz:


src/build/utils/map.ts
```diff
@@ -16,3 +16,14 @@ export function addToStringMap(
...
   map.set(name, `${old}\n${value}\n`);
 }
+
+export function addToNestedMap(
+  map: Map<string, Map<string, string>>,
+  name: string,
+  key: string,
+  value: string,
+): void {
+  const nested = map.get(name) || new Map();
```
HolgerJeromin, line 26:
```suggestion
  const nested = map.get(name) ?? new Map();
```
But should not be a big difference.

sandersn (approved):

sandersn:
This has been around a while but I just now looked at it. @saschanaz is it ready to merge?

saschanaz:
Yes 👍

PR #1666, opened 2023-12-29
Remove invalid documentation for pause method of HTMLMediaElement

thkruz:
Fairly simple one. Since there is no return value, the paused property should be used for testing if a media element is paused instead of the pause method.

[Looking at the spec](https://html.spec.whatwg.org/multipage/media.html#dom-media-pause-dev)
> The paused attribute represents whether the [media element](https://html.spec.whatwg.org/multipage/media.html#media-element) is paused or not.

That comment has been there 6 years, so I am not sure of the origin, but it confused me so I made the PR.

References:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/pause
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/paused

### DIFF
--- a/inputfiles/comments.json
+++ b/inputfiles/comments.json
@@ -1574,7 +1574,7 @@
                 "methods": {
                     "method": {
                         "pause": {
-                            "comment": "Pauses the current playback and sets paused to TRUE. This can be used to test whether the media is playing or paused. You can also use the pause or play events to tell whether the media is playing or not."
+                            "comment": "Pauses the current playback and sets paused to TRUE."
                         },
                         "play": {
                             "comment": "Loads and starts playback of a media resource."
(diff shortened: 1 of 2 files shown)


inputfiles/comments.json
```diff
@@ -1811,11 +1811,11 @@
                 "properties": {
                     "property": {
                         "parent": {
-                            "comment" : "Refers to either the parent WindowProxy, or itself.\n\nIt can rarely be null e.g. for contentWindow of an iframe that is already removed from the parent."
```
thkruz, line 1814: This was done automatically by prettier, but I can remove it if it is an issue. It does appear to be a valid correction as well.

thkruz:

saschanaz:
Yeah, it has been there forever since 2015:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/34bf3527d372ec24eac132a596096e61575fc7e8/inputfiles/comments.json#L2650-L2653

Perhaps it was somewhere from MSDN?

saschanaz:
LGTM

thkruz:
> Yeah, it has been there forever since 2015:

Haha yea when I saw that I assumed I was wrong since no one else has fixed it.

> Perhaps it was somewhere from MSDN?

I couldn't find it on google, but who knows what i might have found back then.

Regardless, glad to help. Thanks.

PR #1670, opened 2024-01-04
Support `HTMLElement.hidden` value `"until-found"`

pje:
`HTMLElement.hidden` can now be any of `true | false | "until-found"`. ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden#until-found))

This has support from two major browsers (Chrome and Edge). ([caniuse](https://caniuse.com/mdn-html_global_attributes_hidden_until-found_value))

Screen Shot 2024-01-03 at 19 08 05
Screen Shot 2024-01-03 at 19 07 00

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -484,9 +484,7 @@
                 "properties": {
                     "property": {
                         "hidden": {
-                            // https://github.com/whatwg/html/pull/7475
-                            // Revisit when this get browser supports.
-                            "overrideType": "boolean",
+                            "overrideType": "boolean | \"until-found\"",
                             "nullable": false
                         }
                     }


pje:
> ```
> @microsoft-github-policy-service agree
> ```

HolgerJeromin:
> This has support from two major browsers (Chrome and Edge). (caniuse)

We need support from two major engines. Chrome and Edge are using both the blink engine.

pje:
> We need support from two major engines.

Ah yes, I misread that clause in the readme!

In that case this isn't mergeable yet.

saschanaz:
This is good to go as of 2025:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1761043
* https://github.com/WebKit/WebKit/pull/43043

Care to rebase it?

laymonage:
I have rebased this in #2254.

PR #1676, opened 2024-01-21
Updating added types for `ReadableStreamReadDoneResult`

### LINKED ISSUE #1675: `ReadableStreamReadDoneResult` shouldn't be generic and should have `value: undefined`
As per https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/read#return_value and https://streams.spec.whatwg.org/#ref-for-default-reader-read%E2%91%A0

> If the stream becomes closed, the promise will be fulfilled with an object of the form `{ value: undefined, done: true }`.

The current types are:

https://github.com/microsoft/TypeScript/blob/cea0b61320ee69ae863cc9d7bf873113eb420e3b/src/lib/dom.generated.d.ts#L1622-L1625

```typescript
interface ReadableStreamReadDoneResult<T> {
    done: true;
    value?: T;
}
```

While I'd expect them to be

```typescript
interface ReadableStreamReadDoneResult {
    done: true;
    value: undefined;
}
```

I believe this is a result of:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/f4e507c3d0ba16fde16026d9265c201fecdb6538/inputfiles/addedTypes.jsonc#L1290-L1310

kraenhansen:
This fixes #1675 by generating:

```typescript
type ReadableStreamReadResult<T> = ReadableStreamReadValueResult<T> | ReadableStreamReadDoneResult;

interface ReadableStreamReadDoneResult {
    done: true;
    value: undefined;
}

interface ReadableStreamReadValueResult<T> {
    done: false;
    value: T;
}
```

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -1289,11 +1289,6 @@
             },
             "ReadableStreamReadDoneResult": {
                 "name": "ReadableStreamReadDoneResult",
-                "typeParameters": [
-                    {
-                        "name": "T"
-                    }
-                ],
                 "members": {
                     "member": {
                         "done": {
@@ -1303,7 +1298,8 @@
                         },
                         "value": {
                             "name": "value",
-                            "overrideType": "T"
+                            "overrideType": "undefined",
+                            "required": true
                         }
                     }
                 }
@@ -1535,7 +1531,7 @@
                         "type": "ReadableStreamReadDoneResult"
                     }
                 ],
-                "overrideType": "ReadableStreamReadValueResult<T> | ReadableStreamReadDoneResult<T>"
+                "overrideType": "ReadableStreamReadValueResult<T> | ReadableStreamReadDoneResult"
             },
             {
                 "name": "EventListenerOrEventListenerObject",


saschanaz:
Makes sense, bu you need to build and commit the result.

saschanaz:
Actually byte streams indeed return the value, btw. Perhaps split the type? https://streams.spec.whatwg.org/#byob-reader-internal-slots

saschanaz:
Hi @MattiasBuelens, would you be satisfied by this?

Also it would be good to have some unit tests here.

kraenhansen:
I've updated the code to separate the `ReadableStreamReadResult` into two:
- `ReadableStreamDefaultReadResult`
- `ReadableStreamBYOBReadResult`

I've also added `ReadableStreamReadResult` which union both, to keep breaking changes to a minimum.

Which re-use the `ReadableStreamReadValueResult` but have their separate "DoneResult" types:
- `ReadableStreamDefaultReadDoneResult`
- `ReadableStreamBYOBReadDoneResult`

The PR is currently failing tests:

```
Test failed: could not compile 'dom.generated.d.ts':
generated/dom.generated.d.ts(28201,75): error TS2552: Cannot find name 'ReadableStreamBYOBReadDoneResult'. Did you mean 'ReadableStreamBYOBReadResult'?
generated/dom.generated.d.ts(28203,78): error TS2552: Cannot find name 'ReadableStreamDefaultReadDoneResult'. Did you mean 'ReadableStreamDefaultReadResult'?
```

I can't seem to get the specialised `*DoneResult` types into the `dom.generated.d.ts` 🤔

inputfiles/addedTypes.jsonc
```diff
@@ -1303,7 +1320,8 @@
                         },
                         "value": {
                             "name": "value",
-                            "overrideType": "T"
+                            "overrideType": "T",
```
MattiasBuelens, line 1323: `ReadableStreamBYOBReadDoneResult.value` should be `T | undefined`, as [per spec](https://streams.spec.whatwg.org/#ref-for-byob-reader-read%E2%91%A4):
> If the reader is canceled, the promise will be fulfilled with an object of the form `{ value: undefined, done: true }`. In this case, the backing memory region of view is discarded and not returned to the caller.

See also: [Deno](https://github.com/denoland/deno/blob/v1.39.4/ext/web/lib.deno_web.d.ts#L609-L624) and [web-streams-polyfill](https://github.com/MattiasBuelens/web-streams-polyfill/blob/v3.3.2/src/lib/readable-stream/byob-reader.ts#L27-L38).

inputfiles/addedTypes.jsonc
```diff
@@ -1303,7 +1320,8 @@
...
+                            "required": true
```
MattiasBuelens, line 1324: If we make `value: T | undefined`, should we also make it optional? Or does that not work well together with [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes)?

MattiasBuelens (changes requested):

MattiasBuelens:
> The PR is currently failing tests:
> 
> ```
> Test failed: could not compile 'dom.generated.d.ts':
> generated/dom.generated.d.ts(28201,75): error TS2552: Cannot find name 'ReadableStreamBYOBReadDoneResult'. Did you mean 'ReadableStreamBYOBReadResult'?
> generated/dom.generated.d.ts(28203,78): error TS2552: Cannot find name 'ReadableStreamDefaultReadDoneResult'. Did you mean 'ReadableStreamDefaultReadResult'?
> ```
> 
> I can't seem to get the specialised `*DoneResult` types into the `dom.generated.d.ts` 🤔

The Streams types have `[Exposed=*]` in their IDL, so I think you need to add `"exposed": "*"` (or something equivalent) to your `addedTypes.jsonc` changes?

inputfiles/addedTypes.jsonc
kraenhansen: I don't think so, the spec is pretty clear that the `value` property would exist: https://streams.spec.whatwg.org/#default-reader-read

kraenhansen:
> I can't seem to get the specialised *DoneResult types into the dom.generated.d.ts 🤔

I found the issue 👍

inputfiles/addedTypes.jsonc
kraenhansen: You're right. Reading the spec you've linked to, it seems pretty explicit that the `value` property will be in the object returned from `read`, even from a `ReadableStreamBYOBReader`.

PR #1686, opened 2024-02-08
Fix codeql

jakebailey:
`GITHUB_TOKEN`'s default perms were changed, and this older CodeQL could no longer push security events. Update it to a version based on the one from the TypeScript repo.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,34 +1,71 @@
-# copied from microsoft/TypeScript; see that file for boilerplate commentary
-# (which was certainly copied from somewhere else)
-name: "Code scanning - action"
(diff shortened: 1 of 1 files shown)


saschanaz:
Hi Jake, is this related to the bot failure too? https://github.com/microsoft/TypeScript-DOM-lib-generator/actions/runs/7864326517/job/21455941675

jakebailey:
Yes, definitely.

saschanaz:
Any idea how to fix that too? Maybe the bot should get a new token or?

jakebailey:
I think it's just adding `pull-requests: write` or similar to a permissions block.

I'll try and fix it, just need to read over the code to see what perms it needs (meant to say that originally but forgot)

PR #1688, opened 2024-02-11
Add permissions to codeowners merge workflow

jakebailey:
This should fix auto merging. The permissions block might need `issue: write`, too, but I think that all of the PR stuff can be accessed using `pull-requests: write` by my reading of https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests


jakebailey:
Merging, this is a `pull_request_target` so there's no way to test it before merging as those are always pulled from main for security reasons.

PR #1697, opened 2024-03-11
Strengthen type of `HTMLSelectElement.type`

andersk:
According to https://html.spec.whatwg.org/multipage/form-elements.html#dom-select-type-dev: ‘Returns "`select-multiple`" if the element has a `multiple` attribute, and "`select-one`" otherwise.’


saschanaz:
Makes sense, thanks! LGTM.

PR #1699, opened 2024-03-20
Add [Symbol.toStringTag] property to all interfaces

### LINKED ISSUE #1641: `Symbol.toStringTag` is missing on `File`, `Blob` and `FormData`

jdufresne:
Fixes #1641

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -52,13 +52,23 @@ async function emitFlavor(
     result,
   );
 
-  const iterators = emitWebIdl(exposed, options.global[0], "sync");
+  const toStringTag = emitWebIdl(exposed, options.global[0], "toStringTag");
+  await fs.writeFile(
+    new URL(`${options.name}.tostringtag.d.ts`, options.outputFolder),
+    toStringTag,
+  );
+
+  const iterators = emitWebIdl(exposed, options.global[0], "iterator");
   await fs.writeFile(
     new URL(`${options.name}.iterable.generated.d.ts`, options.outputFolder),
     iterators,
   );
 
-  const asyncIterators = emitWebIdl(exposed, options.global[0], "async");
+  const asyncIterators = emitWebIdl(
+    exposed,
+    options.global[0],
+    "asyncIterator",
+  );
   await fs.writeFile(
     new URL(
       `${options.name}.asynciterable.generated.d.ts`,
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -135,7 +135,7 @@ function isEventHandler(p: Browser.Property) {
 export function emitWebIdl(
   webidl: Browser.WebIdl,
   global: string,
-  iterator: "" | "sync" | "async",
+  flavor: "" | "iterator" | "asyncIterator" | "toStringTag",
 ): string {
   // Global print target
   const printer = createTextWriter("\n");
@@ -230,10 +230,12 @@ export function emitWebIdl(
     getParentsWithConstant,
   );
 
-  switch (iterator) {
-    case "sync":
+  switch (flavor) {
+    case "toStringTag":
+      return emitES6ToStringTag();
+    case "iterator":
       return emitES6DomIterators();
-    case "async":
+    case "asyncIterator":
       return emitES2018DomAsyncIterators();
     default:
       return emit();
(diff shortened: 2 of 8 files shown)


jdufresne:
I made an attempt but I'm hoping to receive a bit guidance and early feedback on the approach and how to proceed. Running tests results in errors that look like:

```
generated/dom.generated.d.ts(28925,18): error TS1169: A computed property name in an interface must refer to an expression whose type is a literal type or a 'unique symbol' type.
generated/dom.generated.d.ts(28925,19): error TS2585: 'Symbol' only refers to a type, but is being used as a value here. Do you need to change your target library? Try changing the 'lib' compiler option to es2015 or later.
```

I'm not sure how to reconcile it.

saschanaz (commented):
You are already going in the right direction! One problem is that, as the error message says, Symbols are ES2015+ while dom.generated.d.ts needs to be ES5-compatible as of now. We probably need another set of files e.g. `*.stringtag.d.ts` just like `*.iterable.d.ts`.

jdufresne:

jdufresne:
Thanks for the feedback. I took your advice and tried to follow the existing iterable pattern.

src/build/emitter.ts
```diff
@@ -135,7 +135,7 @@ function isEventHandler(p: Browser.Property) {
 export function emitWebIdl(
   webidl: Browser.WebIdl,
   global: string,
-  iterator: "" | "sync" | "async",
+  flavor: "" | "sync" | "async" | "toStringTag",
```
saschanaz, line 138: sync/async is now confusing in this case, maybe iterator/asyncIterator?

src/build/emitter.ts
```diff
@@ -1786,6 +1788,38 @@ export function emitWebIdl(
...
+      name: "[Symbol.toStringTag]",
+      readonly: true,
+      comment:
+        "The well-known symbol @@toStringTag.\n\n" +
+        "[MDN Reference](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag)",
+    };
+
+    const interfaces = getElements(webidl.interfaces, "interface");
+    interfaces.sort(compareName).forEach((i) => {
+      if (!i.noInterfaceObject && !i.namespace && !i.extends && !i.implements) {
```
saschanaz, line 1808: I get `i.extends` as it'll be inherited from parent interface, but why others checks? We do need to filter out i.legacyNamespace and emit it inside a namespace, see also:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/6621974de7d3d03ade59760ea4b1cdb12698f492/src/build/emitter.ts#L1491

saschanaz (commented):
Generally nice!

saschanaz (commented):
Sorry for the long silence, this looks nice. cc @sandersn for further opinion from TS team as this introduces a new set of files.

sandersn:
I don't much like the implementation since it requires so many new files. Also, I don't understand how this is useful to anybody. It's not a property that you need to refer to directly, is it?

Following the links in the original bug and skimming through them, it seems like there's some complex assignability problems without this property. Would it be possible to fix those in a simpler way (or stop using this property entirely -- I don't think it's meant for `instanceof` checks in the first place).

@saschanaz I skimmed the webidl spec discussion but couldn't find a definitive place that says that these are part of the DOM. How important do you think it is for spec compliance to have them here?

saschanaz:
> @saschanaz I skimmed the webidl spec discussion but couldn't find a definitive place that says that these are part of the DOM. How important do you think it is for spec compliance to have them here?

By DOM I guess you meant Web Platform? Anything in Web IDL spec is part of Web Platform in that case, otherwise it won't be there.

About the size, given that the new files are pretty much duplicated with same symbolic member definitions, perhaps there can be an empty `interface WebIDLInterface {}` that extends every interface definition in es5 lib and then es6 lib - well, dom.iterable.d.ts - can add a member on it`. Then we don't have to add new files.

saschanaz:
About how important, I think it's pretty much edge case to use this at all, so maybe not too much worth extra effort 🤔

src/build.ts
```diff
@@ -52,13 +52,23 @@ async function emitFlavor(
     result,
   );
 
-  const iterators = emitWebIdl(exposed, options.global[0], "sync");
+  const toStringTag = emitWebIdl(exposed, options.global[0], "toStringTag");
+  await fs.writeFile(
+    new URL(`${options.name}.tostringtag.d.ts`, options.outputFolder),
```
petamoriken, line 57:
```suggestion
    new URL(`${options.name}.tostringtag.generated.d.ts`, options.outputFolder),
```

jdufresne:
Just a heads up. I'm not actively working on this at the moment. If someone would like to pick up the remaining work to take it over the finish line, that is welcome.

PR #1701, opened 2024-04-14
Update to bcd@5.5.21

### LINKED ISSUE #1677: `ReadableStream` should be async-iterable
As per https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration and https://streams.spec.whatwg.org/#rs-asynciterator

> `ReadableStream` implements the [async iterable protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols). This enables asynchronous iteration over the chunks in a stream using the [for await...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) syntax.

The current types declare no `Symbol.asyncIterator` nor extends `AsyncIterator`:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/f4e507c3d0ba16fde16026d9265c201fecdb6538/baselines/dom.generated.d.ts#L18604-L18631

I'd expect a `Symbol.asyncIterator` member on the `ReadableStream`.

I noticed that it does seem like an `iterator` is added to the input file: https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/f4e507c3d0ba16fde16026d9265c201fecdb6538/inputfiles/overridingTypes.jsonc#L2629-L2635

MattiasBuelens:
ReadableStream is finally async-iterable! 🎉 

Fixes #1677

### DIFF
--- a/baselines/audioworklet.asynciterable.generated.d.ts
+++ b/baselines/audioworklet.asynciterable.generated.d.ts
@@ -1,3 +1,8 @@
 /////////////////////////////
 /// AudioWorklet Async Iterable APIs
 /////////////////////////////
+
+interface ReadableStream<R = any> {
+    [Symbol.asyncIterator](options?: ReadableStreamIteratorOptions): AsyncIterableIterator<R>;
+    values(options?: ReadableStreamIteratorOptions): AsyncIterableIterator<R>;
+}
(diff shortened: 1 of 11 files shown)


saschanaz:
Huh, the auto updater is broken again. Thanks! LGTM

DDDDD12138:
image

It seems to be not working now, TypeScript version `5.7.3`.

PR #1709, opened 2024-04-28
Let dependabot to update in group and touch package.json

saschanaz:


saschanaz:
It seems touching dependabot.yml somehow does not trigger the bot. That means I should manually ping @sandersn.

.github/dependabot.yml
```diff
@@ -10,5 +10,21 @@ updates:
...
-    versioning-strategy: "lockfile-only"
+    # Usually only bump versions in package-lock.json, but update package.json for major version updates
+    versioning-strategy: "increase-if-necessary"
+    # Update all dependencies in a single PR
+    groups:
+      # ESLint usually requires updating together for major updates
+      eslint:
+        patterns:
+          - "eslint*"
+          - "@typescript-eslint/*"
```
MattiasBuelens, line 21: Looking at [what will be needed for TypeScript ESLint in the future](https://typescript-eslint.io/getting-started#quickstart), we may want to already add `@eslint/*` too:
```suggestion
      eslint:
        patterns:
          - "eslint*"
          - "@eslint/*"
          - "@typescript-eslint/*"
```

sandersn (approved):
Looks good according to my reading of the documentation.

1. Maybe add `@eslint/*` in a followup PR?
2. Only slightly related, but I wonder whether exact versions would be safer long-term? But that may not be well-supported by tools like dependabot -- the documentation I read assumes version ranges.

PR #1711, opened 2024-04-29
Update Body json() method to allow generic types.

incogiscool:
This all started when I tried to use the Body > json() API for my server using Typescript.

Unfortunately, the json() function returns a `Promise<any>` type, which is not ideal when trying to build type-safe code.
A way I got around this was by using the `as` operator, which allowed me to cast my types. This isn't a completely ideal method but it worked.

```ts
const data = (await request.json()) as SetCreatorDataRequest;
```
A much better way to use this API with type-safety would be to how axios does it:
```ts
axios.get<MyTypeHere>(...)
```

So I adjusted the Body > json() function so that it allows for a generic type (`T`), and defaults to an `any` type.
```ts
json<T = any>(): Promise<T>
```

Hopefully this pull request gets implemented and becomes a feature of the Typescript DOM API.

Thanks :)

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3165,7 +3165,7 @@ interface Body {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/formData) */
     formData(): Promise<FormData>;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/json) */
-    json(): Promise<any>;
+    json<T = any>(): Promise<T>;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Request/text) */
     text(): Promise<string>;
 }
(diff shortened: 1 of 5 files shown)


saschanaz:
I'm pretty sure there was a similar PR years ago which was rejected because "we only add type parameter when it's actually used in paremeters, for this case you can just cast the return type".

Is that changed, i.e. are there similar declarations as this PR does in existing JS types in TypeScript?

incogiscool:

incogiscool:
Hey, I was just wondering if there is any update on the state of this PR. Is there anything I need to change, or should I close if It won't get implemented?

Thanks.

saschanaz:
This was my previous response, which still seems valid: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1078#issuecomment-889181856

PR #1712, opened 2024-05-03
feat: add type parameter to EventTarget

43081j:
This introduces a type parameter to `EventTarget` such that the following is now possible:

```ts
interface CustomMap {
  'test-event': CustomEvent<{x: number; y: number}>;
}

declare const customTarget: EventTarget<CustomMap>;

customTarget.addEventListener('test-event', (event) => {
  // event is now `CustomEvent<{x: number; y: number}>`
  event.detail.x;
  event.detail.y;
});
```

Defaults to `T = any` to maintain existing behaviour otherwise.

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -583,7 +583,7 @@ interface EventListenerObject {
  *
  * [MDN Reference](https://developer.mozilla.org/docs/Web/API/EventTarget)
  */
-interface EventTarget {
+interface EventTarget<T = any> {
     /**
      * Appends an event listener for events whose type attribute value is type. The callback argument sets the callback that will be invoked when the event is dispatched.
      *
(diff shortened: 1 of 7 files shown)


saschanaz:
This sounds nice to me and perhaps we can stop emitting duplicate addEventListeners and migrate to this method.

@sandersn might want to check too.

sandersn:
How is this different from https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1624 ? My memory is that that (plus previous attempts) was too slow and also likely to cause breaks, but I can't remember if I finished testing the most recent attempt.

43081j:
@sandersn it seems the other PR is for making events generic such that `target` can be specified (`Event<T>`), while this makes `EventTarget` generic such that `addEventListener` can be strongly typed

if performance allows, we'd probably want both

whats the best way for me to perf test this?

Issue #1578, opened 2023-06-19
cloneNode definition is a bit too generic


DavidBruant:
Hi, this is my first issue here so i'm trying my best and hope it'll work\
and if it doesn't, let me know and i'll try to improve my feedback

I'm sharing this example:
```js
const el = document.createElement('input')

const el2 = el.cloneNode()

el2.querySelector('')
```
In playground: https://www.typescriptlang.org/play?#code/MYewdgzgLgBApgGxgXhgExMArgWzmKAOmACc4BDKOAUQTjwIAoByASzAAcspmBKAKH6hIsRACYU8BMQTg4AORBo4jAf3GEAjljgkAngGVEcYFBAkWfIA

In this example, the TypeScript parser sends an error: "Property 'querySelector' does not exist on type 'Node'."

However, in the real DOM (at least in Firefox, but i'm pretty sure in other browsers too), `el2` is an `HTMLInputElement` (like `el`) and this property does exist on it

The type I see in the [generated version of the dom library](https://github.com/microsoft/TypeScript/blob/5021dd87ef36e87bae39cbb5a30dabd295410920/src/lib/dom.generated.d.ts#L16076) is:
`cloneNode(deep?: boolean): Node;`

and i feel it probably should be something like:
`cloneNode<T extends Node>(deep?: boolean): T;`

or maybe something with the type of `this`, i'm not super-familiar yet with this part of TypeScript

I haven't taken the time to understand how this repo works, so i'm not sure how to transform this idea into a PR, but in any case, at this point, i'm more interested in understanding whether you agree on the problem i'm describing here

Thanks for your time reading this issue!

HolgerJeromin:
This is a duplicate of https://github.com/microsoft/TypeScript/issues/283
It had some tests to fix in the past, but they all had problems.
Ref #842 for the newest attempt.

PR #302, opened 2017-10-01
Improve cloneNode method signature

falsandtru:
Fixes https://github.com/Microsoft/TypeScript/issues/17818, https://github.com/Microsoft/TypeScript/issues/283

PR #1723, opened 2024-05-12
cloneNode Return Type Must Be Active Class, Not Base Class (Node)

andria-dev:
The `cloneNode` return type is incorrect as per #283, #302, https://github.com/microsoft/TypeScript/issues/17818 and #1578. The type before this PR indicated that `cloneNode` returned a `Node`, which then broke any class that extended `Node`. In reality, `cloneNode` will return an instance of whatever the current class type is, a.k.a. `this` (`DocumentFragment#cloneNode` returns a `DocumentFragment`, not a `Node`).

This PR solves this issue by changing the return type of `Node#cloneNode` to `this`.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -16450,7 +16450,7 @@ interface Node extends EventTarget {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/cloneNode)
      */
-    cloneNode(deep?: boolean): Node;
+    cloneNode(deep?: boolean): this;
     /**
      * Returns a bitmask indicating the position of other relative to node.
      *
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -846,6 +846,19 @@
                                 }
                             }
                         },
+                        "cloneNode": {
+                            "signature": {
+                                "0": {
+                                    "param": [
+                                        {
+                                            "name": "deep",
+                                            "overrideType": "boolean"
+                                        }
+                                    ],
+                                    "overrideType": "this"
+                                }
+                            }
+                        },
                         "removeChild": {
                             "signature": {
                                 "0": {


andria-dev:

andria-dev:
@HolgerJeromin, I will test that later today. But that issue should be fixed as we aren't manually overriding each subclass'es `cloneNode` return type anymore, just the original `Node#cloneNode` return type.

saschanaz:
Manual overridding was a workaround to reduce performance problem: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/220#discussion_r108237579

andria-dev:
Ah, my mistake, @saschanaz. I missed that. If there's no possible workaround that is both performant and covariant, what is the project's focus, performance or accuracy? Making the change to `this` locally works fine for me performance-wise, but I'm not exactly running on a low-end computer either.

saschanaz:
You can run `tsc --diagnostics --lib es5 baselines/dom.generated.d.ts` and see how the numbers increase before and after the patch 👍

andria-dev:
@saschanaz I took 50 samples before and after the change and averaged each sample set.

## Results

Version | Files | Lines | Identifiers | Symbols | Types | Instantiations | Memory used (Kilobytes) | I/O read (seconds) | I/O write (seconds) | Parse time (seconds) | Bind time (seconds) | Check time (seconds) | Emit time (seconds) | Total time (seconds)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Before | 177 | 89627 | 78827 | 95960 | 44887 | 56733 | 164941.94 | 0.021 | 0 | 0.3876 | 0.1182 | 0.9082 | 0 | 1.4132
After | 177 | 89627 | 78825 | 96281 | 45038 | 57035 | 166090.88 | 0.022 | 0 | 0.3862 | 0.1168 | 0.912 | 0 | 1.4172

So it looks like an average total increase of 4 milliseconds on my system.

## Methodology

The results were aggregated with some PowerShell. I ran the following script under the main branch, and then I changed `before.csv` to `after.csv` and ran it under this branch. After that, I used LibreOffice Calc's `AVERAGE` function to calculate the averages of each set of diagnostic runs. I made sure I wasn't doing anything on my computer during either set so that they were as consistent as possible.

```powershell
Write-Output "Files,Lines,Identifiers,Symbols,Types,Instantiations,Memory used,I/O read,I/O write,Parse time,Bind time,Check time,Emit time,Total time" | Out-File -FilePath ..\before.csv; 1..50 | ForEach-Object { (npx tsc --diagnostics --lib es5 baselines/dom.generated.d.ts | ForEach-Object { ($_ -Split ':')[1].Trim() -ireplace '(s|k)' }) -Split "`n" -Join "," | Out-File -Append -FilePath ..\before.csv }
```

andria-dev:
@HolgerJeromin I did a little reproduction, and it looks like `HTMLElement` type parameters are still covariant with this change.

```typescript
// Same type parameter as Tablesorter.
interface A<T = HTMLElement> {
	x?: T;
}
type An<T = HTMLElement> = A<T>;

type ANode = A<Node>;
type AnElement = An<Element>;
type AnHTMLElement = An;
type AnHTMLAnchorElement = An<HTMLAnchorElement>;

let node: ANode = {};
let element: AnElement = {};
let htmlElement: AnHTMLElement = {};
let htmlAnchorElement: AnHTMLAnchorElement = {};

// All of the return types are correct!
let clonedNode: Node = node.x?.cloneNode()!;
let clonedElement: Element = element.x?.cloneNode()!;
let clonedHtmlElement: HTMLElement = htmlElement.x?.cloneNode()!;
let clonedHtmlAnchorElement: HTMLAnchorElement = htmlAnchorElement.x?.cloneNode()!;

htmlElement = htmlAnchorElement; // Covariance works! 👍🏻
htmlElement = element; // Contravariance doesn't work (as expected).
```

Shows an error with the contravariance example and nothing else.

andria-dev:
@saschanaz Just wanted to follow up with a ping. This was my conclusion:

> So it looks like an average total increase of 4 milliseconds on my system.

A 4 millisecond increase seems much more acceptable than [the previous 700 millisecond increase](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/166#issue-190247907). If you have any other concerns please let me know.

callionica:
I'm very happy to see work done on this and since it looks like perf impact is minimal, hope this will get merged soon. This would be a huge improvement for front end devs using typescript.

PR #1729, opened 2024-05-23
add duplex property in RequestInit fixes #1483

### LINKED ISSUE #1483: `#duplex` is missing from `RequestInit`
The Fetch WG introduced a new `duplex` property to `RequestInit`[^1] back in July that [undici](https://npm.im/undici) is now abiding by when making payload requests with a `ReadableStream`[^2]. They have added this new `duplex` property into their own `RequestInit` types, but as this is now a part of the core `fetch` specification it should be included here.

https://fetch.spec.whatwg.org/#request-class

[^1]: https://github.com/microsoft/TypeScript-DOM-lib-generator
[^2]: https://github.com/nodejs/node/issues/46221

thecotne:


thecotne:

PR #1739, opened 2024-06-11
Allow `null` candidate in `RTCPeerConnection.addIceCandidate`

### LINKED ISSUE #1738: RTCPeerConnection.addIceCandidate cannot be called with `null`
In WebRTC, the `icecandidate` event’s [`candidate`][] property is either “an `RTCIceCandidate` object … or `null` to indicate that there are no further candidates”. This currently has the correct type in `dom.d.ts`:

```ts
interface RTCPeerConnectionIceEvent extends Event {
    readonly candidate: RTCIceCandidate | null;
}
```

`null` has the same special meaning on the receiver side of the ICE gathering process. The [`addIceCandidate`][] is used both to submit a candidate from the remote peer, or to convey the end-of-candidates signal “[i]f no candidate object is specified, or its value is null”.

But it’s not currently valid to pipe a `null` from the `icecandidate` event to `addIceCandidate`:

```ts
interface RTCPeerConnection extends EventTarget {
    addIceCandidate(candidate?: RTCIceCandidateInit): Promise<void>;
}
```

> *error TS2345:* Argument of type `null` is not assignable to parameter of type `RTCIceCandidateInit | undefined`

[`candidate`]: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnectionIceEvent/candidate
[`addIceCandidate`]: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate

silverlyra:
Add an override for [`addIceCandidate`][]’s `candidate` parameter to allow it to be `null`, which is a special value indicating that ICE gathering is complete.

> If no `candidate` object is specified, **or its value is `null`**, an end-of-candidates signal is sent to the remote peer *[emphasis added]*

Fixes #1738.

[`addIceCandidate`]: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -18150,9 +18150,9 @@ interface RTCPeerConnection extends EventTarget {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/signalingState) */
     readonly signalingState: RTCSignalingState;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/addIceCandidate) */
-    addIceCandidate(candidate?: RTCIceCandidateInit): Promise<void>;
+    addIceCandidate(candidate?: RTCIceCandidateInit | null): Promise<void>;
     /** @deprecated */
-    addIceCandidate(candidate: RTCIceCandidateInit, successCallback: VoidFunction, failureCallback: RTCPeerConnectionErrorCallback): Promise<void>;
+    addIceCandidate(candidate: RTCIceCandidateInit | null, successCallback: VoidFunction, failureCallback: RTCPeerConnectionErrorCallback): Promise<void>;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/addTrack) */
     addTrack(track: MediaStreamTrack, ...streams: MediaStream[]): RTCRtpSender;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/addTransceiver) */
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -2173,7 +2173,7 @@
                                     "param": [
                                         {
                                             "name": "candidate",
-                                            "overrideType": "RTCIceCandidateInit | null"
+                                            "nullable": true
                                         }
                                     ]
                                 },
@@ -2181,7 +2181,7 @@
                                     "param": [
                                         {
                                             "name": "candidate",
-                                            "overrideType": "RTCIceCandidateInit | null"
+                                            "nullable": true
                                         }
                                     ],
                                     "deprecated": true
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -2169,7 +2169,21 @@
                     "method": {
                         "addIceCandidate": {
                             "signature": {
+                                "0": {
+                                    "param": [
+                                        {
+                                            "name": "candidate",
+                                            "overrideType": "RTCIceCandidateInit | null"
+                                        }
+                                    ]
+                                },
                                 "1": {
+                                    "param": [
+                                        {
+                                            "name": "candidate",
+                                            "overrideType": "RTCIceCandidateInit | null"
+                                        }
+                                    ],
                                     "deprecated": true
                                 }
                             }


silverlyra:

saschanaz:
Passing `null` is same as passing `undefined` in this case per https://webidl.spec.whatwg.org/#js-dictionary. I'm not sure we really need a special case here?

saschanaz:
We could instead allow null in general this case, hmm.

silverlyra:
> We could instead allow null in general this case, hmm.

If there’s a general fix possible in the generator itself, that might be a good idea – I don’t have the context to help with that myself, though. However, in lieu of that

> I'm not sure we really need a special case here?

Well, the types right now are incorrect. If there’s no general fix, I do think it’s appropriate to declare these parameters as `nullable`.

The [simplest `RTCPeerConnection` sample](https://webrtc.github.io/samples/src/content/peerconnection/pc1/) just creates two peer connections in the same browser tab. If I reproduce the relevant parts of [its code](https://github.com/webrtc/samples/blob/gh-pages/src/content/peerconnection/pc1/js/main.js#L181) in TypeScript:

```ts
let pc1 = new RTCPeerConnection();
pc1.addEventListener('icecandidate', e => onIceCandidate(pc1, e));

let pc2 = new RTCPeerConnection();
pc2.addEventListener('icecandidate', e => onIceCandidate(pc2, e));

async function onIceCandidate(pc: RTCPeerConnection, event: RTCPeerConnectionIceEvent) {
  let target = pc === pc2 ? pc1 : pc2;
  await target.addIceCandidate(event.candidate);
}
```

it [doesn’t build](https://www.typescriptlang.org/play/?#code/DYUwLgBADgxgjBAvBAdiA7hASgFQMIAKIIATngPYpoxgCWlAFAJQDcAULHAHQCGAJnwCiANxAowAGVoBnMGNIMA5LRggYPFH1p8ecxQBoIIJAD4IlAJKq8GrTrkNOhkE1Zs2oSLABMSVBmx8IlIKKjU6RjdOXgERMUkZOTQSJRU1W21dEAMjU3MUKxAbTUyHH2dXdjYeaQBPFBgIADMAVwaIlHzC4rssxxgALkDCYjJKag7nUXEh3BGQ8fD6AtU48SYIAG82CAhPCDAeEgBzcD9YJERkHwgAfmh4CCGfdl2edB5aSEOT8Bi+boZewgBggaZgLjqErAtwAXyAA):

> Argument of type `RTCIceCandidate | null` is not assignable to parameter of type `RTCIceCandidateInit | undefined`.
> Type `null` is not assignable to type `RTCIceCandidateInit | undefined`.

nikwen:
Thanks for the fix! Would love to see it merged as I ran into the same issue.

saschanaz:
We should make null work for dictionaries in general. That wouldn't be straightforward, so for now I'll merge this.

LGTM

saschanaz:
Oops, merge conflicts.

nikwen:
Thanks, @saschanaz!

I don't think it's a merge conflict. I tried merging it locally and it worked perfectly. I think the GitHub actions bot ran into a network issue.

From the log:

```
Creating comments and merging
Merging (or commenting) failed:
Error: HttpError: Resource not accessible by integration
Error: Failed to merge
```

saschanaz:
Local merging somehow is different from GitHub merging time to time. Can you push your merge here?

saschanaz:
LGTM (retrying) but other PR worked, so if this still doesn't work then I think it's merge problem.

saschanaz:
Perhaps auto merge needs some permission? @jakebailey might know...

nikwen:
Got it! How/where do you want me to push? (This is not my PR.)

I did the following to merge locally:

```bash
git clone git@github.com:microsoft/TypeScript-DOM-lib-generator.git
cd TypeScript-DOM-lib-generator
git remote add silverlyra git@github.com:silverlyra/TypeScript-DOM-lib-generator.git
git fetch silverlyra
git merge silverlyra/null-ice-candidate
```

jakebailey:
I tightened perms on a bunch of stuff; it's possible this broke (I will look tomorrow)

silverlyra:
Hello! Glad to see activity here 😄 I’ve rebased

saschanaz:
Test is failing, can you check whether it passes for you? (I just pushed #1786, so might want to merge that too)

jakebailey:
I did nothing but https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1787 didn't fail to merge, so maybe this was a temporary problem.

saschanaz:
LGTM

saschanaz:
Huh, seems so!

saschanaz:
Might also because @silverlyra rebased it. Worth checking with non-rebased PRs...

nikwen:
Thanks for working on this, everyone! :raised_hands:

PR #1740, opened 2024-06-12
Add override for `import.meta.resolve()`

Renegade334:
Specification: [HTML Standard 8.1.6.5.1](https://html.spec.whatwg.org/multipage/webappapis.html#hostgetimportmetaproperties)
Reference: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve)
Reference: microsoft/TypeScript#58828

Converting `ImportMeta` to an interface seems fairly unavoidable, but I don't think it should break anything.


saschanaz:
Maybe the extra field can have a lambda function type to make it simpler and keeping dictionary?

Renegade334:
> Maybe the extra field can have a lambda function type to make it simpler while keeping dictionary?

Considered this, but this approach leads to interface merging errors in places like [`@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/68a06dbd340bd1a6442f1d22d01ac071e7ec9dd1/types/node/module.d.ts#L307). Because of this, it really needs to be defined as a method rather than a property.

saschanaz:
Oh no, but ImportMeta is not an interface at all. 😞

Can you at least add some comment why it should be an interface?

Renegade334:
The only alternative I can see is the deliciously hacky e341817...

saschanaz:
I guess interface is fine compared to that 😂

saschanaz:
(But please add some comment in addedTypes)

inputfiles/addedTypes.jsonc
```diff
@@ -449,6 +449,38 @@
     },
     "interfaces": {
         "interface": {
+            // ImportMeta is not a true DOM interface, but we are forced to declare it as one in order to allow method definitions.
```
saschanaz, line 452: Mentioning @types/node would be helpful because otherwise we wouldn't need a "method".

HolgerJeromin:
ref https://github.com/microsoft/TypeScript/pull/23327 where the "interface" was added to be mergeable.

saschanaz (commented):
Thanks!

inputfiles/addedTypes.jsonc
```diff
@@ -449,6 +449,40 @@
...
+                },
+                "methods": {
+                    "method": {
+                        "resolve": {
+                            "name": "resolve",
+                            "signature": [
+                                {
+                                    "param": [
+                                        {
+                                            "name": "moduleName",
```
saschanaz, line 475: One nit, as the spec says "specifier", can you also match this param name with that?

> 4. Let steps be the following steps, given the argument *specifier*:

Renegade334: I went with MDN's nomenclature, but happy to go off the spec if that's preferable.

HolgerJeromin: I use it to find the folder of the script via
`import.meta.resolve("./")`
So IMHO both are not always matching but `specifier` is generic enough.

HolgerJeromin (approved):

PR #1742, opened 2024-06-19
Change return type of CredentialsContainer.create for WebAuthn

asmsuechan:
Hi developers,

The return value for CredentialsContainer.create is one of `FederatedCredential`, `PasswordCredential`, and `PublicKeyCredential` 
according to https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/create#return_value

The return types are subclass of Credential. So I changed using `PublicKeyCredential` from `Credential`. Note that `FederatedCredential` and `PasswordCredential` haven't been implemented yet, therefore, I didn't use them.

Thank you for reading my PR.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5920,7 +5920,7 @@ declare var Credential: {
  */
 interface CredentialsContainer {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/CredentialsContainer/create) */
-    create(options?: CredentialCreationOptions): Promise<Credential | null>;
+    create(options?: CredentialCreationOptions): Promise<PublicKeyCredential | null>;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/CredentialsContainer/get) */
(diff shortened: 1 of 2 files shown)


asmsuechan:

saschanaz:
This can be a problem when other types of credentials, maybe there should be additional overload based on the value of `credential`?

asmsuechan:
@saschanaz  Thank you for your comment! Yes, what you say is true. `Promise<PublicKeyCredential | FederatedCredential | PasswordCredential | null>` is the correct interface. However, `FederatedCredential` and `PasswordCredential` are not yet implemented. So I added only `PublicKeyCredential` because it's better than the current implementation `Promise<Credential | null>`.

If it's unacceptable, I'll add the other interfaces above in this PR. I would love to know the manners of this repo first.

asmsuechan:
Oops, FederatedCredential and PasswordCredential is supported only chrome.

* https://github.com/mdn/browser-compat-data/blob/main/api/FederatedCredential.json
* https://github.com/mdn/browser-compat-data/blob/main/api/PasswordCredential.json

Issue #1486, opened 2023-01-31
navigator.vibrate is typed as always being defined, but is undefined in Safari


FuXiuHeng:
The `navigator.vibrate` API is currently always defined. However, according to the [navigator.vibrate MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/vibrate), this is currently not supported in Safari:

<img width="400" alt="Screen Shot 2023-01-31 at 1 58 36 pm" src="https://user-images.githubusercontent.com/19201517/215651787-90ccc64f-33a6-4682-9c17-bae434529ead.png">

I'm proposing to update the `Navigator` interface to:
```ts
interface Navigator {
    // Proposed change from being required to being optional
    vibrate?: (pattern: Iterable<number>) => boolean;
}

```

orta:
Yeah, I agree, this is probably the right call (having made this exact mistake myself also)

HolgerJeromin:
https://github.com/microsoft/TypeScript-DOM-lib-generator#why-is-my-fancy-api-still-not-available-here
We have the policy to need two js engine support. 
Perhaps an API with only two engines chould be marked as optional...

saschanaz:
Yeah, I'd like to have some policy rather than doing special treatment to a specific API.

orta:
Hah, yeah, you're both not wrong there - should probably not do it

saschanaz:
Oh I mean, I'm open to this but I'd like to modify the policy so that things can be consistent for other APIs too.

That said, not sure how the impact would be.

HolgerJeromin:
Yeah, my suggestion was to make it optional in the emitter code (where we handle the number of implementations anyway).

> That said, not sure how the impact would be.

This would probably make quite a few APIs optional with some breakings... Some good, some bad (depending on the target audience of the customer code)

RyanCavanaugh:
I think our longstanding policy of marking things as fully present once they're in the relevant spec plus two major browsers is a sensible one, Safari's eccentricities notwithstanding. I realize Safari isn't exactly niche, but "a browser that exists doesn't support a spec'd, otherwise-available API" is not a road that goes to a good place.

guillaumebrunerie:
[...]

PR #1748, opened 2024-06-26
Make Navigator.canShare and Navigator.share methods optional

rjimenezda:
Hi!

I've recently worked with `navigator.share` and `navigator.canShare` and have noticed that the types report them as non-optional.

I saw #837 where it's stated that the w3c says they are non-optional, but the truth is that most desktop browsers do not implement them and are only supported on secure contexts, so they can be undefined in certain cases. 

I'm not sure if there's something else that should be changed, let me know.


chemitaxis (approved):

chemitaxis:
Good job @rjimenezda LGTM!

saschanaz:
As stated in #1486 I'd rather not do this only for this specific API. We could make any not-supported-everywhere things as optional but such breaking change needs agreement from TS team. Discussion can go on in #1486.

PR #1755, opened 2024-07-03
Update the `URL.canParse` and `URL.parse` types

petamoriken:
Modified the `URL.canParse` and `URL.parse` types to match the URL constructor type.


saschanaz:
LGTM

PR #1757, opened 2024-07-14
Add /** @deferred */ comments to deferred callback parameters

ahejlsberg:
This PR adds `/** @deferred */` comments to deferred callback parameters in preparation for https://github.com/microsoft/TypeScript/pull/58729.

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -230,6 +230,8 @@ export function emitWebIdl(
     getParentsWithConstant,
   );
 
+  const deferredParameterNames = new Set(["handler", "listener", "callback", "successCallback", "errorCallback", "failureCallback"]);
+
   switch (iterator) {
     case "sync":
       return emitES6DomIterators();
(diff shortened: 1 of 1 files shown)


saschanaz:
You also need to `npm run build && npm run baseline-accept`.

saschanaz:
(Looks good but feel free to merge when the TS PR is ready)

saschanaz:
One question, why not use `deferred` modifier when we are using TypeScript here?

saschanaz:
And another question, is it the requirement here for the callback to be called asynchronously? What happens if those callbacks can be called both sync/async?

ahejlsberg:
> And another question, is it the requirement here for the callback to be called asynchronously?

The `deferred` modifier (and the `/** @deferred */` JSDoc comment) asserts that the callback is never called synchronously. If a synchronous call is a possibility, then the parameter shouldn't have the modifier.

PR #1759, opened 2024-07-21
Improve some types from Deno

petamoriken:
I am working on making the Deno types match the DOM types in this repository as closely as possible (https://github.com/denoland/deno/pull/24599). I send a PR for the types that I think should be included in the TypeScript side.

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -2697,6 +2697,18 @@
                 }
             },
             "ReadableStreamBYOBReader": {
+                "constructor": {
+                    "signature": {
+                        "0": {
+                            "param": [
+                                {
+                                    "name": "stream",
+                                    "overrideType": "ReadableStream<Uint8Array>"
+                                }
+                            ]
+                        }
+                    }
+                },
                 "methods": {
                     "method": {
                         "read": {
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -2987,6 +2987,38 @@
                     }
                 }
             },
+            "CompressionStream": {
+                "properties": {
+                    "property": {
+                        "readable": {
+                            "name": "readable",
+                            "readonly": true,
+                            "overrideType": "ReadableStream<Uint8Array>"
+                        },
+                        "writable": {
+                            "name": "writable",
+                            "readonly": true,
+                            "overrideType": "WritableStream<BufferSource>"
+                        }
+                    }
+                }
+            },
+            "DecompressionStream": {
+                "properties": {
+                    "property": {
+                        "readable": {
+                            "name": "readable",
+                            "readonly": true,
+                            "overrideType": "ReadableStream<Uint8Array>"
+                        },
+                        "writable": {
+                            "name": "writable",
+                            "readonly": true,
+                            "overrideType": "WritableStream<BufferSource>"
+                        }
+                    }
+                }
+            },
             "TextDecoderStream": {
                 "properties": {
                     "property": {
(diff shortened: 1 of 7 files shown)


inputfiles/overridingTypes.jsonc
```diff
@@ -2573,7 +2573,7 @@
                                 },
                                 {
                                     "name": "strategy",
-                                    "overrideType": "{ highWaterMark?: number }"
+                                    "overrideType": "{ highWaterMark?: number; size?: undefined }"
```
saschanaz, line 2576: Why an unused property? Who would be happy with this? Can you add a comment about that? (it's jsonc so it supports comment)

saschanaz (commented):
Mostly looks good, thanks!

inputfiles/overridingTypes.jsonc
saschanaz: Hmm okay, but doesn't TypeScript already show error for excessive fields?

petamoriken: It's a minor detail, but we can make `new ReadableStream({ type: “bytes” }, { size: undefined })` be inferred as `ReadableStream<Uint8Array>`.



https://www.typescriptlang.org/play/?#code/MYewdgzgLgBATgRhgXhmApgdxgJXQQwBN8AjAG3QGUo4CBbACgG8YoBPAB3QC4YAiEmyjoIfGAF8ANDCbiAlACgA9EphqYAPQD8ChaEiw4AJhRosuAsXJUa9Zq049+g4aInSWEAJYAvJwFcwQnQAMy8MQglFFXVNHSA

saschanaz: Hmm that's interesting... Is that something frequently done by users?

petamoriken: In general, I believe that implementations that set properties to `undefined` are often used because it's easy to write, while spread syntax is required to implement empty properties in some cases.

```js
const option1 = {
  foo: "foo",
  bar: setBar ? "bar" : undefined,
};

const option2 = {
  foo: "foo",
  ...(setBar ? { bar: "bar" } : {}),
};
```

saschanaz: But what is the hypothetical setBar here? readAsBytes? I would expect either always pass size or never pass size here than your conditional pass example, unless the caller also has the hypothetical setBar for `type`, which would be kinda complex setup to create ReadableStream. Especially when the way to read the resulting stream would be significantly different.

petamoriken: I think it is nonsense to consider whether or not to give `?: undefined` per API; it is consistent to allow for the `undefined` assignment, as in TypeScript's `Partial`. However, if you have strong concerns here, I am not opposed to revert.

saschanaz: > I think it is nonsense to consider whether or not to give `?: undefined` per API; it is consistent to allow for the `undefined` assignment, as in TypeScript's `Partial`. However, if you have strong concerns here, I am not opposed to revert.

Agreed that it's nonsense to decide per API (or at least hard to maintain), and I think it should not exist for consistency sake because it would be the first case if we add this here. That is unless we have a strong proof that there's many users who do setBar-style thing here (which I think doesn't make sense at least for this one).

petamoriken: > I think it should not exist for consistency sake because it would be the first case if we add this here. That is unless we have a strong proof that there's many users who do setBar-style thing here (which I think doesn't make sense at least for this one).

Agreed. I reverted this change.

saschanaz: Thanks, I'll see whether #1713 is ready enough to be merged and when I can land this.

saschanaz:
#1713 should probably just rebuild if merge conflict happens, so I'll just proceed here. LGTM, thanks!

PR #1761, opened 2024-07-28
Remove some compat quirk

petamoriken:
https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1713#issuecomment-2248890913

> types/web also has some compat quirk that had to mismatch with the spec, e.g. ChildNode being only a mixin interface inherited by some Node variants but types/web had to make it inherit from Node, reversing the inheritance, because back in the day we thought maybe we should put the compat first before correctness.
>
> But if https://github.com/microsoft/TypeScript/pull/58222 didn't break DefinitelyTyped then maybe it's a good chance to remove one quirk.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -20413,7 +20413,7 @@ declare var SVGSwitchElement: {
  *
  * [MDN Reference](https://developer.mozilla.org/docs/Web/API/SVGSymbolElement)
  */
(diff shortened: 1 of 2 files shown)


saschanaz:
Oh I said the quirk #1713 is removing, not this quirk. I'm much less confident that this would be compatible with the ecosystem. But you can try submitting a draft PR into microsoft/TypeScript with this change and see what it catches.

PR #1762, opened 2024-07-28
Add [Symbol.toStringTag] property to all interfaces

### LINKED ISSUE #1641: `Symbol.toStringTag` is missing on `File`, `Blob` and `FormData`

petamoriken:
Fixes #1641
Past PR #1699


saschanaz:
I wonder we should instead have WebIDLInterface interface with toStringTag and let others extend it. Doing so could just fold the extra member in existing .iterable.d.ts without extra files. (because .iterable.d.ts is effectively for es2015 which introduced toStringTag). This would allow easier change when the Web IDL spec ever adds any other well-known symbol in interface binding.

```ts
// dom.generated.d.ts
interface WebIDLInterface {}

interface AbortSignal extends WebIDLInterface {} // Let others extend it

// dom.iterable.generated.d.ts - augment the interface
interface WebIDLInterface { /* */ } 
```

petamoriken:
IMHO, it's weird to have unrelated types in `*.iterable.d.ts`, so separate files is better.

jakebailey:
Theoretically this is now unblocked, so long as it's only applied to TS 6.0.

PR #1765, opened 2024-07-28
Add midi and storage-access to PermissionName

### LINKED ISSUE #1758: Type "midi" is not assignable to type "PermissionName"
## Description
Existing type PermissionName is missing "midi", actual type:
```
type PermissionName = "geolocation" | "notifications" | "persistent-storage" | "push" | "screen-wake-lock" | "xr-spatial-tracking";
```

According to [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query#name), "midi" value is supported by at least 3 browser engines.

## Solution
Add "midi" to PermissionName type here:
https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/2939d4ffe6c500fb32de334d4d23eb877d5bfe3a/inputfiles/addedTypes.jsonc#L335-L354

## Typescript version
5.4.5

## Workaround
Force type:
```
navigator.permissions.query({ name: "midi" as PermissionName })
```

saschanaz:
And removing xr-spatial-tracking which is still Blink only.\

Closes #1758


saschanaz:
LGTM

PR #1786, opened 2024-08-23
Use printUnifiedDiff for test

saschanaz:
printInlineDiff prints all the text which doesn't make sense for our use case.


saschanaz:
LGTM

PR #1806, opened 2024-09-22
fix: `DedicatedWorkerGlobalScope.close()` and `SharedWorkerGlobalScope.close()` return type - `never`

turansky:
* `DedicatedWorkerGlobalScope.close()` - [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope/close)
* `SharedWorkerGlobalScope.close()` - [MDN](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorkerGlobalScope/close)


turansky:

saschanaz:
This is not necessarily true:

```js
new Worker(URL.createObjectURL(new Blob([`
  self.close();
  console.log("closed?");
`])))

// You can see "closed" on devtools console
```

turansky:
@saschanaz thank you for example

PR #1837, opened 2024-11-02
Manually remove some deprecations

### LINKED ISSUE #1642: Why is pageYOffset marked as deprecated?
`scrollY` is an alias of `pageYOffset`, but there's no documentation anywhere else (as far as I'm aware) that `pageYOffset` is deprecated or that it is a legacy api, but it is marked as deprecated in the definitions.

Why was [this](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1082#pullrequestreview-719665841) decision made? 

References to `pageYOffset` and `scrollY`:
https://developer.mozilla.org/en-US/docs/web/api/window/pageyoffset
https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollY

These references also say that `pageYOffset` has more support than `scrollY` in older browsers, and I don't seem to find any discussion about deprecating it or saying that `pageYOffset` is a legacy API os `scrollY`, all the documentations say they are both equal.

The main issue I am having with this is that IntelliJ uses these definitions in the IDE and so it's being marked as deprecated [link to issue](https://youtrack.jetbrains.com/issue/WEB-55890/wrong-deprecated-warning-for-window.pageXoffset-and-window.pageYoffset).

(I created an issue [here](https://github.com/microsoft/TypeScript/issues/56308) and they recommended I move it here.)

### LINKED ISSUE #1826: Missing properties on CSSStyleDeclaration in upstream TypeScript
We're seeing a deprecation error in our tooling when someone tries to use the property `webkitLineClamp` in their code. Error: 

```
webkitLineClamp is deprecated: This is a legacy alias of `lineClamp`.
``` 

I checked that the property is defined in https://fossies.org/linux/TypeScript/src/lib/dom.generated.d.ts and is marked @deprecated. 

```
...
 4900     /**
 4901      * @deprecated This is a legacy alias of `lineClamp`.
 4902      *
 4903      * [MDN Reference](https://developer.mozilla.org/docs/Web/CSS/-webkit-line-clamp)
 4904      */
 4905     webkitLineClamp: string;
...
```

1. Why is this property (and others) not a part of TypeScript's definition of CSSStyleDeclaration interface in the lib.dom.d.ts file? Instead I see an ~empty definition of CSSStyleDeclaration in the [dom.iterable.generated.d.ts](https://github.com/microsoft/TypeScript/blob/40caf3431986884b3da7a607de22ced5d364d9fc/src/lib/dom.iterable.generated.d.ts#L40) file instead. Am I mistaken to think this property should exist in lib.dom.d.ts? 
2. Even if the property gets defined in the right place in the future, it will be deprecated. So which property should we use instead? I don't see any property named lineClamp defined anywhere in either of the two .d.ts files in TypeScript.

saschanaz:
Fixes #1642, fixes #1826

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5255,11 +5255,7 @@ interface CSSStyleDeclaration {
      * [MDN Reference](https://developer.mozilla.org/docs/Web/CSS/justify-content)
      */
     webkitJustifyContent: string;
-    /**
-     * @deprecated This is a legacy alias of `lineClamp`.
-     *
-     * [MDN Reference](https://developer.mozilla.org/docs/Web/CSS/-webkit-line-clamp)
-     */
+    /** [MDN Reference](https://developer.mozilla.org/docs/Web/CSS/-webkit-line-clamp) */
     webkitLineClamp: string;
     /**
      * @deprecated This is a legacy alias of `mask`.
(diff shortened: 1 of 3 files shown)


saschanaz:
LGTM

PR #1839, opened 2024-11-02
Do not generate iterable signatures for static methods

### LINKED ISSUE #1803: AbortSignal improper `any` type definition
### 🔍 Search Terms

AbortSignal

### Description

Hello,
In the `AbortSignal` definition, we can see there's an `any` instance method that actually does not exist.
It should be a static method instead.
"Culprit" : https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/main/baselines/dom.iterable.generated.d.ts#L7

### Sample code

[Playground link](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBA3hAlgczAQwDYF8YF4ZgCmA7jAIIBGIATlAMLhTUgYaHUAUAlAFB+U0oAZRToMAOjRgAnhwDaSVJgC6XGAHp1MQLwbgWR2eisZJnzDKtZpgAREIQgEQsNMCgBXTBmkxCAD0TQIABoYCAALEFcMABMAcjBYCkIYDxBiQiigA)

### Documentation

[MDN Reference](https://developer.mozilla.org/docs/Web/API/AbortSignal/any_static)

### Other

Also, gentle ping on https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1162 since I intended to report this too at first.

saschanaz:
The current logic is bugged, and the correct fix is not possible because of #222.

Closes #1803

### DIFF
--- a/baselines/audioworklet.iterable.generated.d.ts
+++ b/baselines/audioworklet.iterable.generated.d.ts
@@ -2,11 +2,6 @@
 /// AudioWorklet Iterable APIs
 /////////////////////////////
 
-interface AbortSignal {
-    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/AbortSignal/any_static) */
-    any(signals: Iterable<AbortSignal>): AbortSignal;
-}
-
 interface MessageEvent<T = any> {
     /** @deprecated */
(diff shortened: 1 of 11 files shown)


saschanaz:
LGTM

Issue #1835, opened 2024-11-02
Add Literal Suggestions to `ARIAMixin` and Fallback for Unknown Values


DeepDoge:
The `ARIAMixin` currently uses `string | null` for each aria property type. The reason it’s just `string` rather than a union of valid values is to allow for future, yet unknown, valid values.

However, in TypeScript, we can prevent `string` from collapsing with literal unions by using `string & {}`. For instance, `ariaCurrent` could be typed as `"page" | "step" | "location" | (string & {}) | null`, suggesting valid options like `"page"`, `"step"`, and `"location"` while still allowing any other string as a fallback.

There is a related closed issue here: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1184. While that issue suggested disallowing "unknown" values, my suggestion would still permit them but make the known options more discoverable in TypeScript.

saschanaz:
Similar suggestion with `string & {}` happened somewhere else, but I think in 2024 the situation is different, we already have several hardcoded enums. PR welcome.

PR #1840, opened 2024-11-03
Added strict typings for `ARIAMixin` with literals

DeepDoge:
Added stricter typings for `ARIAMixin` with a fallback to also allow any other string. Also used `${bigint}` and `${number}` to type, integer and number only properties.

I wasn't sure if I should create a new `enum` type for each ARIA property, so I inlined them in the `overrideType`.

Fixes the issue: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1835

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2404,89 +2404,89 @@ interface ANGLE_instanced_arrays {
 
 interface ARIAMixin {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaAtomic) */
-    ariaAtomic: string | null;
+    ariaAtomic: ("false" | "true" | (string & {})) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaAutoComplete) */
-    ariaAutoComplete: string | null;
+    ariaAutoComplete: ("inline" | "list" | "both" | "none" | (string & {})) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaBrailleLabel) */
-    ariaBrailleLabel: string | null;
+    ariaBrailleLabel: ("string" | (string & {})) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaBrailleRoleDescription) */
-    ariaBrailleRoleDescription: string | null;
+    ariaBrailleRoleDescription: ("string" | (string & {})) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaBusy) */
-    ariaBusy: string | null;
+    ariaBusy: ("true" | "false" | (string & {})) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaChecked) */
-    ariaChecked: string | null;
+    ariaChecked: ("true" | "mixed" | "false" | "undefined" | (string & {})) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaColCount) */
-    ariaColCount: string | null;
+    ariaColCount: `${bigint}` | "" | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaColIndex) */
-    ariaColIndex: string | null;
+    ariaColIndex: `${bigint}` | "" | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaColIndexText) */
     ariaColIndexText: string | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaColSpan) */
-    ariaColSpan: string | null;
+    ariaColSpan: `${bigint}` | "" | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaCurrent) */
-    ariaCurrent: string | null;
+    ariaCurrent: ("page" | "step" | "location" | "date" | "time" | "true" | "false" | (string & {})) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaDescription) */
     ariaDescription: string | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaDisabled) */
-    ariaDisabled: string | null;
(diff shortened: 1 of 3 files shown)


DeepDoge:

baselines/dom.generated.d.ts
```diff
@@ -2404,89 +2404,89 @@ interface ANGLE_instanced_arrays {
 
 interface ARIAMixin {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaAtomic) */
-    ariaAtomic: string | null;
+    ariaAtomic: ("false" | "true" | (string & {})) | null;
```
saschanaz, line 2407: Please don't use `string & {}`. Extra context: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1467#issuecomment-1377655691

baselines/dom.generated.d.ts
```diff
@@ -2404,89 +2404,89 @@ interface ANGLE_instanced_arrays {
...
-    ariaValueNow: string | null;
+    ariaValueNow: `${number}` | "" | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/ariaValueText) */
     ariaValueText: string | null;
-    role: string | null;
+    role: ("toolbar" | "tooltip" | "feed" | "math" | "presentation" | "none" | "note" | "application" | "article" | "cell" | "columnheader" | "definition" | "directory" | "document" | "figure" | "group" | "heading" | "img" | "list" | "listitem" | "meter" | "row" | "rowgroup" | "rowheader" | "separator" | "table" | "term" | "associationlist" | "associationlistitemkey" | "associationlistitemvalue" | "blockquote" | "caption" | "code" | "deletion" | "emphasis" | "insertion" | "paragraph" | "strong" | "subscript" | "superscript" | "time" | "scrollbar" | "searchbox" | "separator" | "slider" | "spinbutton" | "switch" | "tab" | "tabpanel" | "treeitem" | "combobox" | "menu" | "menubar" | "tablist" | "tree" | "treegrid" | "banner" | "complementary" | "contentinfo" | "form" | "main" | "navigation" | "region" | "search" | "alert" | "log" | "marquee" | "status" | "timer" | "alertdialog" | "dialog" | (string & {})) | null;
```
saschanaz, line 2489: Okay, this one is huge enough that I think exporting as an enum would be better.

inputfiles/overridingTypes.jsonc
```diff
@@ -286,6 +286,138 @@
...
+                        },
+                        "ariaHidden": {
+                            "overrideType": "\"true\" | \"false\" | \"undefined\" | (string & {})"
+                        },
+                        // "ariaKeyShortcuts": {
+                        //     "overrideType": ""
+                        // },
+                        // "ariaLabel": {
+                        //     "overrideType": ""
+                        // },
```
saschanaz, line 346: Please remove commented ones.

inputfiles/overridingTypes.jsonc
```diff
@@ -286,6 +286,138 @@
                         }
                     }
                 }
+            },
+            "ARIAMixin": {
+                "properties": {
+                    "property": {
+                        "ariaColCount": {
+                            "overrideType": "`${bigint}` | \"\""
```
saschanaz, line 294: Does bigint have any difference than number when used here? Quickly tried but number doesn't reject huge number either.

inputfiles/overridingTypes.jsonc
```diff
@@ -286,6 +286,138 @@
...
+                        // }
+                        "role": {
+                            "overrideType": "\"toolbar\" | \"tooltip\" | \"feed\" | \"math\" | \"presentation\" | \"none\" | \"note\" | \"application\" | \"article\" | \"cell\" | \"columnheader\" | \"definition\" | \"directory\" | \"document\" | \"figure\" | \"group\" | \"heading\" | \"img\" | \"list\" | \"listitem\" | \"meter\" | \"row\" | \"rowgroup\" | \"rowheader\" | \"separator\" | \"table\" | \"term\" | \"associationlist\" | \"associationlistitemkey\" | \"associationlistitemvalue\" | \"blockquote\" | \"caption\" | \"code\" | \"deletion\" | \"emphasis\" | \"insertion\" | \"paragraph\" | \"strong\" | \"subscript\" | \"superscript\" | \"time\" | \"scrollbar\" | \"searchbox\" | \"separator\" | \"slider\" | \"spinbutton\" | \"switch\" | \"tab\" | \"tabpanel\" | \"treeitem\" | \"combobox\" | \"menu\" | \"menubar\" | \"tablist\" | \"tree\" | \"treegrid\" | \"banner\" | \"complementary\" | \"contentinfo\" | \"form\" | \"main\" | \"navigation\" | \"region\" | \"search\" | \"alert\" | \"log\" | \"marquee\" | \"status\" | \"timer\" | \"alertdialog\" | \"dialog\" | (string & {})"
```
saschanaz, line 417: Where did you get this list? Have you checked everything here has multiple implementation support?

saschanaz (changes requested):
This needs some unit test, take a look here: https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/main/unittests/files/autocomplete.ts

inputfiles/overridingTypes.jsonc
DeepDoge: `${bigint}` only allows integers, `${number}` also allows floats.
So I used `${bigint}` for integer types and `${number}` for number types.

Also I allowed empty string `""` on these. Thinking people might use empty string instead of null.
But now since I'm also removing `string & {}` from enums, making them stricter. Should I also remove the empty string `""`?

DeepDoge:
Removed commented items. Removed `string & {}` from types, and empty string from number and integer types.
Updated roles based on [W3C](https://w3c.github.io/aria) instead of MDN, filtered out deprecated and abstract ones.
Added type for `ariaInvalid`. MDN was missing documention for `ariaInvalid`, but has one for `aria-invalid` attribute. So I missed it as well since I was looking into Element API on MDN. `ariaInvalid` has full browser support already.

DeepDoge:
@saschanaz Should I include roles listed at [https://www.w3.org/TR/dpub-aria-1.1](https://www.w3.org/TR/dpub-aria-1.1)?

myasonik:
Stumbled across this PR - seems like a great addition! @saschanaz Anything you need before this can be merged? 

@DeepDoge I wouldn't include the dpub roles as they are only a proposed recommendation and are not fully baked.

PR #1859, opened 2024-12-04
Delete TS PR workflow

jakebailey:
We don't use it, and I need to reissue a PAT in a way that wouldn't allow this to work.


saschanaz:
>reissue a PAT in a way that wouldn't allow this to work.

What exactly would such change disallow?

jakebailey:
I am scoping tokens down such that the token used here can only push to the bot's fork of the repo. We'll see if it works the next time it has to do something.

PR #1878, opened 2025-01-13
Add object form of `Ed25519` overload for `SubtleCrypto#generateKey`

steveluscher:
As a follow up for #1594, I'd like to add the object form of the Edwards curve identifier (`{ name: "Ed25519" }`) to complement the string form (`"Ed25519"`).

This will prevent having to constantly refine the return type of `generateKey` when all this API ever returns is a `CryptoKeyPair` when used with this algorithm.

```ts
const key = await crypto.subtle.generateKey(
  { name: 'Ed25519' },
  false, /* extractable */
  ['sign', 'verify'],
) as CryptoKeyPair;
```

See #1594 for more.


saschanaz:
LGTM for now, but this still needs more general form.

PR #1882, opened 2025-01-17
Override MessageEventTarget's events' this-parameter type

sandersn:
This adds the ability to override the `this` parameter of a type's event handlers, which I need for MessageEventTarget to avoid breaking code.
Technically, I believe `this: this` is correct for all event handlers, but it's expensive to compile and confusing to understand so I'd prefer to use it only when needed.

I don't know if this is the right way to do this in the generator code. Corrections welcome.

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -684,9 +684,9 @@ interface MessageEventTargetEventMap {
 
 interface MessageEventTarget {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/DedicatedWorkerGlobalScope/message_event) */
-    onmessage: ((this: MessageEventTarget, ev: MessageEvent) => any) | null;
+    onmessage: ((this: this, ev: MessageEvent) => any) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/DedicatedWorkerGlobalScope/messageerror_event) */
-    onmessageerror: ((this: MessageEventTarget, ev: MessageEvent) => any) | null;
+    onmessageerror: ((this: this, ev: MessageEvent) => any) | null;
     addEventListener<K extends keyof MessageEventTargetEventMap>(type: K, listener: (this: MessageEventTarget, ev: MessageEventTargetEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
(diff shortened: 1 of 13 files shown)


saschanaz:
Probably better to autodetect whether there's inheritance/mixin relationship and use `this` automatically in such case, wdyt? (In that case anything related to HTMLElement should be excluded for now or otherwise all the HTML event handlers would get `this`.)

saschanaz:
But in case this is urgent I think this is fine too.

sandersn:
I think that's theoretically a good idea. But .. how many inheritance/mixin relationships are there? Even though `this: this` is correct I'd prefer to default it off for performance reasons. 

If we do this, would you recommend putting the HTMLElement exclusion in overridingtypes.jsonc or hard-coded in the generator?

sandersn:
However, when I add manual exclusions, there are only 3 of them: AbstractWorker, GlobalEventHandlers and WindowEventHandlers. I'll try to come up with a property for overridingTypes.jsonc.

saschanaz (approved):
This should probably apply to interfaces inherited by others, but looks good

sandersn:
This still is still invariant and breaks on examples like partytown's tests:

```ts
interface PartytownWorker {
  // ...
}
interface TestWorker extends PartytownWorker {
  messages: any[]
}
// but you can't use a TestWorker as a PartytownWorker
// because this: this is ultimately invariant
declar var tw: TestWorker
var ptw: PartytownWorker = tw
```

@rbuckton suggested making MessageEventTarget generic so that Worker and the other extenders can specify a fixed `this` type:

```ts
interface MessageEventTarget<T> {
  onmessage: ((this: T, ...
}
interface Worker extends MessageEventTarget<Worker> {
}
```

This is a delimited form of `this: this` that stops with Worker (and GlobalEventHandler and WindowEventHandler). I'll see if I can figure out how to emit this.

sandersn:
https://github.com/microsoft/TypeScript/pull/60987 is the PR I'm using to test this on TS.

sandersn:
Because it was taking me so long to figure out changes here, I manually made changes to https://github.com/microsoft/TypeScript/pull/60987.

However, I think I've got a commit that generates the same fix as there. Still feels a bit hacky, but maybe less than before?

rbuckton (approved):

saschanaz:
overrideImplements would be very fragile when spec change happens. But that can be a followup.

PR #1884, opened 2025-01-23
Automate overrideImplements

### LINKED ISSUE #1883: Replace overrideImplements with array replacement
Any extra mixin inclusions would be ignored by overrideImplements. Array item replacement should be used instead.

saschanaz:
Ultimately we should add type parameter to all mixins with event handlers.

Fixes #1883


saschanaz:
LGTM

PR #1893, opened 2025-01-31
fix: make `TextEncoder.prototype.encode` return `Uint8Array<ArrayBuffer>`

dsherret:
It's surprising that this repo has zero references to `Uint8Array<ArrayBuffer>` given how disruptive the change is for typed arrays.

For https://github.com/microsoft/TypeScript/issues/60846

### DIFF
--- a/baselines/ts5.5/audioworklet.generated.d.ts
+++ b/baselines/ts5.5/audioworklet.generated.d.ts
@@ -934,7 +934,7 @@ interface TextEncoder extends TextEncoderCommon {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/TextEncoder/encode)
      */
-    encode(input?: string): Uint8Array;
+    encode(input?: string): Uint8Array<ArrayBuffer>;
     /**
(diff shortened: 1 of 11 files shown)


baselines/ts5.5/audioworklet.generated.d.ts
```diff
@@ -934,7 +934,7 @@ interface TextEncoder extends TextEncoderCommon {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/TextEncoder/encode)
      */
-    encode(input?: string): Uint8Array;
+    encode(input?: string): Uint8Array<ArrayBuffer>;
```
dsherret, line 937: This is wrong for TS 5.5. The minimum should be TS 5.7. How do I deal with this?

saschanaz:
The vast majority of web API parameters should use `<ArrayBuffer>` (unless marked with `[AllowShared]`), except there's no corresponding marker for return values. I believe ArrayBuffer should be the default everywhere for web APIs.

And yeah, it seems we need another set of versioned libs. Not sure we want to keep adding more and more versions though, @sandersn?

petamoriken:
superseded by #1944

dsherret:
Closing in favour of https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1944

PR #1902, opened 2025-02-03
Make package name scoped to @typescript

zombie:
Apparently this package name was name squatted on npm by a malicious actor: https://github.com/advisories/GHSA-5h3m-q8pm-rw64.

The issue is that anyone importing this package gets a "**critical** severity vulnerability" because of the package name.  Making the package name scoped to @microsoft avoids any confusion today, and guards against any [potential future npm substitution attacks](https://github.blog/security/supply-chain-security/avoiding-npm-substitution-attacks/).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "typescript-dom-lib-generator",
+  "name": "@microsoft/typescript-dom-lib-generator",
   "version": "0.0.1",
   "private": true,
   "description": "Provides TypeScript types for the latest web APIs.",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@microsoft/typescript-dom-lib-generator",
+  "name": "@typescript/dom-lib-generator",
   "version": "0.0.1",
   "private": true,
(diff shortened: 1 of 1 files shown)


saschanaz:
Makes sense to me but the merge should happen from @sandersn.

Note that users in general should use `@types/web`, but Mozilla is giving it's own browser IDL files to the generator and [thus is importing the whole package](https://searchfox.org/mozilla-central/rev/b1662b447f306e6554647914090d4b73ac8e1664/tools/ts/package.json#10).

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "typescript-dom-lib-generator",
+  "name": "@microsoft/typescript-dom-lib-generator",
```
sandersn, line 2:
```suggestion
  "name": "@typescript/dom-lib-generator",
```

since typescript has its own scope on npm.

sandersn (changes requested):
Thanks for bringing this to our attention.

zombie:
Oops, pushed the wrong button.

I accepted the proposed change, and renamed the PR.  This is now ready to be merged.

sandersn (approved):

PR #1907, opened 2025-02-07
Sort namespaces before emitting

zombie:
This makes the output more stable and diff-friendly.


saschanaz:
Should rebuild and accept the new baseline 👀

zombie:

saschanaz:
LGTM

PR #1910, opened 2025-02-09
Emit only instance members from global polluters

zombie:
I believe this is correct, static members from global polluters shouldn't show up as globals.

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -803,7 +803,8 @@ export function emitWebIdl(
     if (
       p.name === "name" &&
       i.name === "Window" &&
-      emitScope === EmitScope.All
+      emitScope === EmitScope.InstanceOnly &&
+      prefix === "declare var "
     ) {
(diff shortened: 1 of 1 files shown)


zombie:
I tried looking in `inputfiles/`, but I don't really understand where this difference from baselines is coming from.
```
    -/** @deprecated */
    -declare const name: void;
    +declare var name: string;
```

saschanaz (changes requested):
It's special cased here:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/be2cad95f54f1b4b0e97e47ecd8061e2554d26e4/src/build/emitter.ts#L803-L810

Changing that should fix the failure.

(But Gecko has static member in global interfaces? 🤔🤔🤔 / Edit: Oh, isInstance, yeah.)

zombie:
> It's special cased here:
Thanks, I had a feeling I was looking in the wrong places.

saschanaz:
LGTM, thanks!

PR #1914, opened 2025-02-20
Correct optionality of `ReadableStreamReadDoneResult.value`

Renegade334:
The IDL specifies for `ReadableStreamReadResult` that the `value` property is always present. If the stream is cancelled, then this property won't be absent, but will be present with the value `undefined`. (This differs from _eg._ the iterator protocol, where a result object's `value` property may indeed be absent if the iterator is closed.)

The difference is only really manifest when `exactOptionalPropertyTypes` is enabled, but where it is, `ReadableStreamReadDoneResult` is currently divergent from the web standard.

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -79,7 +79,7 @@ interface ReadableStreamIteratorOptions {
 
 interface ReadableStreamReadDoneResult<T> {
     done: true;
-    value?: T;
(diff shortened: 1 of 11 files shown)


saschanaz:
Technically it's optional per the IDL , it's [the behavior](https://streams.spec.whatwg.org/#default-reader-read) that always assigns value there. So LGTM.

PR #1918, opened 2025-02-26
Restore StartViewTransitionOptions

### LINKED ISSUE #1892: Missing Types: ViewTransition
## Missing `types`
`startViewTransition` now support an object as parameter `{ update: () => void | Promise<void>; types: string[] }`
https://developer.chrome.com/docs/web-platform/view-transitions/same-document#view-transition-types
```ts
document.startViewTransition({
  update: updateDOM,
  types: ['spa-navigation'],
});
```
Also the `ViewTransition` object now includes a `types` key of type `ViewTransitionTypeSet` which implement the [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#set-like_browser_apis) object.

## Missing `viewTransitionClass` in `CSSStyleDeclaration`
The `view-transition-class` declaration should be included in the `CSSStyleDeclaration` 
https://developer.chrome.com/docs/web-platform/view-transitions/same-document#view-transition-class
```ts
element.style.viewTransitionClass;
```

This feature is supported by Chome (125) and Safari (18).

Note: Both features are not available on MDN yet as for today (27 January 2024)

saschanaz:
The multiple implementation requirement is achieved in 2024-09.

Closes #1892

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2090,6 +2090,11 @@ interface SpeechSynthesisEventInit extends EventInit {
     utterance: SpeechSynthesisUtterance;
 }
 
+interface StartViewTransitionOptions {
+    types?: string[] | null;
+    update?: ViewTransitionUpdateCallback | null;
+}
+
 interface StaticRangeInit {
     endContainer: Node;
     endOffset: number;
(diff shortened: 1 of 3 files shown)


saschanaz:
LGTM

PR #1931, opened 2025-03-11
Update overridingTypes.jsonc to add `PointerEvent` to `contextmenu` and `auxclick`

titoBouzout:
It has been noted in SolidJS discord that the interface for `contextmenu` and `auxclick` events is `PointerEvent` not `MouseEvent`.  MDN supports 
- https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event 
- https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event

The file being updated here has other 10 sibling entries also for the same event `PointerEvent`, so I am adding 2 more.

Now, I'm not completely sure why this is still _outdated?_/needs manual overriding, may be worth to look at that. 

Thanks!

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -66,12 +66,20 @@
                             "type": "AnimationEvent"
                         },
                         {
+                            "name": "auxclick",
+                            "type": "PointerEvent"
+                        },
+                        {
                             "name": "cut",
                             "type": "ClipboardEvent"
                         },
                         {
                             "name": "copy",
                             "type": "ClipboardEvent"
+                        },
+                        {
+                            "name": "contextmenu",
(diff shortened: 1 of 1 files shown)


saschanaz:
Probably because of GlobalEventHandlers being a mixin, there's some work needed to fully automate it for mixins. LGTM

saschanaz:
Oh wait this did not update the baseline.

titoBouzout:
> Oh wait this did not update the baseline.

Sorry I didn't know, want me to update baseline? I can do it

saschanaz:
Fixed it in #1932, but this is how: `npm run build && npm run baseline-accept`

titoBouzout:
@saschanaz Thanks!

PR #1935, opened 2025-03-12
fix: Update MDN references and change types to Set<string> in generated definitions

### LINKED ISSUE #1928: Missing ViewTransition types field
## ViewTransition types
The ViewTransition object now includes a types key of type `ViewTransitionTypeSet` which implement the [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#set-like_browser_apis) object.

```ts
/** Not sure if this is needed, but this is the name of the class when logs in the console */
type ViewTransitionTypeSet = Set<string>;

interface ViewTransition {
  // Existing
  readonly finished: Promise<undefined>;
  readonly ready: Promise<undefined>;
  readonly updateCallbackDone: Promise<undefined>;
  skipTransition(): void;
  // Missing
  types: ViewTransitionTypeSet ;
}
```

Feature supported by Chome (125) and Safari (18).

Bashamega:
fixes #1928

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -9931,7 +9931,7 @@ interface GlobalEventHandlers {
      * Fires immediately after the browser loads the object.
      * @param ev The event.
      *
-     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/load_event)
+     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/SVGElement/load_event)
      */
     onload: ((this: GlobalEventHandlers, ev: Event) => any) | null;
     /**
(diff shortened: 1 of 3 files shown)


Bashamega:
I ran the baseline command. I don't understand why it is not working.

baselines/dom.generated.d.ts
```diff
@@ -24469,7 +24467,7 @@ interface ViewTransition {
     readonly finished: Promise<void>;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ViewTransition/ready) */
     readonly ready: Promise<void>;
-    types: ViewTransitionTypeSet;
+    types: Set<string>;
```
saschanaz, line 24470: ViewTransitionTypeSet is Set<string>, why do we need a change here?

Bashamega: If this is the case. What is the issue?

saschanaz: There's no issue, I just closed it.

Bashamega: Okay

PR #1940, opened 2025-03-15
API Descriptions Update

### LINKED ISSUE #1937: Proposal: Automate API Descriptions Update via GitHub Action
Hello @saschanaz,  

The `inputfiles/mdn/apiDescriptions.json` file is outdated and has not been maintained for the past five years. To ensure that we have up-to-date API descriptions, I propose implementing a weekly GitHub Action that:  

- Scrapes the relevant MDN documentation  
- Generates an updated JSON file  
- Automatically opens a pull request with the latest data  

I have reviewed MDN’s [`[robots.txt](https://developer.mozilla.org/robots.txt)`](https://developer.mozilla.org/robots.txt) file, and it explicitly allows robots to scrape `en-US/docs/Web/API`, which covers the pages we need.  

I would be happy to take on this task and contribute the necessary implementation.  

**Reference:** #1909  

Looking forward to your thoughts!

Bashamega:
fixes #1937


saschanaz:
We still depend on that, so we can't just remove it until we get a new mdn/content dependency?

Bashamega:
> We still depend on that, so we can't just remove it until we get a new mdn/content dependency?

I know

src/build.ts
```diff
@@ -85,6 +86,101 @@ async function emitFlavor(
...
+  }
+
+  const octokit = new Octokit({
+    auth: authToken,
+    request: { headers: { "Accept": "application/vnd.github.v3+json" } },
+  });
+
+  const OWNER = "mdn";
+  const REPO = "content";
+  const BASE_PATH = "files/en-us/web/api";
```
saschanaz, line 107: I think maybe we should add a submodule so that we can make the build more reproducible locally

Bashamega: Sure, I can do that
I will get started on that

Bashamega:
Hello @saschanaz,
I have created a script to generate the JSON, ran the baseline command, and fixed some linting problems, but I don't know how to solve the error you see in the logs. It is just one error.

saschanaz:
Maybe `npm i`?

Bashamega:
> Maybe `npm i`?

Hello:)
How will `npm i` help?
I ran it and it didn't do anything

saschanaz:
I mean `npm i && npm run build && npm run baseline-accept`? Because it seem like it's caused by mismatch of installed dependency version.

Bashamega:
> I mean `npm i && npm run build && npm run baseline-accept`? Because it seems like it's caused by mismatch of dependency version. `npm i` fixes that.
oh okay

baselines/audioworklet.generated.d.ts
```diff
@@ -289,7 +297,11 @@ declare var ByteLengthQueuingStrategy: {
     new(init: QueuingStrategyInit): ByteLengthQueuingStrategy;
 };
 
-/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/CompressionStream) */
+/**
+ * Of the {{domxref('Compression Streams API','','',' ')}} is an API for compressing a stream of data.
```
saschanaz, line 301: This looks weird. Now that we are at it, maybe we should remove:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/cede2ca8e8dbf8344de3f5fc341b01e7ff6c9458/src/build.ts#L94-L113

Because I don't think it makes sense to deal with all kinds of human language without accidentally breaking things.

package.json
```diff
@@ -13,6 +13,7 @@
   ],
   "scripts": {
     "build": "tsc && node ./lib/build.js",
+    "generate": "tsc && node ./lib/generateDescriptions.js",
```
saschanaz, line 16: Can this be part of `build` instead (ideally called by lib/build.js?)

baselines/audioworklet.generated.d.ts
Bashamega: I think it would be easier if we converted the Markdown to HTML. I can utilize Cheerio like the mdn and get the summary like the mdn. If you think this is a good way to do it, I can implement it.

saschanaz: Not sure why that's needed here, can you explain more?

saschanaz: Oh you mean the domxref thing. Hmm.

Bashamega: Because the reason that I made all these regexes is to avoid the weird template code. But if I convert the markdown to HTML first, I think it would be easier to extract the data without the weird templates

package.json
Bashamega: Sure, I can do that :)

baselines/audioworklet.generated.d.ts
Bashamega: Oh

saschanaz: If using Cheerio is easy to remove domxref thing then I guess it's worth trying!

Bashamega: > If using Cheerio is easy to remove domxref thing then I guess it's worth trying!

Okay

Bashamega:
We need to figure out a way to compile the markdown without the weird templates.

baselines/audioworklet.generated.d.ts
```diff
@@ -999,7 +1059,7 @@ declare var TransformStreamDefaultController: {
 };
 
 /**
- * The URL interface represents an object providing static methods used for creating object URLs.
+ * The **URL** interface is used to parse, construct, normalize, and encode . It works by providing properties which allow you to easily read and modify the components of a URL.
```
saschanaz, line 1062: Hmm, encode what? It seems links are gone.

saschanaz (commented):
Wow, getting a lot better!

Bashamega:
> Wow, getting a lot better!

Thank you

baselines/audioworklet.generated.d.ts
Bashamega: I will try to fix it

Bashamega:
Hello @saschanaz 
I have tried to run `npm i && npm run build && npm run baseline-accept` but nothing got updated.
How can i fix the failing test?

Bashamega:
Hello @saschanaz 
I have fixed the builds.
Please review :)

Bashamega:
Hello @saschanaz,

I’ve completed the necessary changes in this PR, and I believe it’s ready for review and merging. Could you take a look when you have a moment?

Thanks!

baselines/audioworklet.generated.d.ts
```diff
@@ -339,7 +355,7 @@ declare var CustomEvent: {
 };
 
 /**
- * An abnormal event (called an exception) which occurs as a result of calling a method or accessing a property of a web API.
+ * The **DOMException** interface represents an abnormal event (called an **exception**) that occurs as a result of calling a method or accessing a property of a web API. This is how error conditions are described in web APIs.
```
saschanaz, line 358: This what? 🤔 Should we try to get only the first sentence?

Bashamega: Sure, I can do that. Do you have any other changes?

src/generateDescriptions.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+function cleanText(text: string): string {
+  return text
+    .replace(
+      /\{\{\s*(Glossary|HTMLElement|SVGAttr|SVGElement|cssxref|jsxref|HTTPHeader)\s*\(\s*["']((?:\\.|[^"\\])*?)["'].*?\)\s*\}\}/gi,
+      "$2",
+    ) // Extract first argument from multiple templates, handling escaped quotes & spaces
```
saschanaz, line 13: I guess the plan to use the parser didn't work well? 👀 (For now this is fine as-is in your patch)

src/generateDescriptions.ts
```diff
@@ -0,0 +1,114 @@
...
+  );
+  const outputDir = path.resolve(__dirname, "../inputfiles/mdn/");
+  const outputFile = path.join(outputDir, "apiDescriptions.json");
+
+  try {
+    await fs.mkdir(outputDir, { recursive: true });
+    const folders = await getFolders(basePath);
+    if (folders.length) {
+      const data = await getIndexMdContents(folders);
+      await fs.writeFile(outputFile, JSON.stringify(data, null, 2), "utf-8");
```
saschanaz, line 108: We don't have to store a file, emitDom should instead receive an object.

src/generateDescriptions.ts
Bashamega: Yes, it didn't work out

baselines/audioworklet.generated.d.ts
Bashamega: Done

src/generateDescriptions.ts
Bashamega: This would cause the test workflow to break because, by default, it won't clone the submodule. As a result, it will return an empty array, and during the build process, it will report that the baselines are not the same. You can modify the workflow to clone the submodule, but unfortunately, I can't do that since I don't have permission to update the workflows.

Bashamega: > > I can't do that since I don't have permission to update the workflows.
> 
> The CI won't run but we can still verify it locally and eventually merge it, right?

Yes

.gitignore
```diff
@@ -285,4 +285,5 @@ inputfiles/browser.webidl.json
 !.vscode/launch.template.json
 yarn.lock
 TypeScript
-deploy/generated
\ No newline at end of file
+deploy/generated
+.early.coverage
```
saschanaz, line 289: What is this file?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "mdn-content"]
+	path = mdn-content
```
saschanaz, line 2: Can this go inside `inputfiles/`?

src/generateDescriptions.ts
saschanaz: Okay. I think given we have this function here we can probably remove removeVerboseIntroductions() from src/build.

baselines/dom.generated.d.ts
```diff
@@ -6918,7 +7175,7 @@ declare var DOMMatrixReadOnly: {
 };
 
 /**
- * Provides the ability to parse XML or HTML source code from a string into a DOM Document.
+ * The **DOMParser** interface provides
```
saschanaz, line 7178: Err, looks like some files have newlines within a sentence: https://github.com/mdn/content/blob/a52689c74c6c89f45c54447bb148e54ed320db62/files/en-us/web/api/domparser/index.md?plain=1#L3

.gitignore
Bashamega: Sorry:)
It is something locally for me, I will remove it

.gitmodules
Bashamega: Sure

Bashamega:
**Hello @saschanaz,**  

I've made the following updates:  
- Removed an extra folder that was mistakenly added to `.gitignore`.  
- Updated the directory path from `/mdn-content` to `/inputfiles/mdn`.  
- Removed the cleaner function from the generator file since it's already handled in the build file.  
- Changed the splitting logic to split by sentence instead of by new line.    

Let me know if any further changes are needed!

Bashamega:
> Hmm. What happens if you modify workflow file here? Does it outright reject your commit? It shouldn't as it's still your own fork 🤔

I have tried it again, and it worked. I guess yesterday it didn't work because I had an issue with my laptop and had to use Gitpod.

.github/workflows/test_typescript.yml
```diff
@@ -14,6 +14,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true  # Ensures submodules are cloned
+
+      - name: Initialize and update submodules
+        run: |
+          git submodule update --init --recursive  # Explicitly initialize submodules
```
saschanaz, line 22: This is test_typescript, you should also update ci.yml.

saschanaz: And also deploy.yml.

Bashamega:
Hello @saschanaz 
I have fixed the templates. Please review 😃

src/build.ts
```diff
@@ -394,5 +418,4 @@ async function emitDom() {
     }
   }
 }
-
```
saschanaz, line 397: Maybe restore this?

src/build.ts
```diff
@@ -195,7 +196,30 @@ async function emitDom() {
...
+      ) // Extract first argument from domxref, handling spaces
+      .replace(
+        /\{\{\s*(?:event|jsxref|cssref|specname)\s*\|\s*([^}]+)\s*\}\}/gi,
+        "$1",
+      ) // Handle event, jsxref, cssref, etc.
+      .replace(/\{\{\s*([^}]+)\s*\}\}/g, (_, match) => `[MISSING: ${match}]`) // Catch any remaining unhandled templates
+      .replace(/\\(["'])/g, "$1") // Remove backslashes from escaped quotes
+      .replace(/&lt;/g, "<")
+      .replace(/&gt;/g, ">") // Decode HTML entities
+      .replace(/`([^`]+)`/g, "$1") // Keep inline code readable
```
saschanaz, line 216: Why do we remove backticks? 👀

baselines/dom.generated.d.ts
```diff
@@ -3919,7 +4018,11 @@ declare var CSSMathClamp: {
     new(lower: CSSNumberish, value: CSSNumberish, upper: CSSNumberish): CSSMathClamp;
 };
 
-/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/CSSMathInvert) */
+/**
+ * The **CSSMathInvert** interface of the CSS Typed Object Model API represents a CSS calc used as calc(1 / ). It inherits properties and methods from its parent CSSNumericValue.
```
saschanaz, line 4022: This lacks `<value>` in `calc(1 / )`, that's tricky... maybe we should have a real markdown parser for that. I think that can be done in a followup and not necessarily here.

baselines/dom.generated.d.ts
```diff
@@ -8348,7 +8631,7 @@ interface ElementEventMap {
 }
 
 /**
- * Element is the most general base class from which all objects in a Document inherit. It only has methods and properties common to all kinds of elements. More specific classes inherit from Element.
+ * **Element** is the most general base class from which all element objects (i.e.
```
saschanaz, line 8634: Oh no. This is tricky too without having an actual language processor library. Something for a followup.

baselines/dom.generated.d.ts
```diff
@@ -9511,7 +9859,7 @@ declare var GamepadButton: {
 };
 
 /**
- * This Gamepad API interface contains references to gamepads connected to the system, which is what the gamepad events Window.gamepadconnected and Window.gamepaddisconnected are fired in response to.
+ * This Gamepad API interface contains references to gamepads connected to the system, which is what the gamepad events {{domxref("Window.gamepadconnected_event", "gamepadconnected")}} and {{domxref("Window.gamepaddisconnected_event", "gamepaddisconnected")}} are fired in response to.
```
saschanaz, line 9862: Still some domxref? 👀

Bashamega:
I will try to fix more

baselines/dom.generated.d.ts
Bashamega: I have fixed it. I forgot to remove templates when I have migrated it to the build file

baselines/dom.generated.d.ts
Bashamega: I have fixed it

src/build.ts
Bashamega: Done

.github/workflows/ci.yml
```diff
@@ -7,6 +7,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true  # Ensures submodules are cloned
+
+      - name: Initialize and update submodules
+        run: |
+          git submodule update --init --recursive  # Explicitly initialize submodules
```
saschanaz, line 15: This shouldn't be needed anymore, right?

baselines/dom.generated.d.ts
saschanaz: We still lack backticks here, and in some markdown implementations lacking backticks may mean `<value>` won't be visible at all. (e.g. if it uses GFM)

src/build.ts
saschanaz: Hmm? It still removes backticks 👀

.github/workflows/ci.yml
Bashamega: I have forgot to remove them. I have removed them

Bashamega:
Hello @saschanaz 
I forgot to remove the backticks. They are now removed

baselines/ts5.5/dom.generated.d.ts
```diff
@@ -8624,7 +8624,7 @@ interface ElementEventMap {
 }
 
 /**
- * **Element** is the most general base class from which all element objects (i.e.
+ * **`Element`** is the most general base class from which all element objects (i.e.
```
saschanaz, line 8627: (Something for a followup)

baselines/webworker.generated.d.ts
```diff
@@ -2623,11 +2774,7 @@ interface EXT_color_buffer_half_float {
 interface EXT_float_blend {
 }
 
-/**
- * The EXT_frag_depth extension is part of the WebGL API and enables to set a depth value of a fragment from within the fragment shader.
- *
- * [MDN Reference](https://developer.mozilla.org/docs/Web/API/EXT_frag_depth)
- */
+/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/EXT_frag_depth) */
```
saschanaz, line 2777: Hmm? How did we lose it 🤔

src/build.ts
saschanaz: (Still a removed newline)

baselines/serviceworker.generated.d.ts
```diff
@@ -8477,7 +8744,12 @@ declare namespace WebAssembly {
     function validate(bytes: BufferSource): boolean;
 }
 
-/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/console) */
+/** The **`console`** object provides access to the debugging console (e.g., the Web console in Firefox). */
+/**
+ * The **`console`** object provides access to the debugging console (e.g., the Web console in Firefox).
+ *
```
saschanaz, line 8750: This gets double comments? 🤔

saschanaz (commented):
Skimmed again and mostly looks nice, except a few now-missing comments and duplicated comments. This is HUGE, thank you for the effort here!

baselines/ts5.5/dom.generated.d.ts
Bashamega: This is because we do not check for abbreviations

Bashamega: I have fixed it

baselines/webworker.generated.d.ts
Bashamega: I have fixed it

src/generateDescriptions.ts
```diff
@@ -0,0 +1,91 @@
...
+    .split("\n")
+    .map((line) => line.trim())
+    .filter(
+      (line) =>
+        line &&
+        !line.startsWith("#") &&
+        !line.startsWith(">") &&
+        !line.startsWith("{{"),
+    )
+    .join(" ");
```
saschanaz, line 23: transformVerbosity should be merged here. (And it's not about "verbosity" anymore)

baselines/dom.generated.d.ts
```diff
@@ -8348,7 +8675,7 @@ interface ElementEventMap {
 }
 
 /**
- * Element is the most general base class from which all objects in a Document inherit. It only has methods and properties common to all kinds of elements. More specific classes inherit from Element.
+ * **`Element`** is the most general base class from which all element objects (i.e.
```
saschanaz, line 8678: (This is still a problem)

src/build.ts
```diff
@@ -160,7 +139,7 @@ async function emitDom() {
     for (const [key, value] of Object.entries(descriptions)) {
       const target = idl.interfaces!.interface[key] || namespaces[key];
       if (target && !value.startsWith("REDIRECT")) {
```
saschanaz, line 141: REDIRECT seems like an old thing that can be removed now.

src/generateDescriptions.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  // Extract the first sentence (ending in . ! or ?)
+  const sentenceMatch = normalizedText.match(
+    /(.*?(?<!\.\s(?:i\.e|e\.g))([.!?]))(?:\s|$)/i,
```
saschanaz, line 27: This doesn't really solve the "e.g." case, as replacing this with `/(.*?[.!?])(?=\s|$)/` doesn't change anything. I'm okay with deferring that issue for now, but in that case we should revert this regex, to prevent giving false sense that this solves the case when it does not.

src/generateDescriptions.ts
```diff
@@ -0,0 +1,91 @@
+import { promises as fs } from "fs";
+import * as path from "path";
+import { fileURLToPath } from "url";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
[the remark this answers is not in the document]
Bashamega: Done

src/generateDescriptions.ts
```diff
@@ -0,0 +1,91 @@
+import { promises as fs } from "fs";
```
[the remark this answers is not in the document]
Bashamega: Done

src/generateDescriptions.ts
Bashamega: I don't understand

baselines/dom.generated.d.ts
Bashamega: Okay

package.json
```diff
@@ -58,10 +58,12 @@
     "globals": "^15.9.0",
     "jsonc-parser": "^3.2.1",
     "node-fetch": "^3.3.2",
+    "path": "^0.12.7",
```
saschanaz, line 61: This and `url` below too.

src/build/mdn-comments.ts
```diff
@@ -1,9 +1,10 @@
...
+import fs from "fs/promises";
 import * as path from "path";
-import { fileURLToPath } from "url";
 
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
+const basePath = new URL(
+  "../../inputfiles/mdn/files/en-us/web/api/",
+  import.meta.url,
+).pathname;
```
saschanaz, line 7: Using pathname still requires using `path` elsewhere, but the file system API can understand URL directly.

src/build/mdn-comments.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  return normalizedText.split(" ")[0] || ""; // Fallback: first word if no sentence found
+}
+
+async function getFolders(dirPath: string): Promise<string[]> {
+  try {
+    const entries = await fs.readdir(dirPath, { withFileTypes: true });
+    return entries
+      .filter((entry) => entry.isDirectory())
+      .map((entry) => path.join(dirPath, entry.name));
```
saschanaz, line 40: Here for example, this could be `new URL(entry.name, dirPath)` and return `Promise<URL[]>`.

src/build.ts
```diff
@@ -178,24 +157,32 @@ async function emitDom() {
...
-
-      const reg = new RegExp(template.replace(/\$\{name\}/g, name) + "\\s*");
-      const product = description.replace(reg, replace);
-      if (product !== description) {
-        return product.charAt(0).toUpperCase() + product.slice(1);
-      }
-    }
-
-    return description;
+  function transformVerbosity(description: string): string {
```
saschanaz, line 166: We should move all MDN comment generation code to `mdn-comments.ts`. IMO this can be merged right into `extractSummary()`.

Bashamega:
I have fixed all the issues @saschanaz

baselines/ts5.5/dom.generated.d.ts
Bashamega: Hello @saschanaz
[This pr](https://github.com/mdn/content/pull/38905) should fix this problem. Once it is merged I will regenerate the baselines

.github/workflows/update-core-deps.yml
```diff
@@ -12,6 +12,17 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true  # Ensure submodules are checked out
+
+      - name: Update submodules
+        run: |
+          git submodule update --init
+          cd inputfiles/mdn  # Replace with actual submodule path
+          git checkout main  # Change if using a different branch
```
saschanaz, line 22: We are certainly using a different branch (well, no branch actually), so no "if". And `git switch` is preferred for branches nowadays.

.github/workflows/update-core-deps.yml
```diff
@@ -1,4 +1,4 @@
-name: Update core dependencies
+name: Update core dependencies and submodules
```
saschanaz, line 1: The MDN submodule is also a core dependency so we can skip it. Same for the changes below.

src/build/mdn-comments.ts
```diff
@@ -32,34 +50,37 @@ function extractSummary(markdown: string): string {
   return normalizedText.split(" ")[0] || ""; // Fallback: first word if no sentence found
 }
 
-async function getFolders(dirPath: string): Promise<string[]> {
+async function getFolders(dirPath: URL): Promise<URL[]> {
   try {
-    const entries = await fs.readdir(dirPath, { withFileTypes: true });
+    const entries = await fs.readdir(fileURLToPath(dirPath), {
```
saschanaz, line 55: We don't need `fileURLToPath`, you can pass `dirPath` directly.

src/build/mdn-comments.ts
```diff
@@ -32,34 +50,37 @@ function extractSummary(markdown: string): string {
...
 ): Promise<{ [key: string]: string }> {
   const results: { [key: string]: string } = {};
 
   for (const folder of folders) {
-    const indexPath = path.join(folder, "index.md");
+    const indexPath = new URL("index.md", folder);
 
     try {
-      const content = await fs.readFile(indexPath, "utf-8");
+      const content = await fs.readFile(fileURLToPath(indexPath), "utf-8");
```
saschanaz, line 76: Same here, `fileURLToPath` can be skipped.

src/build/mdn-comments.ts
```diff
@@ -32,34 +50,37 @@ function extractSummary(markdown: string): string {
...
 
       // Improved title extraction
       const titleMatch = content.match(/title:\s*["']?([^"'\n]+)["']?/);
+      const filename = fileURLToPath(folder).split("/").pop();
```
saschanaz, line 80: You can do `path.basename(folder.toString())`

Bashamega:
Done @saschanaz

.github/workflows/update-core-deps.yml
```diff
@@ -34,13 +41,13 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.TS_GITHUB_BOT_AUTH }}
       - id: git-diff
-        run: git diff --quiet HEAD baselines
+        run: git diff --quiet HEAD baselines submodules
```
saschanaz, line 44: There's no path `submodules`, and just comparing `baselines` is enough, nothing more is needed here.

.github/workflows/update-core-deps.yml
```diff
@@ -34,13 +41,13 @@ jobs:
...
         continue-on-error: true
       - uses: peter-evans/create-pull-request@v7
         if: ${{ steps.git-diff.outcome == 'failure' }}
         with:
-          commit-message: "🤖 Update core dependencies"
-          title: "Update core dependencies"
+          commit-message: "🤖 Update core dependencies and submodules"
+          title: "Update core dependencies and submodules"
```
saschanaz, line 50: Please also revert these two.

src/build/mdn-comments.ts
```diff
@@ -0,0 +1,106 @@
...
+  // Extract the first sentence (ending in . ! or ?)
+  const sentenceMatch = normalizedText.match(/(.*?[.!?])(?=\s|$)/);
+  if (sentenceMatch) {
+    return sentenceMatch[0]; // Return the first full sentence
+  }
+
+  return normalizedText.split(" ")[0] || ""; // Fallback: first word if no sentence found
+}
+
+async function getFolders(dirPath: URL): Promise<URL[]> {
```
saschanaz, line 53: getDirectories, to match other names (it's readdir and dirPath, not readfolder and folderPath)

```suggestion
async function getDirectories(dirPath: URL): Promise<URL[]> {
```

saschanaz (commented):
The only blocking change is the workflow thing, otherwise this looks mostly fine. 🎉

Bashamega:
Done @saschanaz

README.md
```diff
@@ -88,7 +88,7 @@ If you are familiar with Web IDL, you may also want to check whether the upstrea
 - `overridingTypes.jsonc`: types that are defined in the spec but have TypeScript-friendly modifications in the json files.
 - `removedTypes.jsonc`: types that are defined in the spec but should be removed.
 - `comments.json`: comment strings to be embedded in the generated .js files.
-- `deprecatedMessage.json`: the reason why one type is deprecated. The reason why it is a separate file rather than merge in comment.json is mdn/apiDescriptions.json would also possibly be deprecated.
+
```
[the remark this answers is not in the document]
saschanaz: Removing that is the last request from me

Bashamega:
Done @saschanaz

Bashamega:
Hello @saschanaz,

My PR to the MDN repo has been merged, and I’ve synced the repo to resolve the abbreviation issue.

saschanaz:
LGTM, thanks! (This may also require Microsoft review, but let me try given there's no maintainers tag...)

Bashamega:
Hello @sandersn
Can you please check this PR and merge it?
Thank you
CC - @saschanaz

.github/workflows/update-core-deps.yml
```diff
@@ -12,6 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true  # Ensure submodules are checked out
+
+      - name: Update submodules
+        working-directory: inputfiles/mdn  # Replace with actual submodule path
```
sandersn, line 19: these comments look left over from an example, since `inputfiles/mdn` is the actual submodule path. Can you remove them?

sandersn (changes requested):
One tiny cleanup request. 

I'm not a fan of submodules but if @saschanaz is OK with the overhead of working with them it's fine with me.

saschanaz:
I considered adding it as a npm package but that causes a terribly deep dependency tree, so I think submodule is the best choice for now.

Bashamega:
Thank you @sandersn for your review. I have updated the workflow

Bashamega:
> One tiny cleanup request. 
> 
> I'm not a fan of submodules but if @saschanaz is OK with the overhead of working with them it's fine with me.

Hello @sandersn, I understand you worry that it won't be updated regularly because the update dependencies workflow will auto update it. So I don't think you should worry about it 🙂

Bashamega:
Hello @jakebailey, and @sandersn 
I have applied all the requested changes. Can we merge this?

jakebailey:
I guess I'm of the opinion that adding a submodule is a bad idea 😄 It's the entire MDN repo! Is there no other way?

> I understand you worry that it won't be updated regularly because the update dependencies workflow will auto update it. So I don't think you should worry about it 

The updating script now just keeps pulling main from the submodule, which I guess works but sounds like it's going to be a lot of churn, since before the dep update may not actually do anything, but of course MDN is going to be updated frequently...

saschanaz:
1. Manually download the tree, as initially this PR was doing. You'll have to download the same size for every update (even a small update would trigger a full download), and also have to manually maintain the code to make sure the local download matches the in-tree reference, caching if one has to switch to different branches, etc etc.
2. `git subtree --squash`. This would allow downloading only some part of the repo. But the same issue exists: even a small update would trigger a full download.
3. `git subtree` without `--squash`. I don't think we want to merge all the commits to this repository. Plain No.

jakebailey:
I'm a big fan of subtrees but they're definitely the wrong tool for this particular job (pulling in someone else's code but not modifying it or pushing back) 😄

Maybe a submodule is fine. I haven't looked deeply to see what all is done here, so this is mainly a gut reaction. Will harder look soon.

Bashamega:
Hello @jakebailey,
Technically, there is another option: scraping the MDN docs programmatically. However, this would require more code and may not be as efficient. I believe using a submodule is the best approach, but if you have another suggestion, I'm open to ideas!

jakebailey:
Read the PR, and yeah, I think it's untenable to do anything other than something that produces a local clone. I was thinking "what about asking github for the raw files", but this quite literally loops over every single markdown file in MDN (more or less), so that won't work.

package-lock.json
jakebailey, line 1: Why is this file changing? `package.json` hasn't changed, so I wouldn't expect a lockfile update here.

README.md
```diff
@@ -88,7 +88,6 @@ If you are familiar with Web IDL, you may also want to check whether the upstrea
 - `overridingTypes.jsonc`: types that are defined in the spec but have TypeScript-friendly modifications in the json files.
 - `removedTypes.jsonc`: types that are defined in the spec but should be removed.
 - `comments.json`: comment strings to be embedded in the generated .js files.
-- `deprecatedMessage.json`: the reason why one type is deprecated. The reason why it is a separate file rather than merge in comment.json is mdn/apiDescriptions.json would also possibly be deprecated.
```
jakebailey, line 91: This line is being removed, but the `deprecatedMessage.json` file still exists in the repo. Should it be removed?

Bashamega: This was a mistake

jakebailey: Also, given we're going to depend on mdn, I do wonder if we can extract this information from there too. It seems like mdn includes warnings for deprecations.

Bashamega: Yeah, but this PR is already huge, I think making it a separate PR would be better. I am planning on making more PRs like for more messages

jakebailey: Yes, just thinking aloud.

package-lock.json
```diff
@@ -348,9 +348,9 @@
       }
     },
     "node_modules/@mdn/browser-compat-data": {
-      "version": "6.0.4",
-      "resolved": "https://registry.npmjs.org/@mdn/browser-compat-data/-/browser-compat-data-6.0.4.tgz",
-      "integrity": "sha512-d9Em7f44DDP1BkJdqBvbjohC25efLgdGoNJvobxm5iMN0EGK677pC83SBjyPmU7OObc9s8Xu+Sv00dzKS+Xdng==",
+      "version": "6.0.5",
```
jakebailey, line 351: This is still modified even though nothing changed in the repo that would do this.

Consider doing something like `git checkout main -- package-lock.json` or something.

.github/workflows/update-core-deps.yml
```diff
@@ -12,6 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true  # Ensure submodules are checked out
+
+      - name: Update submodules
+        working-directory: inputfiles/mdn
+        run: |
+          git pull origin main
```
jakebailey, line 21: I would think a better way to do this would be `git submodule update --init --remote`? That way, the gitmodules config the source of truth for pulling new refs, without hardcoding any dirs or branch names.

Bashamega: done

src/build/mdn-comments.ts
```diff
@@ -0,0 +1,106 @@
...
+    } catch (error) {
+      console.warn(`Skipping ${indexPath}: ${error}`);
+    }
+  }
+
+  return results;
+}
+
+export async function generateDescription(): Promise<Record<string, string>> {
+  try {
```
jakebailey, line 96: In this function I would do a `stat` to verify that `basePath` exists. If it doesn't, we should print out a message that says something like:

> MDN submodule does not exist; try running `git submodule update --init`

Or something.

Honestly, I feel like printing errors here in this function is wrong; we should just throw and let the script break, since something is very off.

Bashamega: done

README.md
saschanaz: That info is already being imported with more readable form via `@mdn/browser-compat-data`.

jakebailey (approved):
Theoretically seems good to me; @saschanaz seem alright as a first go?

.github/workflows/update-core-deps.yml
```diff
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true  # Ensure submodules are checked out
```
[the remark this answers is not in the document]
saschanaz: But to update it doesn't make much sense to have `depth=1` because it would cause double full downloads?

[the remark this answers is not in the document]
saschanaz: Maybe we can just skip this line when we call `git submodule update` below anyway.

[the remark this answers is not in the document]
jakebailey: No, there's no inefficiency; git will correctly handle reusing data and not double download. Mainly I was just thinking about whether or not something like:

```
git clone --depth=1 ...
git pull
```

Works for new refs. Which, I think it does.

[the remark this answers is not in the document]
saschanaz: Hmm, interesting, I guess it's fine then. 👍🏻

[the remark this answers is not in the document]
Bashamega: Hello @jakebailey 
According to the documentation for this action, the default is set to one.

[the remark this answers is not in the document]
jakebailey: Yes, that's what I'm saying it does. I haven't personally attempted to update a submodule that was cloned with depth 1, but I will probably go try that quick.

Bashamega:
Hi @saschanaz, @jakebailey, @sandersn 👋

I've looked into ways to clone only specific parts of a repo without fetching the entire repository. Git doesn't support this natively, but I found a practical workaround using [SVN (Subversion)](https://subversion.apache.org/).

Although it's an older tool, it’s still supported by GitHub and allows downloading a single directory without pulling the whole repo. This might be a viable solution for our case—just wanted to share as a proposal for consideration.

jakebailey:
I don't think we need to do anything fancy. The submodule is fine.

Bashamega:
> I don't think we need to do anything fancy. The submodule is fine.

Okay, do you have any estimation of when this pr will be merged?

saschanaz:
SVN is *not* supported by GitHub https://github.blog/news-insights/product-news/sunsetting-subversion-support/ (I'm pretty surprised that it was supported at all)

saschanaz:
I see no blocker to not merge.

Bashamega:
> SVN is _not_ supported by GitHub https://github.blog/news-insights/product-news/sunsetting-subversion-support/ (I'm pretty surprised that it was supported at all)

Good to know, I didn't know

saschanaz:
I mean only if I could... 🤷🏻‍♀️

LGTM

Bashamega:
Hello @jakebailey 
Sorry for pinging you, but I think this PR is ready to be merged

jakebailey:
I'll merge it now, I had just replied late at night 😄

jakebailey (approved):

Bashamega:
> I'll merge it now, I had just replied late at night 😄

Thank you very much for merging this huge PR. You just made my day :)

Issue #1518, opened 2023-03-10
ResponseConstructor


mfulton26:
I would like to globally augment the `ResponseConstructor` (e.g. add the missing [`json(data, init)` static method](https://fetch.spec.whatwg.org/#dom-response-json), safely add my own using unique symbols, etc.).

However I do not see how to do that as `var Response` isn't defined with a named type.

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/48c4695b3f61534f1a6de8a598b35b81000491e3/baselines/dom.generated.d.ts#L12436-L12441

I suggest extracting the type to an interface, e.g. named `ResponseConstructor` to enable global augmentation of these types.

There are likely other types/constructors that could also be given named types too (e.g. `AbortController`, etc.).

saschanaz:
The way to extend static part is requested multiple times, I still think the right way to do this is to convert them to class declarations.

saschanaz:
This specific case will be solved once Gecko or WebKit adds the support in https://bugzilla.mozilla.org/show_bug.cgi?id=1758943 or https://github.com/WebKit/WebKit/pull/10593.

mfulton26:
> The way to extend static part is requested multiple times, I still think the right way to do this is to convert them to class declarations. (Because, well, they are classes.)

Until https://github.com/microsoft/TypeScript/issues/36813 is supported, declaring them as classes won't work for augmenting `Response` with properties keyed using unique symbols.

> This specific case will be solved once Gecko or WebKit adds the support in https://bugzilla.mozilla.org/show_bug.cgi?id=1758943 or [WebKit/WebKit#10593](https://github.com/WebKit/WebKit/pull/10593).

That's true but I want to define my own too using unique symbols. e.g. `Response[html]()`, `Response[notFound]()`, etc.

PR #1941, opened 2025-03-15
Implement `ResponseConstructor` to allow overrides from other runtimes

alii:
This PR converts `Response` to be declared as the interface `ResponseConstructor`

This allows for types from some runtimes to implement extra overloads for the Response class and static methods

```ts
interface ResponseConstructor {
  prototype: Response;
  new(body?: BodyInit | null, init?: ResponseInit): Response;
  redirect(url: string | URL, status?: number): Response;
  // ... more properties
}

var Response: ResponseConstructor
```

For example, Bun supports `Response.redirect(location, statusOrRequestInit)`. Bun's types cannot declare `var Response` because if a user needs lib.dom loaded then the declarations clash, and typechecking fails

`Response` is currently declared as `var Response`, which means we cannot use interface merging to append these extra properties

I'm not sure if this is the best approach for modifying/overriding the types. This was implemented by dumping the webidl JSON and then copying in the overrides for making a ResponseConstructor exist

There was also an existing issue (#1518) that asked for this

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -454,6 +454,262 @@
     },
     "interfaces": {
         "interface": {
+            "Response": {
+                "name": "Response",
+                "noInterfaceObject": true,
+
+                "properties": {
+                    "property": {
+                        "type": {
+                            "name": "type",
+                            "type": "ResponseType",
+                            "nullable": false,
+                            "static": false,
+                            "stringifier": false,
+                            "readonly": true,
+                            "exposed": "Window Worker",
+                            "secureContext": false,
+                            "mdnUrl": "https://developer.mozilla.org/docs/Web/API/Response/type"
+                        },
+                        "url": {
+                            "name": "url",
+                            "type": "USVString",
+                            "nullable": false,
+                            "static": false,
+                            "stringifier": false,
+                            "readonly": true,
+                            "exposed": "Window Worker",
+                            "secureContext": false,
+                            "mdnUrl": "https://developer.mozilla.org/docs/Web/API/Response/url"
+                        },
+                        "redirected": {
+                            "name": "redirected",
+                            "type": "boolean",
+                            "nullable": false,
+                            "static": false,
+                            "stringifier": false,
+                            "readonly": true,
+                            "exposed": "Window Worker",
+                            "secureContext": false,
+                            "mdnUrl": "https://developer.mozilla.org/docs/Web/API/Response/redirected"
+                        },
+                        "status": {
+                            "name": "status",
+                            "type": "unsigned short",
+                            "nullable": false,
+                            "static": false,
+                            "stringifier": false,
+                            "readonly": true,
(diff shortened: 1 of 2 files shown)


Bashamega (changes requested):
Please regenerate the baselines

petamoriken:
IMO, there should be more overall changes, not just special treatment of `Response`. The reason is that `ReadableStream.from` is not typed yet because it is not implemented in two browsers, but Deno is currently patching it. There may be other examples.

Bashamega (approved):

alii:

alii:
> IMO, there should be more overall changes, not just special treatment of `Response`. The reason is that `ReadableStream.from` is not typed yet because it is not implemented in two browsers, but Deno is currently patching it. There may be other examples.

Agree, but wasn't sure which ones should be added, and it's sort of a pandoras box to some extent. The "real fix" could actually be to make the generator emit `interface [Type]Constructor {}` for all class-like types by default, and then we'd be able to remove the overrides this PR adds anyway.

My suggestions would be:

- Include a `RequestConstructor` here, feels "correct" to have at-least both
- Make no changes (so this can merge😄), and instead reference this issue in the future if similar changes need to be made

petamoriken:
> Include a RequestConstructor here, feels "correct" to have at-least both

Looks good to me. In addition, I would like to do the same for `ReadableSteam`, but I think I will do it in a separate PR.

Another concern is that the current way of specifying `"Response": null` in removedTypes.jsonc does not reflect changes in browser compat data, so I feel it would be better to have a new config file such as exportConstructorTypes.jsonc (bikeshed).

baselines/audioworklet.generated.d.ts
```diff
@@ -862,6 +862,9 @@ interface ReadableStreamGenericReader {
     cancel(reason?: any): Promise<void>;
 }
 
+interface ResponseConstructor {
+}
```
petamoriken, line 866: The Audio Worklet type has been affected.

saschanaz:
I still think we should do #222, and this will block #222 (we won't be able to have both class declaration and var declaration together)

petamoriken:
I see, so https://github.com/microsoft/TypeScript/issues/2957 has to move on...

alii:
Got it, thanks for clarifying, @saschanaz.

Would love to get this working, so have asked about the feasibility of an ambient-only implementation of this behaviour in the TypeScript issue linked above.

Renegade334:
> IMO, there should be more overall changes, not just special treatment of `Response`.

+1 to not implementing a divergent declaration paradigm for a few arbitrarily-selected interfaces.

petdomaa100:
I recently found myself needing to override `Response.json` to only accept certain types. `ResponseConstructor` would have been a good solution, but it seems to have been rejected in favor of #222.

I'm not familiar with this project's internal roadmap, but #222 has been open for a very long time. Maybe consider merging this as a temporary solution.

PR #1942, opened 2025-03-20
Make most textContent not nullable

### LINKED ISSUE #1939: Element.textContent should not be nullable
`textContent` is inherited from Node, but in an `Element` context it is never `null`. It can be added to `Element` as `string` instead of `string | null` to eliminate unnecessary null checks.

alex-kinokon:
Closes #1939 and https://github.com/microsoft/TypeScript/issues/10315.


alex-kinokon:

inputfiles/addedTypes.jsonc
```diff
@@ -11,6 +11,19 @@
                     ]
                 }
             },
+            "Comment": {
```
saschanaz, line 14: There's no mixin named Comment and thus this doesn't do anything. Comment inherits from CharacterData so this is not needed either.

saschanaz (commented):
Thanks! One thing, given [the spec](https://dom.spec.whatwg.org/#get-text-content) also mentions DocumentFragment, can you also cover that here?

saschanaz:
(Making it `textContent: null;` for other Node-inheriting interfaces would be a bonus)

VisionAustraliaLuis:
Hello all, is this pull request going to be reviewed so that it can be merged? Thanks!

alex-kinokon:
@saschanaz I addressed your feedback. Can’t figure out how to make it `null` for other interfaces for now so let’s fix that in a future PR.

saschanaz:
> @saschanaz I addressed your feedback. Can’t figure out how to make it `null` for other interfaces for now so let’s fix that in a future PR.

Hmm? Any specific reason why DocumentFragment is any different than what you already did?

alex-kinokon:
DocumentFragment is already covered.

saschanaz:
I'm confused, where is it covered?

VisionAustraliaLuis:
Thank you @alex-kinokon! One step closer for this pull request to be merged. I've been hating how much more I need to type to bypass this `null` situation with the `textContent` property.

alex-kinokon:
@saschanaz do you mind rereviewing it again

saschanaz:
LGTM

jakebailey (approved):

saschanaz:
LGTM

PR #1944, opened 2025-03-21
Support generic `Uint8Array<ArrayBuffer>`

rbuckton:
This updates the DOM to use `Uint8Array<ArrayBuffer>` in some cases where the specification expects it. This addresses an error reported in https://github.com/Microsoft/TypeScript/issues/61432 for `ajayyy/SponsorBlock`.

This now also fixes `AllowSharedBufferSource`, which fails to include `SharedArrayBuffer`.

Fixes https://github.com/microsoft/TypeScript/issues/61480


sandersn (approved):

src/build/helpers.ts
```diff
@@ -114,6 +114,25 @@ export function exposesTo(o: { exposed?: string }, target: string[]): boolean {
...
+  }
+  if (Array.isArray(obj)) {
+    return obj.map(clone) as any as T;
+  }
+  if (obj instanceof Map) {
+    return new Map([...obj.entries()].map(clone)) as any as T;
+  }
+  const result = {} as T;
+  for (const k in obj) {
+    if (Object.getOwnPropertyDescriptor(obj, k)) {
```
petamoriken, line 129: `Object.hasOwn` might be better.

```suggestion
    if (Object.hasOwn(obj, k)) {
```

saschanaz:
This should use ArrayBuffer by default than just overriding some cases, and use SAB+AB when `[AllowShared]` is used.

rbuckton:
@sandersn: The implementation changed significantly. Can you take another look?

saschanaz:
This adds 5.6, but is it worth? Can it be folded into 5.7?

inputfiles/overridingTypes.jsonc
```diff
@@ -3233,12 +3243,14 @@
...
                             "readonly": true,
-                            "overrideType": "ReadableStream<Uint8Array>"
+                            "type": "ReadableStream",
+                            "subtype": { "type": "Uint8Array" }
                         },
                         "writable": {
                             "name": "writable",
                             "readonly": true,
-                            "overrideType": "WritableStream<string>"
+                            "type": "WritableStream",
```
sandersn, line 3252: The output doesn't change that I could tell, so I guess this is an update to use the new subtype property for consistency?

sandersn (approved):
Makes sense to me. One question about the change to WritableStream in overridingtypes.json to make sure I understood it correctly.

inputfiles/overridingTypes.jsonc
saschanaz: Refraining from overrideType is a good change 👍🏻 (use of overrideType disables type dependency tracking)

eslint.config.mjs
```diff
@@ -22,6 +22,7 @@ export default typescriptEslint.config(
       "@typescript-eslint/no-explicit-any": 0,
       "@typescript-eslint/no-non-null-assertion": 0,
       "@typescript-eslint/no-var-requires": 0,
+      "prettier/prettier": ["error", { "endOfLine": "auto" }] // match .gitattributes
```
saschanaz, line 25: Can we change gitattribute though? The unit test currently is manually normalizing line end and we should just prefer LF everywhere.

inputfiles/overridingTypes.jsonc
saschanaz: (Though I would have preferred if this kind of change is file as a separate PR, as mixing it makes harder to understand it)

src/build/helpers.ts
```diff
@@ -4,6 +4,7 @@ import * as Browser from "./types.js";
 export const bufferSourceTypes = new Set([
   "ArrayBuffer",
   "SharedArrayBuffer",
+  "ArrayBufferLike",
```
saschanaz, line 7: This is for Web IDL keywords and not expected to include TS specific things. This should go elsewhere.

src/build/emitter.ts
```diff
@@ -387,17 +388,51 @@ export function emitWebIdl(
...
+        case "Uint8ClampedArray":
+        case "Int16Array":
+        case "Uint16Array":
+        case "Int32Array":
+        case "Uint32Array":
+        case "Float32Array":
+        case "Float64Array":
+        case "BigInt64Array":
+        case "BigUint64Array":
+        case "Float16Array":
```
saschanaz, line 430: Maybe make a subset of `bufferSourceTypes` in helpers? Otherwise it's prone to break when a new buffer source type appears.

src/build/emitter.ts
```diff
@@ -135,6 +135,7 @@ function isEventHandler(p: Browser.Property) {
 export interface CompilerBehavior {
   useIteratorObject?: boolean;
   allowUnrelatedSetterType?: boolean;
+  genericTypedArrays?: boolean;
```
saschanaz, line 138: Other fields are verbs, would be nice to use verb here too.

src/build/emitter.ts
```diff
@@ -387,17 +388,51 @@ export function emitWebIdl(
             : obj.type;
         types.push(...(obj.additionalTypes ?? []).map((t) => ({ type: t })));
 
+        // propagate `any`
         const converted = types.map(convertDomTypeToTsTypeWorker);
-        const isAny = converted.some((t) => t === "any");
-        return isAny ? "any" : converted.join(" | ");
+        if (converted.includes("any")) return "any";
```
saschanaz, line 393: Nice change but would be nice to return in a block for style consistency. (Maybe there should be an eslint config for this)

saschanaz (changes requested):

src/build/emitter.ts
```diff
@@ -412,24 +415,9 @@ export function emitWebIdl(
...
-        case "Uint32Array":
-        case "Float32Array":
-        case "Float64Array":
-        case "BigInt64Array":
-        case "BigUint64Array":
-        case "Float16Array":
-          subtypeString = obj.allowShared ? "ArrayBufferLike" : "ArrayBuffer";
-          break;
+    if (!subtypeString && compilerBehavior.useGenericTypedArrays) {
+      if (arrayBufferViewTypes.has(type)) {
```
saschanaz, line 419: Maybe add it in the same if condition 👀

saschanaz (commented):
Still would like to change gitattributes to use only LF, looks good otherwise!

sandersn (dismissed):

sandersn (approved):

jakebailey:
What happened to this PR? Main TS uses this, right?

saschanaz:
I believe it's just waiting for you to merge this 😛

saschanaz:
Please do so 🙂 (well if I had the permission... 😜)

jakebailey (approved):

jakebailey:
lol, of course.

image

@sandersn @rbuckton is this PR good to go?

sandersn (approved):

sandersn (approved):

PR #1950, opened 2025-03-28
chore: override OnErrorEventHandlerNonNull

### LINKED ISSUE #1821: onerror for Elements is incorrect
```ts
const img = document.createElement('img')

// type error:
img.onerror = (e: Event) => e
```

Screenshot 2024-10-07 at 8 03 39 PM

[playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAlgWwOYwLwwCYmAVwQUzCgDpgAnfAQynwFEAbfAogCgHJEk2BKAKF4D0AmAAkcACwD8k3p2Lh8ZMiDJoYLfAC4YtAG6Eo3NAD4Y+XkA)

Only `Window.onerror` should have the alternative signature, but not elements. See:

- https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event
- https://udn.realityripple.com/docs/Web/API/GlobalEventHandlers/onerror

Bashamega:
fixes #1821

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -28466,7 +28466,7 @@ interface OnBeforeUnloadEventHandlerNonNull {
 }
 
 interface OnErrorEventHandlerNonNull {
-    (event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error): any;
+    (event: Event, source?: string, lineno?: number, colno?: number, error?: Error): any;
 }
 
 interface PerformanceObserverCallback {
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -8527,7 +8527,7 @@ interface LockGrantedCallback<T> {
 }
 
 interface OnErrorEventHandlerNonNull {
-    (event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error): any;
+    (event: Event, source?: string, lineno?: number, colno?: number, error?: Error): any;
 }
 
 interface PerformanceObserverCallback {
--- a/baselines/sharedworker.generated.d.ts
+++ b/baselines/sharedworker.generated.d.ts
@@ -8465,7 +8465,7 @@ interface LockGrantedCallback<T> {
 }
 
 interface OnErrorEventHandlerNonNull {
-    (event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error): any;
+    (event: Event, source?: string, lineno?: number, colno?: number, error?: Error): any;
 }
 
 interface PerformanceObserverCallback {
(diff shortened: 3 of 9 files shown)


saschanaz:
This would break window.onerror as it does receive a string. We should probably limit this string use case to window and not expose it to other interfaces. Perhaps we can modify GlobalEventHandlers.onerror with the general `(error: Error)` and add the current onerror to WindowEventHandlers? But would that cause collision?

Bashamega:
> This would break window.onerror as it does receive a string. We should probably limit this string use case to window and not expose it to other interfaces. Perhaps we can modify GlobalEventHandlers.onerror with the general `(error: Error)` and add the current onerror to WindowEventHandlers? But would that cause collision?

Okay

Bashamega:
Hello @saschanaz 
Is this better?

saschanaz:
Hmm. Maybe we should first make some unit tests in https://github.com/microsoft/TypeScript-DOM-lib-generator/tree/main/unittests/files, something like:

```ts
// onerror.ts

// this should not show error
window.onerror = (message: string, source: string, lineno: number, colno: number, error: any) => {};

// this also should not show error
declare var img: HTMLImageElement;
img.onerror = (event: Event) => {};
```

The goal here would be to make both pass.

Bashamega:
Hello:)
Please check again @saschanaz

inputfiles/addedTypes.jsonc
```diff
@@ -545,6 +545,10 @@
                         },
                         "loading": {
                             "overrideType": "\"eager\" | \"lazy\""
+                        },
+                        "onerror": {
+                            "name": "onerror",
+                            "overrideType": "(event: Event) => void"
```
saschanaz, line 551: This would only solve HTMLImageElement, maybe it should go to HTMLElement to solve it for other HTML elements.

saschanaz: And the CI fails: `Error: generated/dom.generated.d.ts(11577,11): error TS2430: Interface 'HTMLImageElement' incorrectly extends interface 'HTMLElement'.`

Bashamega: Hello @saschanaz,  

I have applied the changes to all elements.  

I was facing a build error due to the lack of documentation on the JSON format. However, replacing `GlobalEventHandlers` with `Omit<GlobalEventHandlers, 'onerror'>` solves the issue.  

Maybe we could start documenting the format to help others in the future?

saschanaz: Omit sounds ok, but maybe you didn't push your change?

Documentations sounds nice. I wonder we can have a local schema file.

saschanaz: Maybe just adding some comments to existing type declarations would be okay.

Bashamega: Hello @saschanaz 
My question is how to override the type?
I have tried to add it to the overriding types but it didn't work.

PR #1953, opened 2025-03-30
fix(RequestInit): restrict body type based on method to improve type safety

### LINKED ISSUE #1678: fetch type "GET" or "HEAD" with a body should be a type error
You cannot make a GET or HEAD fetch request with a body. I would like this to result in a type error.

Bashamega:
fixes #1678

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -1953,7 +1953,7 @@ interface ReportingObserverOptions {
 
 interface RequestInit {
     /** A BodyInit object or null to set request's body. */
-    body?: BodyInit | null;
+    body?: 'GET' | 'HEAD' extends this['method'] ? never : BodyInit | null | null;
     /** A string indicating how the request will interact with the browser's cache to set request's cache. */
     cache?: RequestCache;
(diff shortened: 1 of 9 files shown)


baselines/dom.generated.d.ts
```diff
@@ -1953,7 +1953,7 @@ interface ReportingObserverOptions {
 
 interface RequestInit {
     /** A BodyInit object or null to set request's body. */
-    body?: BodyInit | null;
+    body?: 'GET' | 'HEAD' extends this['method'] ? never : BodyInit | null | null;
```
saschanaz, line 1956: This won't work if `method` is a general string, e.g.:

```ts
function download(url: URL, method: string, body: any) {
  return fetch(url, { method, body });
}
```

Or `method` is lowercased or something weird like "gET" (this still works in browsers as the browser internally [makes it uppercase](https://fetch.spec.whatwg.org/#concept-method-normalize)). Actually not sure this would work even if method is GET/HEAD, can you add a unit test?

Bashamega: Actually, when I think about it more, it seems that this is a missing feature in TypeScript. TS needs to have a Template Literal Type for case-insensitive matching.

PR #1955, opened 2025-03-31
Delete docs/mediasession.png

Bashamega:
We are not using this file anywhere in the codebase


saschanaz:
LGTM 👍

saschanaz:
Huh? I added it.

PR #1956, opened 2025-03-31
Refactor test script for improved readability and modularity

Bashamega:
This PR refactors the test script to enhance maintainability, readability, and efficiency.  

#### **Changes:**  
- Extracted reusable functions (`getFiles`, `readFileContent`) to reduce redundancy.  
- Simplified file comparison logic for better maintainability.  
- Introduced `compileSets` array to streamline and organize compilation tasks dynamically.  
- Improved error handling and logging for better debugging.  

#### **Benefits:**  
- **Cleaner code:** Eliminates duplication and improves readability.  
- **Easier maintenance:** More modular and extensible structure.  
- **Improved debugging:** Enhanced error messages for easier troubleshooting.


saschanaz (commented):
First of all can you keep the existing function style instead of moving to const? A style change like that should be done with a good consensus and should be done for all files for consistency. And I don't think we have consensus.

Bashamega:
Done @saschanaz

src/test.ts
```diff
@@ -10,157 +10,136 @@ const tscPath = new URL(
   import.meta.url,
 );
 
-function normalizeLineEndings(text: string): string {
-  return text.replace(/\r\n?/g, "\n");
+function normalizeLineEndings(text: string) {
+  text.replace(/\r\n?/g, "\n");
```
saschanaz, line 14: return?

src/test.ts
```diff
@@ -10,157 +10,136 @@ const tscPath = new URL(
...
-    // do nothing
+    return [];
   }
+}
 
-  let outputFiles: string[] = [];
+function readFileContent(filePath: URL) {
   try {
-    outputFiles = fs.readdirSync(outputFolder);
+    return normalizeLineEndings(fs.readFileSync(filePath, "utf-8"));
```
saschanaz, line 27: We already stat the file before calling this and thus doesn't expect any error here. Thus we should skip try-catch here to see actual failures if happens.

src/test.ts
```diff
@@ -10,157 +10,136 @@ const tscPath = new URL(
...
       }
-
       continue;
     }
 
-    if (baselineStats?.isDirectory() || outputStats?.isDirectory()) {
-      const childBaselineFolder = new URL(`${file}/`, baselineFolder);
-      const childOutputFolder = new URL(`${file}/`, outputFolder);
-      if (!compareToBaselines(childBaselineFolder, childOutputFolder)) {
+    if (fs.existsSync(baselinePath) || fs.existsSync(outputPath)) {
```
saschanaz, line 62: This is not same as detecting directory?

src/test.ts
```diff
@@ -10,157 +10,136 @@ const tscPath = new URL(
...
       }
-
-      continue;
     }
   }
   return true;
 }
 
-function compileGeneratedFiles(lib: string, ...files: string[]) {
+function compileGeneratedFiles(lib: string[] | string, ...files: string[]) {
```
saschanaz, line 76: But is `lib` ever an array here?

src/test.ts
```diff
@@ -10,157 +10,136 @@ const tscPath = new URL(
...
-        tscPath,
-      )} --strict --lib ${lib} --types --noEmit ${files
-        .map((file) => fileURLToPath(new URL(file, outputFolder)))
-        .join(" ")}`,
+      `node ${fileURLToPath(tscPath)} --strict --lib ${lib} --types --noEmit ${fileArgs}`,
     );
-  } catch (e: any) {
-    console.error(`Test failed: could not compile '${files.join(",")}':`);
-    console.error(e.stdout.toString());
-    console.error();
```
saschanaz, line 100: Skipping this would hide the actual error which would make debugging harder.

Bashamega:
I have fixed them @saschanaz

Bashamega:
Hello @saschanaz 
Can you please review this? I have updated the PR

jakebailey:
Why do we need to do this? This just seems like a plain refactor, but no behavior is changing. Just seems like churn?

Bashamega:
> Why do we need to do this? This just seems like a plain refactor, but no behavior is changing. Just seems like churn?

This is just a refactor, and making the code more readable and easy to maintain.

jakebailey:
Personally, I would not try and work on this kind of change; it means us maintainers have to review PRs to see what they're doing, if they contain any new bugs, for code which theoretically is already working fine. Other improvements are of course welcome, but for plain code changes like this, I don't really think it's worth the time spent compared to other changes (e.g., you've sent some good other ones!).

saschanaz:
It's indeed a bit hard to make sure all changes are correct, would be much easier to check when each change is done separately.

src/test.ts
```diff
@@ -14,85 +14,68 @@ function normalizeLineEndings(text: string): string {
...
+    const baselinePath = new URL(file, baselineFolder);
+    const outputPath = new URL(file, outputFolder);
 
-    const baseline = baselineStats?.isFile()
-      ? normalizeLineEndings(
-          fs.readFileSync(new URL(file, baselineFolder)).toString(),
-        )
-      : null;
+    const isBaselineFile =
+      fs.existsSync(baselinePath) && fs.statSync(baselinePath).isFile();
```
saschanaz, line 42: It's actually advised to *not* call `exists` before other calls, because think can potentially be racy and the file can go away after that call.

https://nodejs.org/docs/latest/api/fs.html#fsexistspath-callback

>Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile(), or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

Bashamega:
No problem. I will close this one

saschanaz:
I don't think this should be thrown away; just splitting this would be fine.

Bashamega:
I have made it smaller @saschanaz

src/test.ts
```diff
@@ -104,67 +104,36 @@ function compileGeneratedFiles(lib: string, ...files: string[]) {
...
-      "es2018",
-      "audioworklet.generated.d.ts",
-      "audioworklet.asynciterable.generated.d.ts",
-    )
-  ) {
+    modules.every((mod) =>
+      targets.every((target) =>
+        compileGeneratedFiles(
+          target,
+          ...suffixes[target].map((s) => `${mod}.${s}`),
```
saschanaz, line 127:
```suggestion
          ...suffixes[target].map((suffix) => `${mod}.${suffix}`),
```

(I know single character variables exist in this repo but that's probably something to refrain from)

saschanaz (commented):
Now this is what I call a great refactoring.

saschanaz:
LGTM

jakebailey (approved):
Thanks, this seems much more obviously correct.

PR #1967, opened 2025-04-03
fix: update index signature for NamedNodeMap to support both number and string keys

### LINKED ISSUE #344: NamedNodeMap is indexable by string
According to the spec (https://dom.spec.whatwg.org/#interface-namednodemap) and experimentally in Chrome, you can index a NamedNodeMap with a string and get back an Attr.

The TS definition only has:

```ts
interface NamedNodeMap {
    // [...]
    [index: number]: Attr;
}
```

Bashamega:
fixes #344 
MDN: https://developer.mozilla.org/en-US/docs/Web/API/NamedNodeMap/getNamedItem

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -16879,7 +16879,7 @@ interface NamedNodeMap {
     setNamedItem(attr: Attr): Attr | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/NamedNodeMap/setNamedItemNS) */
     setNamedItemNS(attr: Attr): Attr | null;
-    [index: number]: Attr;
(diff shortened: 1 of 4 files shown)


saschanaz:
The test failure is exactly the reason why this hasn't been done (please run tests before submitting PR)

saschanaz:
I think there's a TypeScript side issue but can't remember right now.

Bashamega:
> The test failure is exactly the reason why this hasn't been done (please run tests before submitting PR)

Will do

saschanaz:
I think that's too hacky, and given that this indexer is legacy anyway (use the methods instead whenever possible) I think it's okay to declare wontfix.

Bashamega:
Okay

PR #1970, opened 2025-04-05
Sort and Format JSONC Build Files

Bashamega:
#### Overview
This pull request introduces a script that formats and sorts JSONC (JSON with comments) files alphabetically. The script is designed to enhance the readability and maintainability of build configuration files by ensuring that the keys within these files are consistently ordered.

#### Changes Made
- **Added a sorting function**: A helper function `sortObjectKeys` has been implemented to recursively sort the keys of an object. This function ensures that all nested objects are also sorted, providing a comprehensive sorting solution.
- **File processing**: The `sortFiles` function reads specified JSONC files, parses them while preserving comments, sorts their keys, and then writes the sorted content back to the original files.

#### Code Explanation
The script is implemented in TypeScript and utilizes the `fs` module for file operations and the `comment-json` package to handle JSONC parsing and stringification. Below is a breakdown of the key components:

1. **sortObjectKeys Function**:
   - This function takes an object as input and checks if it is indeed an object and not an array.
   - It sorts the keys of the object alphabetically and recursively applies the same sorting to any nested objects.
   - If the input is not an object, it returns the input as is.

   ```typescript
   function sortObjectKeys<T>(obj: T): T {
     if (obj && typeof obj === "object" && !Array.isArray(obj)) {
       const sorted = {} as any;
       Object.keys(obj)
         .sort()
         .forEach((key) => {
           sorted[key] = sortObjectKeys((obj as any)[key]);
         });
       return sorted;
     }
     return obj;
   }
   ```

2. **sortFiles Function**:
   - This function defines an array of file names that need to be processed.
   - It constructs the file path for each JSONC file, reads its content, and parses it while preserving comments.
   - The parsed object is then sorted using the `sortObjectKeys` function.
   - Finally, the sorted object is converted back to JSONC format and written back to the original file.

   ```typescript
   export function sortFiles() {
     const files = ["overridingTypes", "removedTypes", "addedTypes"];
     files.forEach((file) => {
       const filePath = new URL(`../../inputfiles/${file}.jsonc`, import.meta.url);
       const fileContent = fs.readFileSync(filePath, "utf-8");
       const parsed = commentJson.parse(fileContent, undefined, true);
       const sortedObject = sortObjectKeys(parsed);
       const sortedJsonC = commentJson.stringify(sortedObject, null, 2);
       fs.writeFileSync(filePath, sortedJsonC, "utf-8");
     });
     console.log("JSONC file keys sorted successfully.");
   }
   ```

#### Usage
To use this script, simply call the `sortFiles` function. Ensure that the specified JSONC files are located in the correct directory as indicated in the file path construction.

#### Benefits
- **Improved Readability**: Sorting keys alphabetically makes it easier to locate specific entries within the JSONC files.
- **Consistency**: Ensures that the structure of the configuration files remains consistent across different environments and updates.

#### Next Steps
- Review the code for any potential improvements or optimizations.
- Consider adding unit tests to verify the functionality of the sorting and formatting process.

Thank you for reviewing this pull request! I look forward to your feedback.

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -13,6 +13,7 @@ import { getInterfaceElementMergeData } from "./build/webref/elements.js";
 import { getInterfaceToEventMap } from "./build/webref/events.js";
 import { getWebidls } from "./build/webref/idl.js";
 import jsonc from "jsonc-parser";
+import { sortFiles } from "./build/sorter.js";
 
 function mergeNamesakes(filtered: Browser.WebIdl) {
   const targets = [
@@ -396,3 +397,4 @@ async function emitDom() {
 }
 
 await emitDom();
+sortFiles();
--- a/src/build/sorter.ts
+++ b/src/build/sorter.ts
@@ -1,0 +1,41 @@
+import * as fs from "fs";
+import commentJson from "comment-json";
+
+// A helper function to recursively sort object keys alphabetically.
+function sortObjectKeys<T>(obj: T): T {
+  if (obj && typeof obj === "object" && !Array.isArray(obj)) {
+    const sorted = {} as any;
+    Object.keys(obj)
+      .sort()
+      .forEach((key) => {
+        // Recursively sort keys if the value is an object
+        sorted[key] = sortObjectKeys((obj as any)[key]);
+      });
+    return sorted;
+  }
+  return obj;
+}
+
+export function sortFiles() {
+  const filePath = new URL(
+    "../../inputfiles/overridingTypes.jsonc",
+    import.meta.url,
+  ); // Replace with your JSONC file path
+
+  // Read the JSONC file content
+  const fileContent = fs.readFileSync(filePath, "utf-8");
+
+  // Parse the JSONC file while preserving comments
+  const parsed = commentJson.parse(fileContent, undefined, true);
+
+  // Sort the object keys alphabetically (recursively)
+  const sortedObject = sortObjectKeys(parsed);
+
+  // Stringify the sorted object back to JSONC format (preserving comments)
+  const sortedJsonC = commentJson.stringify(sortedObject, null, 2);
+
+  // Write the sorted content back to the file
+  fs.writeFileSync(filePath, sortedJsonC, "utf-8");
+
+  console.log("JSONC file keys sorted successfully.");
+}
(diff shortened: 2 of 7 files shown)


src/build/sorter.ts
```diff
@@ -0,0 +1,41 @@
+import * as fs from "fs";
+import commentJson from "comment-json";
```
saschanaz, line 2: We already use jsonc-parser, should be able to use here.

The important thing is that the comments should be there, while it's all being stripped out in this PR.

saschanaz: I think we don't need to check in a full sorter, it should be enough to check whether it's sorted and throw error if not when linting.

src/build.ts
```diff
@@ -396,3 +397,4 @@ async function emitDom() {
 }
 
 await emitDom();
+sortFiles();
```
saschanaz, line 400: Shouldn't be part of build, it should rather be part of `npm run lint`.

src/build/sorter.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (lastChar === ",") {
+        foundTrailingComma = true;
+      }
+    },
+  });
+
+  return foundTrailingComma;
+}
+
+export function sortFiles(): void {
```
saschanaz, line 39: It's not about sorting anymore, it's about verifying it's sorted. And it's not sorting "files", it's sorting fields.

src/build/sorter.ts
```diff
@@ -0,0 +1,67 @@
...
+// Recursively sort object keys
+function sortObjectKeys<T>(obj: T): T {
+  if (obj && typeof obj === "object" && !Array.isArray(obj)) {
+    const sorted: Record<string, unknown> = {};
+    for (const key of Object.keys(obj).sort()) {
+      sorted[key] = sortObjectKeys((obj as any)[key]);
+    }
+    return sorted as T;
+  }
+  return obj;
```
saschanaz, line 14: This do not need to return an object, maybe do not return anything and throw if it's not sorted.

saschanaz (commented):
What code did you use to sort it one-time though? Might be nice to be able to validate one-time.

Bashamega:
Hello @saschanaz 
I'm sorry, your requests fell out of my inbox. I think if you want it checked only once, we can create a workflow for it, so it will check once the user opens a PR.
I will create the workflow for it today

PR #1972, opened 2025-04-11
Upgrade to BCD 6

saschanaz:
This is a major version bump that needs Microsoft approval.


jakebailey (approved):
This seems correct.

saschanaz:
But merging this doesn't work even with approval, so this still needs manual merging.

LGTM

PR #1985, opened 2025-04-17
Add CodeQL configuration for TypeScript project analysis

Bashamega:
Updated the configuration file


jakebailey:
I don't think we need this; the defaults should be auto detecting everything. For example, see the one in the main TS repo: https://github.com/microsoft/TypeScript/blob/main/.github/codeql/codeql-configuration.yml

No setting of the languages or queries except to exclude stuff.

PR #1993, opened 2025-04-23
Add more comments

Bashamega:
related to #1940 
Hello:)
In my last PR I have used MDN to generate comments for interface declarations, in this pr I add it for the properties.


Bashamega:
Not ready for review

saschanaz (commented):
I'm yet to review this, but please do not convert to sync APIs.

Bashamega:
> I'm yet to review this, but please do not convert to sync APIs.

Sure

Bashamega:
Hello @saschanaz 
I have switched the old functions to be async, but the new function is still sync, because the emitter is not async. I have tried switch it to be async but it broke.

baselines/audioworklet.iterable.generated.d.ts
```diff
@@ -13,10 +13,7 @@ interface URLSearchParamsIterator<T> extends IteratorObject<T, BuiltinIteratorRe
 
 interface URLSearchParams {
     [Symbol.iterator](): URLSearchParamsIterator<[string, string]>;
-    /** Returns an array of key, value pairs for every entry in the search params. */
```
jakebailey, line 16: Seems like a loss?

src/build/emitter.ts
```diff
@@ -906,7 +907,10 @@ export function emitWebIdl(
       comments.push("Available only in secure contexts.");
     }
     if (entity.mdnUrl) {
-      if (comments.length > 0) comments.push("");
+      if (comments.length == 0) {
+        comments.push(extractSummaryFromFile(entity.mdnUrl));
```
saschanaz, line 911: Emitter should just emit the typescript lib based on the given data, rather than generating data. Any required generation should be done in build.ts and passed to this function.

src/build/mdn-comments.ts
```diff
@@ -110,3 +111,25 @@ export async function generateDescription(): Promise<Record<string, string>> {
 
   return {};
 }
+export function extractSummaryFromFile(url: string): string {
+  const relativePath = url
+    .replace("https://developer.mozilla.org/docs/", "")
+    .split("#")[0]
+    .toLowerCase();
```
saschanaz, line 118: We should compute the path rather than using the URL this way. Basically for `Foo.bar` the path should be `/api/foo/bar/index.md`.

And why do we need this when we already have `generateDescriptions` above that collects index files?

Bashamega:
I have updated the context to not call the API in the emitter, but that cause the test to fail, I will fix it later

Bashamega:
Done @saschanaz

src/build.ts
```diff
@@ -45,38 +45,42 @@ async function emitFlavor(
...
   forceKnownTypes: Set<string>,
   options: EmitOptions,
+  descriptions: Record<string, string>,
 ) {
   const exposed = getExposedTypes(webidl, options.global, forceKnownTypes);
   mergeNamesakes(exposed);
   exposed.events = webidl.events;
 
-  const result = emitWebIdl(
+  const result = await emitWebIdl(
```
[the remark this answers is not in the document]
Bashamega: Done

src/build/mdn-comments.ts
```diff
@@ -43,70 +42,67 @@ function extractSummary(markdown: string): string {
...
-      withFileTypes: true,
-    });
-    return entries
-      .filter((entry) => entry.isDirectory())
-      .map((entry) => new URL(entry.name + "/", dirPath));
-  } catch (error) {
-    console.error("Error reading directories:", error);
-    return [];
-  }
+  return sentenceMatch ? sentenceMatch[0] : normalizedText.split(" ")[0] || "";
```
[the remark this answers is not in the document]
Bashamega: done

src/build/mdn-comments.ts
```diff
@@ -43,70 +42,67 @@ function extractSummary(markdown: string): string {
...
-    const folders = await getDirectories(basePath);
-    if (folders.length > 0) {
-      return await getIndexMdContents(folders);
+    const indexPaths = await walkDirectory(basePath);
+
+    for (const fileURL of indexPaths) {
+      try {
+        const content = await fs.readFile(fileURL, "utf-8");
+
+        const titleMatch = content.match(/title:\s*["']?([^"'\n]+)["']?/);
```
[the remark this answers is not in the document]

src/build/emitter.ts
```diff
@@ -906,7 +907,18 @@ export function emitWebIdl(
...
+          .split("/API/")
+          .pop()
+          ?.replace("/", ".")
+          .split("#")[0]
+          .toLowerCase();
+        if (key && descriptions[key]) {
+          comments.push(descriptions[key]);
+        }
+      }
+      comments.push("");
```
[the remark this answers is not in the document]
Bashamega: I don't know how to do that, because the `mergeApiDescriptions` only runs through the main interfaces, not the properties.
If you don't want to keep it in the emitter, we would need to modify the merge function.

Bashamega:
Hello @saschanaz 
Do you have any other change requests in this PR?

src/build/emitter.ts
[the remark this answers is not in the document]
saschanaz: You can go deeper from an interface via `.properties`, then you can compute the key.

I would prefer more MDN-like structure where members are nested like rather than the current string concatenation though:

```json
{
  "interface1": {
    "__comment": "(comment)",
    "member1": {
      "__comment":  "(comment)"
    },
  }
}
```

Bashamega:
Hello @saschanaz 
I have updated the emitter but I think the formatted is better to be a separate issue

src/build/mdn-comments.ts
```diff
@@ -47,66 +47,73 @@ function extractSummary(markdown: string): string {
...
+  const results: Record<string, string> = {};
   try {
-    const folders = await getDirectories(basePath);
-    if (folders.length > 0) {
-      return await getIndexMdContents(folders);
+    const indexPaths = await walkDirectory(basePath);
+
+    for (const fileURL of indexPaths) {
+      try {
+        const content = await fs.readFile(fileURL, "utf-8");
```
saschanaz, line 94: How much does this take? Should this be called in parallel with Promise.all?

Bashamega: Nice catch

Bashamega:
I have updated the script to use promise @saschanaz

Bashamega:
Hello @saschanaz 
Do you have any changes?

src/build/emitter.ts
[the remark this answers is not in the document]
saschanaz: I still think it should be a nested object than string concat as we use nested objects everywhere while string concat is used nowhere.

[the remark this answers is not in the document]
Bashamega: Sure

Bashamega:
Hello @saschanaz 
Sorry, I was very busy. It seems that the old `comments.json` file was using the same format that I am using. If that format is incorrect, we should create another PR for that.

saschanaz:
Not sure I follow? Following the format of the current comments.json would be actually ideal, but I don't think this PR is currently doing that.

The comment.json:

```json
{
  "InterfaceName": {
    "properties": {
      "property": {
        "foo": {
          "comment": "(comment)"
        }
      }
    }
  }
}
```

This PR right now:

```json
{
  "InterfaceName.foo": "(comment")
}
```

Which looks more concise but with extra conversion step to the comment.json format anyway. Since we are automating it here it should directly emit the comment.json format.

Bashamega:
Sorry, i got confused. I will fix it

Bashamega:
Hello @saschanaz 
I have updated the api

src/build/mdn-comments.ts
```diff
@@ -69,44 +68,60 @@ async function walkDirectory(dir: URL): Promise<URL[]> {
   return results;
 }
 
-function generateTitle(content: string, file: URL) {
-  const match = content.match(/title:\s*["']?([^"'\n]+)["']?/);
-  if (match) {
-    return match[1].replace(/ extension$/, "").split(":")[0];
+function generateSlug(content: string): string {
+  const match = content.match(/slug:\s*["']?([^"'\n]+)["']?/);
```
saschanaz, line 72: Is there anything that `^slug: (.+)` can't cover?

src/build/mdn-comments.ts
```diff
@@ -69,44 +68,60 @@ async function walkDirectory(dir: URL): Promise<URL[]> {
...
+  if (!match) throw new Error("Slug not found");
+  const url = match[1].split(":").pop()!;
+  const parts = url.split("/").slice(2); // remove first 2 segments
+  const result = parts.join("/");
+  return result;
```
saschanaz, line 77: It seems this is going to be split again, so maybe just return an array?

src/build/mdn-comments.ts
```diff
@@ -69,44 +68,60 @@ async function walkDirectory(dir: URL): Promise<URL[]> {
...
+}
+
+function urlToNestedObject(url: string, text: string): Record<string, any> {
+  const keys = url.split("/");
+  return keys.reduceRight<Record<string, any>>((acc, key) => ({ [key]: acc }), {
+    __comment: text,
+  });
+}
+
+function deepMerge(
```
saschanaz, line 87: Reuse merge() from build/utils.ts?

src/build/mdn-comments.ts
```diff
@@ -69,44 +68,60 @@ async function walkDirectory(dir: URL): Promise<URL[]> {
...
+function generateSlug(content: string): string {
+  const match = content.match(/slug:\s*["']?([^"'\n]+)["']?/);
+  if (!match) throw new Error("Slug not found");
+  const url = match[1].split(":").pop()!;
+  const parts = url.split("/").slice(2); // remove first 2 segments
+  const result = parts.join("/");
+  return result;
+}
+
+function urlToNestedObject(url: string, text: string): Record<string, any> {
```
saschanaz, line 80: Clever approach! I wonder we can skip merge() though, if we do something like:

```js
function ensureLeaf(obj, keys) {
  let leaf = obj;
  for (const key of keys) {
    leaf[key] ??= {};
    leaf = leaf[key]
  }
  return leaf;
}
```

Then we can do:

```js
const result = {};
const leaf = ensureLeaf(result, slug); // e.g. when slug is ["AbortSignal", "abort"]
leaf.__comment = "(comment)"
```

src/build.ts
```diff
@@ -129,32 +129,44 @@ async function emitDom() {
...
+      const { properties, methods } = idl.interfaces?.interface?.[key] || {};
+
+      const propertyItems = properties?.property
+        ? Object.values(properties.property)
+        : [];
+      const methodItems = methods?.method ? Object.values(methods.method) : [];
+
+      // Add comments to properties
+      for (const prop of propertyItems) {
+        const propDesc = descObject[prop.name];
```
saschanaz, line 159: This is still less than ideal, the ideal would be that being able to just `merge()` here.

We can extract whether it's property or method from the `page-type:` metadata, e.g. `web-api-instance-property`, `web-api-instance-method`, `web-api-static-property`, `web-api-static-method`, etc.

With that we could make the generateSlug below to generate the keys for full `Browser.Interface`-compliant structure, e.g. `["AbortSignal", "properties", "property", "aborted"]`.

If the full compliance is achieved then we can just do `merge()` here.

(Unfortunately MDN doesn't have much to tell whether it's for namespace though. We should keep the current line 141.)

src/build/mdn-comments.ts
Bashamega: It can cover everything. I have updated it.

src/build/mdn-comments.ts

src/build/mdn-comments.ts

src/build/mdn-comments.ts

src/build.ts
Bashamega: Hello @saschanaz 
I have tried to use merge, but when I use merge it returns a bunch of errors for the emitter, like:
```
Error: Missing 'type' field in {"signal":"The **`signal`** read-only property of the AbortController interface returns an AbortSignal object instance, which can be used to communicate with/abort an asynchronous operation as desired."}
```
I think we can't use it

saschanaz: The object looks like it's not using `{ "__comment": "..." }`, in that case it would overwrite on existing object and fail

Bashamega: I have updated the script and it worked, but some types are structured differently, for example, the CanvasRenderingContext2D extends other kinds. So I got this error:
Error: Missing 'type' field in {"comment":"The **`CanvasRenderingContext2D.direction`** property of the Canvas 2D API specifies the current text direction used to draw text."}

Bashamega:
Hello @saschanaz 
I have updated the script and it worked, but some types are structured differently; for example, the CanvasRenderingContext2D extends other types. So I got this error:
Error: Missing 'type' field in {"comment":"The CanvasRenderingContext2D.direction property of the Canvas 2D API specifies the current text direction used to draw text."}

saschanaz:
Hmmmm, I wonder we should actually get a different version of `merge()` that bails out early when there's no corresponding object. I'll try your pr and try figuring it out. Still I think we are in the right direction, so thank you!

saschanaz:
> Hello @saschanaz I have updated the script and it worked, but some types are structured differently; for example, the CanvasRenderingContext2D extends other types. So I got this error: Error: Missing 'type' field in {"comment":"The CanvasRenderingContext2D.direction property of the Canvas 2D API specifies the current text direction used to draw text."}
> 
> If you want you can cherry-pick this pr

Seems you haven't pushed the failing version, can you push it?

Bashamega:
> Hmmmm, I wonder we should actually get a different version of `merge()` that bails out early when there's no corresponding object. I'll try your pr and try figuring it out. Still I think we are in the right direction, so thank you!

You are welcome. 
I will push the latest changes that are broken right now, so you can work on them.

Bashamega:
I have pushed my latest changes @saschanaz

Bashamega:
Hello @saschanaz 
I have updated the emitter to exclude undefined types. It should be working now

saschanaz:
I don't think it should ignore missing name, it should rather throw. And I wonder how it's happening...

Bashamega:
> I don't think it should ignore missing name, it should rather throw. And I wonder how it's happening...

Can you try to check it?

Bashamega:
Hello @saschanaz 
Sorry for the ping, can you check this pr?

saschanaz:
Just had some long weekend with friends here in Germany, I'll try to take another look today 🙏

Bashamega:
> Just had some long weekend with friends here in Germany, I'll try to take another look today 🙏

No problem, take your time

inputfiles/removedComments.json
```diff
@@ -0,0 +1,9 @@
+[
+    "DOMHighResTimeStamp",
+    "PublicKeyCredentialCreationOptions",
+    "PublicKeyCredentialRequestOptions",
+    "MediaTrackConstraints",
+    "MediaTrackSupportedConstraints",
+    "MediaTrackSettings",
+    "RequestInit"
+]
```
saschanaz: With the new conditional merge(), this list can go as these items will be automatically filtered out by the function.

src/build.ts
```diff
@@ -129,19 +130,28 @@ async function emitDom() {
...
+    )) {
+      const target = idl.interfaces?.interface?.[key] || namespaces[key];
+
+      if (!target) continue;
+
+      // Set interface/class-level comment
+      if (descObject.__comment) {
+        target.comment = descObject.__comment;
       }
     }
```
saschanaz:
```suggestion
     for (const [key, target] of Object.entries(namespaces)) {
      const descObject = descriptions.interfaces.interface[key];
      if (!target) continue;

      merge(target, descObject, { optional: true });
    }
```

1. Only namespaces need this loop as MDN doesn't have namespace-specific page type. Interfaces are being handled by the revised `merge()` 👍🏻
2. Would be a bit more efficient to iterate over namespaces than interfaces, because the list of the former is much shorter.

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,124 @@ function extractSummary(markdown: string): string {
...
-      // Improved title extraction
-      const titleMatch = content.match(/title:\s*["']?([^"'\n]+)["']?/);
-      const filename = basename(folder.toString());
-      const title = titleMatch
-        ? titleMatch[1].replace(/ extension$/, "")
-        : filename || "";
+function generateSlug(content: string): string[] {
+  const match = content.match(/slug:\s*["']?([^"'\n]+)["']?/)!;
+  const url = match[1].split(":").pop()!;
+  const parts = url.split("/").slice(2); // skip `/en-US/web/api/...`
```
```suggestion
  const normalized = url.endsWith("_static") ? url.slice(0, -7) : url;
  const parts = normalized.split("/").slice(2); // skip `/en-US/web/api/...`
```

This will add some missing comments for static members. (`_static` postfix is added for any static members as MDN people considered any non-static member name will never end with `_static`. Thus it should be safe to remove it without worrying about potential collision)

src/build/mdn-comments.ts
saschanaz: Seems this is still valid, but my suggestion was wrong. `\nslug: (.+)\n` works though. There's certainly no quote so no need to process it.

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,124 @@ function extractSummary(markdown: string): string {
...
-): Promise<{ [key: string]: string }> {
-  const results: { [key: string]: string } = {};
+const types: Record<string, string> = {
+  property: "properties",
+  method: "methods",
+  "landing-page": "ignore",
+  interface: "interfaces",
+  overview: "ignore",
+  event: "events",
+};
```
```suggestion
const paths: Record<string, string[]> = {
  "web-api-instance-property": ["properties", "property"],
  "web-api-static-property": ["properties", "property"],
  "web-api-instance-method": ["methods", "method"],
  "web-api-static-method": ["methods", "method"],
  "web-api-interface": [],
  "webgl-extension": [],
  "webgl-extension-method": ["methods", "method"],
};
```

(To make the generate function more simpler below. Also removed `events` as this PR doesn't really handle it. That can be done in a followup.)

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,124 @@ function extractSummary(markdown: string): string {
...
 
-  for (const folder of folders) {
-    const indexPath = new URL("index.md", folder);
+function generateTypes(content: string): string[] | undefined {
+  const pageType = content.match(/page-type:\s*["']?([^"'\n]+)["']?/)!;
+  const type = pageType[1].split("-").pop()!;
+  const plural = types[type];
+  if (!plural) return;
```
```suggestion
function generatePath(content: string): string[] | undefined {
  const pageType = content.match(/\npage-type: (.+)\n/)!;
  const type = pageType[1];
  return paths[type];
```

1. This function certainly processes page type but doesn't really return type. Let's call it a path.
2. Splitting by "-" is potentially problematic for `web-api-constructor` as the member `constructor` exists for any object. We could make it a map instead of an object though.

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,124 @@ function extractSummary(markdown: string): string {
...
+    iface.__comment = summary;
+  } else {
+    const [ifaceName, memberName] = slug;
+    const iface = ensureLeaf(root, [ifaceName, ...types]);
+
+    const mainComment = root[ifaceName]?.__comment;
+    if (mainComment !== summary) {
+      iface[memberName] = { comment: summary };
+    }
+  }
```
```suggestion
  paths: string[],
) {
  if (!paths.length) {
    const iface = ensureLeaf(root, slug);
    iface.comment = summary;
  } else {
    const [ifaceName, memberName] = slug;
    const target = ensureLeaf(root, [ifaceName, ...paths, memberName]);
    target.comment = summary;
  }
```

1. `comment` should be used instead of `__comment`.
2. `mainComment !== summary` check doesn't seem to do anything meaningful, it can go.

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,124 @@ function extractSummary(markdown: string): string {
...
+          )
+            console.log(types, slug, summary);
+          insertComment(results, slug, summary, types);
+        } catch (error) {
+          console.error("Error generating API descriptions:", error);
+        }
+      }),
+    );
+  } catch (err) {
+    console.error("Error generating API descriptions:", err);
```
saschanaz: Please remove any try-catch here and let the code throw.

saschanaz (commented):
Sorry for the delay!

Reviewing would be much easier if I could just push some straightforward patches into the PR branch. Could you try opening a PR into https://github.com/saschanaz/types-web with the exact same branch of yours? That could allow me to push directly.

Bashamega:
Hello @saschanaz 
I have updated the PR as requested, and missing comments have been added. Unfortunately, when I tried to fork your fork, GitHub wouldn't let me, because I have a fork of the original.

saschanaz:
> Hello @saschanaz I have updated the PR as requested, and missing comments have been added. Unfortunately, when I tried to fork your fork, GitHub wouldn't let me, because I have a fork of the original.

You do not need to fork it, only filing a PR would be enough.

saschanaz:
This way: https://github.com/saschanaz/types-web/compare/main...Bashamega:TypeScript-DOM-lib-generator:descriptions

src/build.ts
```diff
@@ -129,19 +129,22 @@ async function emitDom() {
...
     );
-    for (const [key, value] of Object.entries(descriptions)) {
-      const target = idl.interfaces!.interface[key] || namespaces[key];
-      if (target) {
-        target.comment = value;
-      }
+
+    for (const [key, target] of Object.entries(namespaces)) {
+      const descObject = descriptions.interfaces.interface[key];
+      if (!target) continue;
```
saschanaz:
```suggestion
      if (!descObject) continue;
```

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,109 @@ function extractSummary(markdown: string): string {
...
+      const types = generateTypes(content);
+      if (!slug || slug.length === 0 || (types && types[0] === "ignore"))
+        return;
+
+      const summary = extractSummary(content);
+      if (
+        summary ===
+        "When writing code for the Web, there are a large number of Web APIs available."
+      )
+        console.log(types, slug, summary);
```
```suggestion
```

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,109 @@ function extractSummary(markdown: string): string {
...
+  const results: Record<string, any> = {};
+  const indexPaths = await walkDirectory(basePath);
+
+  await Promise.all(
+    indexPaths.map(async (fileURL) => {
+      const content = await fs.readFile(fileURL, "utf-8");
+      const slug = generateSlug(content);
+      const types = generateTypes(content);
+      if (!slug || slug.length === 0 || (types && types[0] === "ignore"))
+        return;
```
```suggestion
      const types = generateTypes(content);
      if (!slug || slug.length === 0 || !types)
        return;
```

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,109 @@ function extractSummary(markdown: string): string {
...
+
+      const summary = extractSummary(content);
+      if (
+        summary ===
+        "When writing code for the Web, there are a large number of Web APIs available."
+      )
+        console.log(types, slug, summary);
+      insertComment(results, slug, summary, types || []);
```
```suggestion
      insertComment(results, slug, summary, types);
```

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,109 @@ function extractSummary(markdown: string): string {
...
+  "webgl-extension-method": ["methods", "method"],
+};
 
-  for (const folder of folders) {
-    const indexPath = new URL("index.md", folder);
+function generateTypes(content: string): string[] | undefined {
+  const pageType = content.match(/page-type:\s*["']?([^"'\n]+)["']?/)!;
+  const type = pageType[1];
+  const plural = paths[type];
+  if (!plural) return;
```
saschanaz: You can just return `paths[type]` directly. Also given this return a "path", would be nice to name the function `generatePath`.

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,109 @@ function extractSummary(markdown: string): string {
...
-    });
-    return entries
-      .filter((entry) => entry.isDirectory())
-      .map((entry) => new URL(entry.name + "/", dirPath));
-  } catch (error) {
-    console.error("Error reading directories:", error);
-    return [];
+async function walkDirectory(dir: URL): Promise<URL[]> {
+  const entries = await fs.readdir(dir, { withFileTypes: true });
+  const parentDirName = dir.pathname.split("/").filter(Boolean).slice(-1)[0];
```
```suggestion
  const parentDirName = dir.pathname.split("/").at(-1);
```

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,109 @@ function extractSummary(markdown: string): string {
     return sentenceMatch[0]; // Return the first full sentence
   }
 
-  return normalizedText.split(" ")[0] || ""; // Fallback: first word if no sentence found
+  const firstWord = normalizedText.split(" ")[0];
+  return firstWord || "";
```
saschanaz, line 49: We should probably drop this, `RTCTrackEvent.track` only gets "The" currently. But it's fine to fix that in a followup.

src/build/mdn-comments.ts
```diff
@@ -47,66 +45,109 @@ function extractSummary(markdown: string): string {
...
+  return plural;
+}
 
-      // Improved title extraction
-      const titleMatch = content.match(/title:\s*["']?([^"'\n]+)["']?/);
-      const filename = basename(folder.toString());
-      const title = titleMatch
-        ? titleMatch[1].replace(/ extension$/, "")
-        : filename || "";
+function generateSlug(content: string): string[] {
```
saschanaz: (This is actually more like extractSlug)

Bashamega:
Thanks @saschanaz 
I didn't know

saschanaz (commented):
A few nits, then I think it's good to go.

saschanaz:
LGTM 🎉

Bashamega:
Thank you very much @saschanaz

jakebailey (approved):
After this, there's nothing else to comment on, right? This is as large as we're going to get?

Each one of these makes parsing the DOM types slower, so I'm hoping that we stop at some point 😅

saschanaz:
Technically we still have events, but adding comments there would be a bit awkward as that would only work for `onevent` attributes and not for `addEventListener` calls. Probably fair to just skip that.

saschanaz:
LGTM

Bashamega:
Hello @jakebailey 
First of all, thank you very much for merging this pr.
Secondly, I think we can improve the compiler's speed by caching comments in a JSON file, which will serve as the comments reference. The GitHub action will be responsible for updating it. If this sound good to you, I can draft a pr for it and compare the speed.
Also, there are still some missing comments.

jakebailey:
I don't know what you mean, sorry. The comments have to be in a declaration file. Moving them elsewhere won't speed things up (and will make the compiler/editor stuff challenging).

Bashamega:
> I don't know what you mean, sorry. The comments have to be in a declaration file. Moving them elsewhere won't speed things up (and will make the compiler/editor stuff challenging).

I mean instead of fetching the data each time you run the generate command, we can put the comments in a json file and the builder will use that json file instead of fetching the summaries every time

HolgerJeromin:
I am pretty sure Jake meant the usage (in the editor and compile of user code) of the generated files will slow down the typescript compiler as it has to parse more data.
Build/generation time of these files is uninteresting.
tsdoc belongs into the `.d.ts` file and not in a (faster?) parallel world which needs to be invented and maintained.

PR #1039, opened 2021-07-02
Add DefinitelyTyped tests to CI

orta:
Fixes #389

PR #1994, opened 2025-04-23
Automate TypeScript Definitions Update from DefinitelyTyped

### LINKED ISSUE #389: CI test with DefinitelyTyped
Sometimes compatibility with DefinitelyTyped breaks, a test will be great to prevent this.

Bashamega:
Fixes #389

Overrides #1039

### DIFF
--- a/.github/workflows/definitely-typed.yml
+++ b/.github/workflows/definitely-typed.yml
@@ -1,0 +1,35 @@
+name: Test DefinitelyTyped
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


jakebailey:
Also note that the workflow isn't actually testing anything... Zero tests ran and that's why it appears to be fast.

saschanaz:
(In general I think it would be good if we discuss it first before opening PR, to save your time)

PR #1996, opened 2025-04-24
Web Ref

Bashamega:
I have fixed the outdated the idl.


Bashamega:
Still working on it

saschanaz:
Once again, let's discuss first before writing PR, to save your time and also reviewers' time.

Bashamega:
> Once again, let's discuss first before writing PR, to save your time and also reviewers' time. Especially the change becomes big like this.

Sorry, for not discussing it first. It seems like the IDL folder in the current codebase is outdated. So the purpose of this pr is to update the IDL folder

saschanaz:
It's some remnant of the age when we manually downloaded IDL from each spec before webref was a thing. It's now only there because of comments, and I think it can mostly go away.

Bashamega:
> It's some remnant of the age when we manually downloaded IDL from each spec before webref was a thing. It's now only there because of comments, and I think it can mostly go away.

Oh, okay. Sorry for the misunderstanding.

saschanaz:
That's exactly what I said.

PR #1999, opened 2025-04-24
Expands CSS custom highlights API

### LINKED ISSUE #1698: CSS custom highlights API?
https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API

Supported in Chrome, Safari, and Firefox nightly.

### LINKED ISSUE #1978: The asynciterable files are not being added to the packages
> This is the unrelated problem I was talking about; we generate these files, but haven't included them in any package. That feels like a bug. But nothing to do with your PR.

_Originally posted by @jakebailey in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1965#discussion_r2042616216_

Bashamega:
I have added more methods to the `HighlightRegistry` interface according to the MDN docs.
fixes #1698

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -14939,21 +14939,21 @@ interface HighlightRegistry {
     /** [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/HighlightRegistry/clear) */
     clear: () => void;
     /** [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/HighlightRegistry/delete) */
-    delete: (id: string) => Promise<boolean>;
+    delete: (id: string) => boolean;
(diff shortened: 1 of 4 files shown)


jakebailey:
Why doesn't this get pulled in via some other types? (I am not personally knowledgeable about what has to be special cased or not)

saschanaz:
This exists already:

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/7ca5f399280c9065a4110834601d8a70391d6951/baselines/dom.iterable.generated.d.ts#L151-L152

Seems the types are now available for ES5, but I don't think it's worth the effort, one should just use ES2015+.

HolgerJeromin:
Can #1698 be closed then?

jakebailey:
Notably the iterable dom types aren't loaded by default; I personally wish we'd change that, but is this the only place it's declared?

PR #2000, opened 2025-04-25
chore: make KeyUsages a read only

### LINKED ISSUE #1929: SubtleCrypto#importKey should accept `readonly KeyUsage[]`
As of `@types/web@0.0.208`, **SubtleCrypto** is declared as:

```ts
interface SubtleCrypto {
    importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>;
    importKey(format: Exclude<KeyFormat, "jwk">, keyData: BufferSource, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm, extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKey>;
}
```

Notably, the `keyUsages` argument in the first overload accepts `readonly KeyUsage[]`, while the second overload requires a mutable array.

[Web Cryptography API](https://www.w3.org/TR/2017/REC-WebCryptoAPI-20170126/#SubtleCrypto-method-importKey) section 14.3.9 does not have any procedure that would mutate this argument.
Hence, this argument should accept `readonly` in both overloads.

The same problem also occurs in one of the overloads in `deriveKey` and `generateKey` and `unwrapKey` methods.

Bashamega:
fixes #1929

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -24267,7 +24267,7 @@ interface SubtleCrypto {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/SubtleCrypto/deriveBits) */
     deriveBits(algorithm: AlgorithmIdentifier | EcdhKeyDeriveParams | HkdfParams | Pbkdf2Params, baseKey: CryptoKey, length?: number | null): Promise<ArrayBuffer>;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/SubtleCrypto/deriveKey) */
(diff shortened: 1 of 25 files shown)


saschanaz:
If we want this we should do for all arrays. Not sure we want this though. Maybe @jakebailey has more strong opinion.z

(discussion before PR please... 😅)

Bashamega:
> If we want this we should do for all arrays. Not sure we want this though. Maybe @jakebailey has more strong opinion.
> 
> (discussion before PR please... 😅)

I thought it was just a problem with one type. I can make it for all if you want

saschanaz:
IDL reads from the array and get its own clone, so technically every `sequence<T>` for input can be ReadonlyArray. But only when it's used for input, because for output it's just an array that can do readwrite.

saschanaz:
Maybe try covering only function arguments, because for dictionaries it's super tricky to figure out whether it's input or output.

saschanaz:
And FYI there's already the logic to get input arrays to receive Iterable, so you should be able to reuse it

PR #2001, opened 2025-04-25
Enhance RTCDataChannel send method to accept multiple data types: string, Blob, ArrayBuffer, and ArrayBufferView

### LINKED ISSUE #1973: RTCDataChannel.send fails to compile
# Code that should compile but doesn't

```ts
function sendIt (peerConnection: RTCPeerConnection, data: ArrayBufferView | string): void {
  const dc = peerConnection.createDataChannel('')
  dc.send(data)
}
```

# Reason

`RTCDataChannel.send` accepts `string | Blob | ArrayBuffer | ArrayBufferView` but because the `.send` method is declared as multiple overrides you have to work out the type of the argument before you invoke the method, even though you don't actually do anything to the data before sending:

```ts
// lib.dom.d.ts line 18,000
/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCDataChannel/send) */
send(data: string): void;
send(data: Blob): void;
send(data: ArrayBuffer): void;
send(data: ArrayBufferView): void;
```

```ts
function sendIt (peerConnection: RTCPeerConnection, data: ArrayBufferView | string): void {
  const dc = peerConnection.createDataChannel('')

  if (typeof data === 'string') {
    dc.send(data)
  } else {
    dc.send(data)
  }
}
```

Changing it to the following works:

```ts
/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCDataChannel/send) */
send(data: string | Blob | ArrayBuffer | ArrayBufferView) void;
```

# Reference

- https://www.w3.org/TR/webrtc/#webidl-rtcdatachannel
- https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/send

Bashamega:
fixes #1973 

I have updated it, and I have added some tests

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -19799,7 +19799,7 @@ interface RTCDataChannel extends EventTarget {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCDataChannel/close) */
     close(): void;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/RTCDataChannel/send) */
-    send(data: string): void;
+    send(data: string | Blob | ArrayBuffer | ArrayBufferView): void;
     send(data: Blob): void;
     send(data: ArrayBuffer): void;
     send(data: ArrayBufferView<ArrayBuffer>): void;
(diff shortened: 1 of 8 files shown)


saschanaz:
We don't really add this kind of workaround here, I think the longstanding https://github.com/Microsoft/TypeScript/issues/14107 is the real culprit.

Bashamega:
> We don't really add this kind of workaround here, I think the longstanding [microsoft/TypeScript#14107](https://github.com/microsoft/TypeScript/issues/14107) is the real culprit.

Okay

PR #2012, opened 2025-05-01
Emit Iterable only for non-return array types

### LINKED ISSUE #1736: dom.iterable.d.ts incorrectly converts `sequence<>` return types to `Iterable<>`
This conversion only makes sense for parameters but not for return types.

`getUniformIndices()` for WebGL is currently affected.

saschanaz:
Fixes #1736


Bashamega (approved):

jakebailey (approved):

saschanaz:
LGTM

PR #2027, opened 2025-05-23
fix: return empty string for missing 'type' field in emitWebIdl function

Bashamega:
related to #1993 

Modified the function to exit silently instead of throwing an error.

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -361,8 +361,7 @@ export function emitWebIdl(
     if (obj.overrideType) {
       return obj.nullable ? makeNullable(obj.overrideType) : obj.overrideType;
     }
-    if (!obj.type)
-      throw new Error("Missing 'type' field in " + JSON.stringify(obj));
+    if (!obj.type) return "";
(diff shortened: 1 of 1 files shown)


Bashamega:
Hello @saschanaz 
Do you have time to review this?
Thank you

saschanaz:
I don't think this is what we want, any broken object without `type` would then emit broken TypeScript without type. As I commented there it would be nicer to have a conditional merger function, that merges only when the object exists. (`merge()` merges every field which makes sense for addedTypes, not much for this case. Perhaps add a flag in `merge()` somehow?)

Bashamega:
> I don't think this is what we want, any broken object without `type` would then emit broken TypeScript without type. As I commented there it would be nicer to have a conditional merger function, that merges only when the object exists. (`merge()` merges every field which makes sense for addedTypes, not much for this case. Perhaps add a flag in `merge()` somehow?)

For example, add a flag to ignore errors?

saschanaz:
To do a conditional merge. Errors should not be ignored.

PR #2044, opened 2025-06-16
chore: correct MDN link for requestAnimationFrame and cancelAnimation…

ice909:
…Frame

The previous MDN reference for requestAnimationFrame was outdated/incorrect. This commit updates it to the current valid MDN URL: https://developer.mozilla.org/docs/Web/API/window/requestAnimationFrame

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -3111,9 +3111,9 @@ declare var AnimationEvent: {
 };
 
 interface AnimationFrameProvider {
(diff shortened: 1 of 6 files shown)


ice909:

saschanaz:
This is tricky, to fix this we need a way to make the comments per-global, otherwise the link becomes incorrect for workers.

For now I filed https://github.com/mdn/content/issues/39951 to mitigate this a bit.

ice909:
@saschanaz So do I still need to handle this pr?

PR #2049, opened 2025-06-21
Refactor property name handling in emitWebIdl function to support hyphenated names

Bashamega:
Solves the failing build of #2047

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -862,7 +862,6 @@ export function emitWebIdl(
       printer.printLine("/** @deprecated */");
       printer.printLine("declare const name: void;");
     } else {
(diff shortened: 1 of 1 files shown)


src/build/emitter.ts
```diff
@@ -862,6 +862,7 @@ export function emitWebIdl(
       printer.printLine("/** @deprecated */");
       printer.printLine("declare const name: void;");
     } else {
+      const propertyName = p.name.includes("-") ? `"${p.name}"` : p.name;
```
saschanaz: Can we move this before optionalModifier so that it's closer to where it's used?

Bashamega: Done @saschanaz

saschanaz:
LGTM

jakebailey (approved):

PR #2050, opened 2025-06-23
feat: completely factorize SubtleCrypto.generateKey()

hkleungai:
Opening this PR in favor of [this comment from @Renegade334](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73066#issuecomment-2989865602) in my another PR in DT repo.

Since https://nodejs.org/api/webcrypto.html#cryptokeyusages lists the usages in its own table, and it is closedly aligned to the MDN reference in https://developer.mozilla.org/docs/Web/API/SubtleCrypto/generateKey, I believe it is possible to raise such change to `@types/web` as well.

Potentially we can do this to other methods in the key-usages table, but let's see how reviewers think about this change, then we may perform the refactoring incrementally in subsequent PRs.

Fallback signature is also updated, in favor of matching richer shape of possible `Algorithm` object.


inputfiles/overridingTypes.jsonc
```diff
@@ -1785,15 +1785,30 @@
...
+                                    "param": [
+                                        {
+                                            "name": "algorithm",
+                                            "overrideType": "string | T"
+                                        },
+                                    ],
                                     "overrideType": "Promise<CryptoKeyPair | CryptoKey>"
                                 }
-                            }
+                            },
```
Bashamega: Unneeded change

hkleungai: Removed.

inputfiles/overridingTypes.jsonc
```diff
@@ -1785,12 +1785,27 @@
...
+                                            "name": "T",
+                                            "extends": "{ name: string }"
+                                        }
+                                    ],
+                                    "param": [
+                                        {
+                                            "name": "algorithm",
+                                            "overrideType": "string | T"
+                                        }
+                                    ],
```
Renegade334: Unnecessary generic?

hkleungai: I believe this generic is needed, or else the fallback signature will fail to capture poosibly-work `algorithm` object with extra attibutes other than `name`.

See https://tsplay.dev/N5oDdW.

Renegade334: No-one's raised any issues with `AlgorithmIdentifier`; this seems an arbitrary modification.

hkleungai: See https://tsplay.dev/Ndp9MN, in which the method call is taken from `types/node/test/globals-non-dom.ts` in DefinitelyTyped. 

Without the use of generic, mismatched `algorithm` and `keyUsages[]` params would yield a confusing error, saying the code is violating the wrong overload.

But with the generic, such mismatch will pass and give `Promise<CryptoKeyPair | CryptoKey>`. It may not be the best way to typecheck, but I think this generic could help prevent certain possible breaking change.

Renegade334: I don't really understand – the whole point of this PR is that it breaks in this case?

The example provided is a mismatch with the accepted usages of the new HmacKeyGenParams signature, and is genuinely incorrect usage that would error at runtime. If we're going to provide a route for the new keyUsages constraints to be ignored anyway, then there's surely there's no real advantage to these changes?

(We can always modify that test – I suspect that the selection of keyUsages there was entirely arbitrary.)

hkleungai: I think you are right. I have repushed the PR to remove the generics.

saschanaz:
Ok, enough procrastinating to review this...

I think this PR is incorrectly based on Node.js docs while it should be based on Web Cryptography spec. Node.js had some freedom to add more algorithms than Web Cryptography has.

Would be nice if this can be limited to what's listed in [the spec](https://w3c.github.io/webcrypto/).

inputfiles/overridingTypes.jsonc
```diff
@@ -1803,15 +1785,13 @@
                         "generateKey": {
                             "name": "generateKey",
                             "additionalSignatures": [
-                                "generateKey(algorithm: \"Ed25519\" | { name: \"Ed25519\" }, extractable: boolean, keyUsages: ReadonlyArray<\"sign\" | \"verify\">): Promise<CryptoKeyPair>",
-                                "generateKey(algorithm: RsaHashedKeyGenParams | EcKeyGenParams, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKeyPair>",
-                                "generateKey(algorithm: AesKeyGenParams | HmacKeyGenParams | Pbkdf2Params, extractable: boolean, keyUsages: ReadonlyArray<KeyUsage>): Promise<CryptoKey>"
+                                "generateKey(algorithm: RsaHashedKeyGenParams & { name: \"RSA-OAEP\" }, extractable: boolean, keyUsages: readonly (\"encrypt\" | \"decrypt\" | \"wrapKey\" | \"unwrapKey\")[]): Promise<CryptoKeyPair>",
```
Bashamega, line 1788: https://w3c.github.io/webcrypto/#rsa-oaep

inputfiles/overridingTypes.jsonc
```diff
@@ -1803,15 +1785,13 @@
...
+                                "generateKey(algorithm: (RsaHashedKeyGenParams & { name: \"RSA-PSS\" | \"RSASSA-PKCS1-v1_5\" }) | EcKeyGenParams | \"Ed25519\" | { name: \"Ed25519\" } | \"Ed448\" | { name: \"Ed448\" }, extractable: boolean, keyUsages: readonly (\"sign\" | \"verify\")[]): Promise<CryptoKeyPair>",
```
Bashamega, line 1789: https://w3c.github.io/webcrypto/#rsa-pss
https://w3c.github.io/webcrypto/#rsassa-pkcs1

inputfiles/overridingTypes.jsonc
```diff
@@ -1803,15 +1785,13 @@
...
+                                "generateKey(algorithm: \"X25519\" | { name: \"X25519\" } | \"X448\" | { name: \"X448\" }, extractable: boolean, keyUsages: readonly (\"deriveKey\" | \"deriveBits\")[]): Promise<CryptoKeyPair>",
```
Bashamega, line 1790: https://w3c.github.io/webcrypto/#x25519

Bashamega, line 1790:
```suggestion
                                "generateKey(algorithm: \"X25519\" | { name: \"X25519\" }, extractable: boolean, keyUsages: readonly (\"deriveKey\" | \"deriveBits\")[]): Promise<CryptoKeyPair>",
```

There is no mention of X448 in the specs

inputfiles/overridingTypes.jsonc
```diff
@@ -1803,15 +1785,13 @@
...
+                                "generateKey(algorithm: AesKeyGenParams & { name: \"AES-CBC\" | \"AES-CTR\" | \"AES-GCM\" }, extractable: boolean, keyUsages: readonly (\"encrypt\" | \"decrypt\" | \"wrapKey\" | \"unwrapKey\")[]): Promise<CryptoKey>",
```
Bashamega, line 1791: https://w3c.github.io/webcrypto/#aes-cbc
https://w3c.github.io/webcrypto/#aes-ctr
https://w3c.github.io/webcrypto/#aes-gcm

inputfiles/overridingTypes.jsonc
```diff
@@ -1803,15 +1785,13 @@
...
+                                "generateKey(algorithm: AesKeyGenParams & { name: \"AES-KW\" }, extractable: boolean, keyUsages: readonly (\"wrapKey\" | \"unwrapKey\")[]): Promise<CryptoKey>",
```
Bashamega, line 1792: https://w3c.github.io/webcrypto/#aes-kw

baselines/webworker.generated.d.ts
```diff
@@ -8403,10 +8409,13 @@ interface SubtleCrypto {
...
-    generateKey(algorithm: AlgorithmIdentifier, extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKeyPair | CryptoKey>;
+    generateKey(algorithm: RsaHashedKeyGenParams & { name: "RSA-OAEP" }, extractable: boolean, keyUsages: readonly ("encrypt" | "decrypt" | "wrapKey" | "unwrapKey")[]): Promise<CryptoKeyPair>;
+    generateKey(algorithm: (RsaHashedKeyGenParams & { name: "RSA-PSS" | "RSASSA-PKCS1-v1_5" }) | EcKeyGenParams | "Ed25519" | { name: "Ed25519" } | "Ed448" | { name: "Ed448" }, extractable: boolean, keyUsages: readonly ("sign" | "verify")[]): Promise<CryptoKeyPair>;
+    generateKey(algorithm: "X25519" | { name: "X25519" } | "X448" | { name: "X448" }, extractable: boolean, keyUsages: readonly ("deriveKey" | "deriveBits")[]): Promise<CryptoKeyPair>;
+    generateKey(algorithm: AesKeyGenParams & { name: "AES-CBC" | "AES-CTR" | "AES-GCM" }, extractable: boolean, keyUsages: readonly ("encrypt" | "decrypt" | "wrapKey" | "unwrapKey")[]): Promise<CryptoKey>;
+    generateKey(algorithm: AesKeyGenParams & { name: "AES-KW" }, extractable: boolean, keyUsages: readonly ("wrapKey" | "unwrapKey")[]): Promise<CryptoKey>;
```
Bashamega, line 8416: ~~I am not sure if you want to use `AesKeyGenParams`, it is not a type
https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/fbc37600c4f811a4b1ad56431a6a59d81c6161a8/baselines/webworker.generated.d.ts#L33C1-L35C2~~

Bashamega (commented):
I reviewed this quickly and added links to the relevant specs. It might be useful to also include those links as comments in the JSONC file for future reference.

I added one question as well, and left a comment about something that doesn’t appear to be covered by the spec.

I’m not familiar with cryptography, so I’d prefer not to make further changes until @saschanaz has reviewed and approved these updates.

Note: It would be preferable to migrate this to KDL now, since we’ll need to do so eventually. If you’re interested in handling the migration, you can refer to the WebGL KDL file for syntax guidance:
https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/main/inputfiles/patches/webgl.kdl

The SubtleCrypto migration should be largely similar to the WebGL one.

hkleungai:
Thank you for your feedback.

It has been a while since I worked on this PR... I will need some time (maybe over the weekend) to digest the comments and the latest dev on this repo first. :)

baselines/webworker.generated.d.ts
saschanaz: Can you elaborate? How is it not a type?

Bashamega: Never mind, this is correct.

PR #2051, opened 2025-06-28
Add CSS properties from BCD 6.0.24

saschanaz:
But disable others, there will be followups.

Notification.badge is removed by https://github.com/mdn/browser-compat-data/pull/27059.


jakebailey (approved):

saschanaz:
LGTM

Issue #2053, opened 2025-06-28
Split addedTypes/overridingTypes/removedTypes by topics


saschanaz:
I don't think monolithic files help much. Smaller topic-by-topic files would help understanding why they exist, e.g. Streams customization can go into "streams" topic.

Given this will add new files, I'd like to try [KDL](https://kdl.dev) for new ones than having to use JSON.

Bashamega:
Hello @saschanaz 
KDL seems cool. Can I try to migrate it, or this is still for discussion?

saschanaz:
I was planning to do this only when we have to add a new set up customization, but given we already have PRs like #2050 maybe we should try this before that, so that we can tell others to use it.

So yes, 👍🏻 from me

saschanaz:
Why KDL - it's mature enough (has been there for years), recently embraced by bikeshed (the web spec generator that over half of the specs depend on), has types so that we can do this:

```kdl
interface Foo {
  method bar returns=DOMString {
    param p1 type=DOMPoint
  }
  property baz returns=short
}
```

(Our input data was migrated from XML which is why it's still XML-ish with `.properties.property`, I think KDL fits better for that.

PR #2056, opened 2025-07-01
Switch from JSON to KDL format

Bashamega:
#2053

### DIFF
--- a/inputfiles/addedTypes.kdl
+++ b/inputfiles/addedTypes.kdl
@@ -1,0 +1,1650 @@
+mixins
+  mixin
+    AbstractWorker
+      events
+        event
+          // See the full list of possible autofill values for the `autocomplete` attribute:
+          // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#determine-a-field's-category
+          // Full spec at https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill.
+          name "error"
+          type "ErrorEvent"
+    DocumentAndElementEventHandlers
+      events
+        event
+          // See the full list of possible autofill values for the `autocomplete` attribute:
+          // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#determine-a-field's-category
+          // Full spec at https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill.
+          name "copy"
+          type "ClipboardEvent"
+          // See the full list of possible autofill values for the `autocomplete` attribute:
+          // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#determine-a-field's-category
+          // Full spec at https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill.
+          name "cut"
+          type "ClipboardEvent"
+          // See the full list of possible autofill values for the `autocomplete` attribute:
+          // https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#determine-a-field's-category
+          // Full spec at https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill.
+          name "paste"
+          type "ClipboardEvent"
+    DocumentOrShadowRoot
+      methods
+        method
+          elementFromPoint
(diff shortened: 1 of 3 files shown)


saschanaz:
Err, I was saying that I'd like to add a new topic-separated files in KDL, not converting existing monolithic file to KDL.

inputfiles/addedTypes.kdl
```diff
@@ -0,0 +1,1650 @@
+mixins
```
jakebailey: Is this formatting a mistake? This is pretty unreadable, way worse than described in the linked issue with the braces and so on

Bashamega:
I understand @saschanaz 
I am just converting the data for now, and trying to see if it works, when It works I will split them into sub files

saschanaz:
Sounds good, but let's do it one by one in each PR so that the PR can be reviewed and merged separately!

saschanaz:
(I think we don't need to keep quirks like `methods.method` in KDL, we can skip that in KDL and convert to the existing JSON format (we'll need such conversion as KDL parse result itself will never be compatible as-is.)

Bashamega:
Hello @saschanaz 
Currently, I am working on the read function.
And I found an issue with KDL that I can't find a fix for.
Some arrays only have one item, so KDL doesn't treat it as an array. 
How can I make it an array?
I can only come up with one solution, which is when I want an array with one item and another item called, for example, `KDLARRAY`, and we will drop it when converting.
If you have a better solution, please tell me

saschanaz:
What arrays are we talking about and what exact result are you getting?

Bashamega:
Like this @saschanaz 
https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/cfe1f6a0ff63b068f48d4325b2a4b7858ad5a6ea/inputfiles/removedTypes.jsonc#L8
We are getting:
```json
{
	"param": "options"
}

**Comment by saschanaz** [comment]
But that's not KDL input?

**Comment by saschanaz** [comment]
(But I think we can defer the removedTypes part, it just has different form)

**Comment by Bashamega** [comment]
> But that's not KDL input?

The input is:
```
{
	"param": ["options"]
}
```
The issue is that KDL doesn't understand arrays for one item very well

**Comment by saschanaz** [comment]
That still doesn't seem like a valid KDL input? Parsing it causes errors.

Again let's skip removedItems for now.

**Comment by Bashamega** [comment]
There is the same issue with added types.
KDL:
```
AbstractWorker {
      events {
        event {
          name "error"
          type "ErrorEvent"
        }
      }
    }
```
Original JSON:
```
"AbstractWorker": {
                "events": {
                    "event": [
                        {
                            "name": "error",
                            "type": "ErrorEvent"
                        }
                    ]
                }
            },
```
KDL to json:
```
AbstractWorker: { events: { event: { name: 'error', type: 'ErrorEvent' } } },
```

I could only find on solution, which is using an extra item that is ignored when compiling back the json.
I hope the issues is clear now @saschanaz

**Comment by saschanaz** [comment]
I think that's a quirk that I don't really want to copy into KDL.

```kdl
interface-mixin AbstractWorker {
  event {
    name "error"
    type "ErrorEvent"
  }
}
```

This should be perfectly fine.

saschanaz:
Or go even further.

```kdl
interface-mixin AbstractWorker {
  event error type="ErrorEvent"
}
```

Bashamega:
Okay

Bashamega:
I have converted a portion of the file to the format you suggested @saschanaz 
Do you think this is the correct format, and I should convert the rest to this?

Bashamega:
Also do you think this is more readable @jakebailey

saschanaz:
Mostly looks good with a few comments:

* overrideSignatures should probably a member instead of a property
* interface-enum should be just enum

> I should convert the rest to this?

If you mean the other object types like interface and namespace, then yes.

I think the migration should happen only progressively with topics rather than converting the whole file at once, that's explicitly not a goal.

Bashamega:
I have made sure everything is working in this pr.
I will split it to smaller chunks now.

PR #2064, opened 2025-07-06
KDL Parser

Bashamega:
This PR is a small part of #2053.

It introduces the **kdljs** dependency and adds a parser utility that converts KDL text into the specific JSON structure we want.

## Changes

* Added **kdljs** to dependencies.
* Created a parser function that takes KDL input and returns our expected JSON format.

## Motivation

Part of the work to address #2053, where we need to handle KDL-formatted data in our system.

## Notes

* This is an initial step to enable downstream processing of KDL.
* Follow-up PRs will build on this foundation.

### DIFF
--- a/src/build/utils/kdl.ts
+++ b/src/build/utils/kdl.ts
@@ -1,0 +1,112 @@
+import { parse } from "kdljs";
+import { Enum, Event, Method, Property } from "../types";
+import { readFile } from "fs/promises";
+
+interface MethodDescriptor {
+  [x: string]: Omit<Method, "signature">;
+}
+
+interface EnumDescriptor {
+  [x: string]: Enum;
+}
+interface PropertyDescriptor {
+  [x: string]: Omit<Property, "type">;
+}
+
+interface InterfaceMixin {
+  [x: string]: {
+    events?: { event: Event[] };
+    methods?: { method: MethodDescriptor };
+  };
+}
+
+/**
+ * Converts KDL text describing interface-mixins into JSON format.
+ * Uses kdljs v3.
+ */
+export function parseKDL(kdlText: string) {
+  const { output, errors } = parse(kdlText);
+
+  if (errors.length) {
+    throw new Error(
+      `KDL parse errors:\n${errors.map((e) => e.message).join("\n")}`,
+    );
+  }
+
+  const nodes = output!;
+  const enums: EnumDescriptor = {};
+  const interfaces: InterfaceMixin = {};
+
+  for (const node of nodes) {
+    if (node.name === "enum") {
+      // Handle enum
+      const enumName = node.values?.[0]?.toString() ?? "";
+      let name = enumName;
+      const values: string[] = [];
+
+      for (const child of node.children ?? []) {
+        if (
+          child.name === "value" &&
+          child.values &&
+          child.values[0] !== undefined
+        ) {
+          values.push(child.values[0]!.toString());
+        } else if (child.name === "name") {
+          name = child.values?.[0]?.toString() ?? enumName;
+        }
+      }
+
+      enums[enumName] = { name, value: values };
+    } else {
+      // Handle interface-mixin
+      const name = node.values?.[0]?.toString() ?? "";
+      const iface: {
+        events?: { event: Event[] };
+        methods?: { method: MethodDescriptor };
+        properties?: { property: PropertyDescriptor };
+      } = {};
+
+      const events: Event[] = [];
+      const methods: MethodDescriptor = {};
+      const properties: PropertyDescriptor = {};
+
+      for (const child of node.children ?? []) {
+        const name = child.values?.[0]?.toString() ?? "";
+        if (child.name === "event") {
+          events.push({
+            name,
+            type: child.properties?.type?.toString() ?? "",
+          });
+        } else if (child.name === "method") {
+          methods[name] = {
+            name,
+            overrideSignatures: [
+              child.properties?.overrideSignatures?.toString() ?? "",
+            ],
+          };
+        } else if (child.name === "property") {
+          properties[name] = {
+            name,
+            exposed: child.properties?.exposed?.toString(),
+          };
+        }
+      }
+
+      if (events.length) iface.events = { event: events };
+      if (Object.keys(methods).length) iface.methods = { method: methods };
+      if (Object.keys(properties).length)
+        iface.properties = { property: properties };
+
+      interfaces[name] = iface;
+    }
+  }
+
+  return { mixins: { mixin: interfaces }, enums: { enum: enums } };
+}
+export default async function readInputKDL(
+  path: string,
+  inputFolder: string,
+): Promise<any> {
+  const text = await readFile(new URL(path, inputFolder), "utf8");
+  return parseKDL(text);
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "eslint-plugin-prettier": "^5.1.3",
         "globals": "^16.0.0",
         "jsonc-parser": "^3.2.1",
+        "kdljs": "^0.3.0",
         "node-fetch": "^3.3.2",
         "prettier": "^3.2.5",
         "print-diff": "^2.0.0",
@@ -45,6 +46,50 @@
       "engines": {
         "node": ">=0.10.0"
       }
+    },
+    "node_modules/@chevrotain/cst-dts-gen": {
+      "version": "11.0.3",
+      "resolved": "https://registry.npmjs.org/@chevrotain/cst-dts-gen/-/cst-dts-gen-11.0.3.tgz",
+      "integrity": "sha512-BvIKpRLeS/8UbfxXxgC33xOumsacaeCKAjAeLyOn7Pcp95HiRbrpl14S+9vaZLolnbssPIUuiUd8IvgkRyt6NQ==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@chevrotain/gast": "11.0.3",
+        "@chevrotain/types": "11.0.3",
+        "lodash-es": "4.17.21"
+      }
+    },
+    "node_modules/@chevrotain/gast": {
+      "version": "11.0.3",
+      "resolved": "https://registry.npmjs.org/@chevrotain/gast/-/gast-11.0.3.tgz",
+      "integrity": "sha512-+qNfcoNk70PyS/uxmj3li5NiECO+2YKZZQMbmjTqRI3Qchu8Hig/Q9vgkHpI3alNjr7M+a2St5pw5w5F6NL5/Q==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@chevrotain/types": "11.0.3",
+        "lodash-es": "4.17.21"
+      }
+    },
+    "node_modules/@chevrotain/regexp-to-ast": {
+      "version": "11.0.3",
+      "resolved": "https://registry.npmjs.org/@chevrotain/regexp-to-ast/-/regexp-to-ast-11.0.3.tgz",
+      "integrity": "sha512-1fMHaBZxLFvWI067AVbGJav1eRY7N8DDvYCTwGBiE/ytKBgP8azTdgyrKyWZ9Mfh09eHWb5PgTSO8wi7U824RA==",
+      "dev": true,
+      "license": "Apache-2.0"
+    },
+    "node_modules/@chevrotain/types": {
+      "version": "11.0.3",
+      "resolved": "https://registry.npmjs.org/@chevrotain/types/-/types-11.0.3.tgz",
+      "integrity": "sha512-gsiM3G8b58kZC2HaWR50gu6Y1440cHiJ+i3JUvcp/35JchYejb2+5MVeJK0iKThYpAa/P2PYFV4hoi44HD+aHQ==",
+      "dev": true,
+      "license": "Apache-2.0"
+    },
+    "node_modules/@chevrotain/utils": {
+      "version": "11.0.3",
+      "resolved": "https://registry.npmjs.org/@chevrotain/utils/-/utils-11.0.3.tgz",
+      "integrity": "sha512-YslZMgtJUyuMbZ+aKvfF3x1f5liK4mWNxghFRv7jqRR9C3R3fAOGTTKvxXDa2Y1s9zSbcpuO0cAxDYsc9SrXoQ==",
+      "dev": true,
+      "license": "Apache-2.0"
     },
     "node_modules/@eslint-community/eslint-utils": {
       "version": "4.7.0",
@@ -1197,6 +1242,21 @@
       },
       "engines": {
         "node": ">=4"
+      }
+    },
+    "node_modules/chevrotain": {
+      "version": "11.0.3",
+      "resolved": "https://registry.npmjs.org/chevrotain/-/chevrotain-11.0.3.tgz",
+      "integrity": "sha512-ci2iJH6LeIkvP9eJW6gpueU8cnZhv85ELY8w8WiFtNjMHA5ad6pQLaJo9mEly/9qUyCpvqX8/POVUTf18/HFdw==",
+      "dev": true,
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@chevrotain/cst-dts-gen": "11.0.3",
+        "@chevrotain/gast": "11.0.3",
+        "@chevrotain/regexp-to-ast": "11.0.3",
+        "@chevrotain/types": "11.0.3",
+        "@chevrotain/utils": "11.0.3",
+        "lodash-es": "4.17.21"
       }
     },
     "node_modules/color-convert": {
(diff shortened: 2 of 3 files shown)


saschanaz:
I think we'll need a minimal amount of migration in this PR, to confirm the code works 👀

Bashamega:
> I think we'll need a minimal amount of migration in this PR, to confirm the code works 👀

It is just the same code as in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2056
I have just cleaned it up and made it smaller

saschanaz:
That's in a different PR. It should be in the same PR so that it's easy to see the CI works on the PR, as the code can change during the review pingpong.

Bashamega:
> That's in a different PR. It should be in the same PR so that it's easy to see the CI works on the PR, as the code can change during the review pingpong.

I will add it today

Bashamega:
I have updated it @saschanaz 
I have migrated AutFill.
I also made a function that would read all KDL files, so we don't need the change the build files when adding new topics.

inputfiles/AutoFill/addedTypes.kdl
```diff
@@ -0,0 +1,70 @@
+enum AutoFillBase {
```
saschanaz: Maybe `inputfiles/patches/autofill.kdl`?

1. "patches" because we are essentially patching over Web IDL for better typing.
2. This way `getAllKDLFileURLs` can collect all files there without filter nor recursion.

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
...
+          child.values[0] !== undefined
+        ) {
+          values.push(child.values[0]!.toString());
+        } else if (child.name === "name") {
+          name = child.values?.[0]?.toString() ?? enumName;
+        }
+      }
+
+      enums[enumName] = { name, value: values };
+    } else {
```
saschanaz: We only have enums here, let's only include enum processing and add anything else in separate PRs. Meaning this else branch can go for now.

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
+import { parse } from "kdljs";
```
saschanaz: `build/patches.ts`?

1. It's a module used by build.ts and nothing else, so no need to go `build/utils`. That's mostly for util functions used by various modules in `build/`.
2. We do use KDL but I think describing the purpose of the module would be better for the name. As described above, the files are for patching over IDL, so maybe `patches`.

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  for (const entry of entries) {
+    const child = new URL(entry.name + "/", folder);
+
+    if (
+      entry.isDirectory() &&
+      !child.pathname.includes("mdn") &&
+      !child.pathname.includes("idl")
+    ) {
+      results.push(...(await getAllKDLFileURLs(child, file)));
```
saschanaz: (This branch can go if we put patch files in `inputfiles/patches/`)

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  const nodes = output!;
+  const enums: EnumDescriptor = {};
+  const interfaces: InterfaceMixin = {};
+
+  for (const node of nodes) {
+    if (node.name === "enum") {
+      // Handle enum
+      const enumName = node.values?.[0]?.toString() ?? "";
```
saschanaz: Missing name should throw an error.

```suggestion
      const enumName = node.values[0];
      if (typeof enumName !== "string") {
        throw new Error("Missing enum name");
      }
```

saschanaz: This branch should be extracted to a new function, something like `parseEnum`.

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
...
+ * Converts KDL text describing interface-mixins into JSON format.
+ * Uses kdljs v3.
+ */
+export function parseKDL(kdlText: string) {
+  const { output, errors } = parse(kdlText);
+
+  if (errors.length) {
+    throw new Error(
+      `KDL parse errors:\n${errors.map((e) => e.message).join("\n")}`,
+    );
```
saschanaz: Given `errors` includes Error objects, let's use `cause` so that Node.js can show it with proper terminal highlights.

```suggestion
    throw new Error("KDL parse errors", { cause: errors });
```

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
...
+interface InterfaceMixin {
+  [x: string]: {
+    events?: { event: Event[] };
+    methods?: { method: MethodDescriptor };
+  };
+}
+
+/**
+ * Converts KDL text describing interface-mixins into JSON format.
+ * Uses kdljs v3.
```
saschanaz:
```suggestion
 * Converts patch files in KDL to match the [types](types.d.ts).
```

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
...
+      for (const child of node.children ?? []) {
+        if (
+          child.name === "value" &&
+          child.values &&
+          child.values[0] !== undefined
+        ) {
+          values.push(child.values[0]!.toString());
+        } else if (child.name === "name") {
+          name = child.values?.[0]?.toString() ?? enumName;
+        }
```
saschanaz:
```suggestion
        if (child.name !== "value" || typeof child.values[0] !== "string") {
          throw new Error("enum values should be in the form of `value {name}`");
        }
        values.push(child.values[0]);
```

1. We should throw when the items are not in the expected form.
2. I see why `child.name === "name"` branch exists... but let's exclude that here, given that 1) the added KDL file doesn't use this branch yet, and 2) I think it should be a property instead of a child node (maybe `override-name=(newname)`).

src/build/utils/kdl.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+export default async function readKDL(
+  inputFolder: URL,
+  file: string,
+): Promise<any> {
+  const fileUrls = await getAllKDLFileURLs(inputFolder, file);
+
+  const parsedContents = await Promise.all(fileUrls.map(readInputKDL));
+
+  return parsedContents.reduce((acc, current) => merge(acc, current), {});
```
saschanaz:
```suggestion
  return parsedContents.reduce(merge, {});
```

saschanaz (changes requested):

src/build/utils/kdl.ts
Bashamega: I had to revert it because typescript wasn't happy

src/build/utils/kdl.ts
Bashamega: I have updated it

src/build/utils/kdl.ts
Bashamega: Done

Bashamega:
I have updated it @saschanaz

inputfiles/AutoFill/addedTypes.kdl
saschanaz: This still applies although we got inputfiles/patches - I don't think we need a subdirectory. For removedTypes I think we can either create a new directory `inputfiles/removals` or add a new node alongside the "normal" patching nodes:

```kdl
removal {
  enum Foo {
    value "should-be-removed"
  }
}
```

src/build/utils/kdl.ts
saschanaz: > This branch should be extracted to a new function, something like `parseEnum`.

This still applies

inputfiles/AutoFill/addedTypes.kdl
Bashamega: okay

Bashamega: I have updated it @saschanaz

src/build/patches.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export async function readInputKDL(fileUrl: URL): Promise<any> {
+  const text = await readFile(fileUrl, "utf8");
+  return parseKDL(text);
+}
+
+/**
+ * Read, parse, and merge all KDL files under the input folder.
+ */
+export default async function readKDL(folder: string): Promise<any> {
```
saschanaz: readPatches? Would be nice if the function names say "patch" instead, as I think "this function reads patch files" is more important information for the caller than "the patch is in KDL format".

Comments are fine though.

src/build/patches.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+async function getAllKDLFileURLs(folder: URL): Promise<URL[]> {
+  const entries = await readdir(folder, { withFileTypes: true });
+  return entries.map((entry) => new URL(entry.name, folder));
+}
+
+/**
+ * Read and parse a single KDL file.
+ */
+export async function readInputKDL(fileUrl: URL): Promise<any> {
```
saschanaz: And this can be readPatch. (You probably followed the existing readInputJSON, which is literally just parsing JSON and returned the result as-is, but in this case we also convert the data so it doesn't fit much.)

src/build/patches.ts
saschanaz: Also we probably don't need to receive `folder`, at least not in this PR as it's always "patches".

saschanaz (changes requested):
Mostly looks nice to me! Some smaller nits:

Bashamega:
Thank you very much for your review @saschanaz 
I have updated the PR.

src/build/patches.ts
```diff
@@ -0,0 +1,75 @@
...
+export async function readPatch(fileUrl: URL): Promise<any> {
+  const text = await readFile(fileUrl, "utf8");
+  return parseKDL(text);
+}
+
+/**
+ * Read, parse, and merge all KDL files under the input folder.
+ */
+export default async function readPatches(): Promise<any> {
+  const inputFolder = new URL("../../inputfiles/patches/", import.meta.url);
```
saschanaz: Patch directory? (folder is not very universal language across platforms)

src/build/patches.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      enums[enumName] = { name, value: values };
+    }
+  }
+
+  return { enums: { enum: enums } };
+}
+
+/**
+ * Recursively collect all KDL file URLs in a directory.
```
saschanaz: > Recursively 

This is not true anymore, right?

Also maybe KDL can go from the name, as it just collects all files now. (even though the caller expects KDL only)

src/build/patches.ts
```diff
@@ -0,0 +1,75 @@
...
+    throw new Error("KDL parse errors", { cause: errors });
+  }
+
+  const nodes = output!;
+  const enums: EnumDescriptor = {};
+
+  for (const node of nodes) {
+    if (node.name === "enum") {
+      // Handle enum
+      const enumName = node.values[0];
```
saschanaz: enumName and name coexist, maybe one of them can go for now.

src/build/patches.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export function parseKDL(kdlText: string) {
+  const { output, errors } = parse(kdlText);
+
+  if (errors.length) {
+    throw new Error("KDL parse errors", { cause: errors });
+  }
+
+  const nodes = output!;
+  const enums: EnumDescriptor = {};
```
```suggestion
  const enums: Record<string, Enum> = {};
```

Then the descriptor type can go.

src/build/patches.ts
```diff
@@ -39,15 +34,15 @@ export function parseKDL(kdlText: string) {
         values.push(child.values[0]);
       }
 
-      enums[enumName] = { name, value: values };
+      enums[enumName] = { name:enumName, value: values };
```
saschanaz: Should we say `name`? 🤔 (And later the overriding name can be named `overridingName`)

src/build/patches.ts
Bashamega: I have updated it @saschanaz
I was just trying to use the same terminology as in the rest of the project. If you would like I can open a separate PR updating the terminology

src/build/patches.ts
```diff
@@ -0,0 +1,70 @@
...
+    throw new Error("KDL parse errors", { cause: errors });
+  }
+
+  const nodes = output!;
+  const enums: Record<string, Enum> = {};
+
+  for (const node of nodes) {
+    if (node.name === "enum") {
+      // Handle enum
+      const name = node.values[0];
```
saschanaz: This was buried in older comments somehow, but would be nicer to extract this enum parsing code to a separate function, as we plan to add more things here. Otherwise parseKDL will become way too large to read.

src/build/patches.ts
Bashamega: Done 😃

src/build/patches.ts
Bashamega: Done @saschanaz

src/build/patches.ts
```diff
@@ -0,0 +1,77 @@
+import { parse } from "kdljs";
+import { Enum } from "./types";
```
```suggestion
import type { Enum } from "./types";
```

src/build/patches.ts
```diff
@@ -0,0 +1,77 @@
...
+import { readdir, readFile } from "fs/promises";
+import { merge } from "./helpers.js";
+
+/**
+ * Converts patch files in KDL to match the [types](types.d.ts).
+ */
+export function parseKDL(kdlText: string) {
```
```suggestion
function parseKDL(kdlText: string) {
```

(I don't think this function is used anywhere else)

saschanaz (approved):
Looks cool. Thanks for the work! Just some last nits:

Bashamega:
Your are welcome @saschanaz 

Thank you for taking the time to review this PR.

Bashamega:
I have migrated more enums.

Can you quickly check it again @saschanaz?
Thank you

saschanaz:
Err I thought we'll do that in a separate PR

saschanaz:
(And I know it's an existing convention to use camelCase in file names, for new files I prefer hyphenated names as done in src/build/)

saschanaz:
(But perhaps @jakebailey would want consistency with the core typescript repository which also uses camel case)

Bashamega:
> Err I thought we'll do that in a separate PR

If you want i can make it a separate PR.


> (And I know it's an existing convention to use camelCase in file names, for new files I prefer hyphenated names as done in src/build/)

But all the input files are camelCased. If you want, I can create another PR to convert everything to be hyphenated even though I prefer camelCased.

saschanaz:
> If you want i can make it a separate PR.

Let's do that

jakebailey:
> (But perhaps @jakebailey would want consistency with the core typescript repository which also uses camel case)

Of course, my personal preference is not to have dashes in filenames like this, no.

As for the PR as a whole, I'm not sure I understand the benefit of KDL specifically in this instance; can't the splitting be done without doing that? Pretty surprised by the repetition of the word "value" over and over.

saschanaz:
Technically we can skip `value`:

```kdl
enum Enum {
  "foo"
  "bar"
  baz // Actually even the quote can be skipped
  baz-baz // Because this is legal
}
```

But this way we won't be able to tell the child node type anymore, as we treat the type as random input string. I was thinking about a chance to add another child node, but perhaps that won't happen and we can do it simpler.

saschanaz:
> Of course, my personal preference is not to have dashes in filenames like this, no.

Ok... 🥺

Bashamega:
I have reverted it @saschanaz

Bashamega:
I have reverted it @saschanaz

saschanaz:
Should we try the shorter form mentioned in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2064#issuecomment-3058808364 ? (we can speed it up if you open another PR to my fork, if you want!)

Bashamega:
> Should we try the shorter form mentioned in [#2064 (comment)](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2064#issuecomment-3058808364) ? (we can speed it up if you open another PR to my fork, if you want!)

Sure, I will open another PR. Because I am very busy

saschanaz:
@jakebailey is it better now? 👀 (I wonder it's better to always quote or always not quote. It has no difference in the parse result though.)

saschanaz:
(LGTM from me)

Bashamega:
Hello @jakebailey 
Do you think the current format is better?

jakebailey:
Yeah, though I would like a quick vibe check from @sandersn.

sandersn (approved):

saschanaz:
Yay! LGTM

saschanaz:
Oopsy.

Bashamega:
Thank you for your reviews and merging @saschanaz, @jakebailey, and @sandersn

For the upcoming migration of JSON files to KDL, would you prefer submitting all changes in a single PR or splitting them into multiple PRs (e.g., one PR per KDL file)?

saschanaz:
Smaller PRs are preferable 👍

Meaning if it's reasonably small then it's okay to have multiple topics in a PR. But I think we first need to implement more things than enums, so let's do that first.

PR #2071, opened 2025-07-17
Remove .d.ts post-process in deploy+migrate scripts

sandersn:
It has not been needed for years but I never noticed because I forgot about these scripts and never used them.

This means that the types packages used for overriding the default DOM types have had AbortSignal.abort incorrectly disabled all this time, but it's rarely used (only a few repos per million at Google, at least), so nobody ever complained about it.

In theory we could need the post-processing capability in the future, but it's trivial to bring back or to write again. That's why I deleted it outright.

### DIFF
--- a/deploy/createTypesPackages.js
+++ b/deploy/createTypesPackages.js
@@ -134,7 +134,6 @@ const go = async () => {
     });
 
     prependAutoImports(pkg, packagePath);
-    postProcessDTSFiles(pkg, packagePath);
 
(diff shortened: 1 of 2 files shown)


jakebailey (approved):
My bad for using this script to import changes into 5.9

sandersn:
I think it is a good idea to have a script. I just forgot about it because I hand-wrote the `cp` commands each time.

PR #2074, opened 2025-07-19
Restore Trusted Types from BCD 6.0.24

saschanaz:


saschanaz:
Every feedback from #1246 applies here too.

saschanaz:
Work item - Trusted* types should be ignored from the getter. And those types need to get the nominal type hack too. (Why in 2025 we still have no official solution for this? 😞)

saschanaz:
> Trusted* types should be ignored from the getter.

This needs TS5.1+ for https://github.com/microsoft/TypeScript/pull/53417. Meaning anyone with pre-5.1 will not be able to assign trusted types without `any` hack.

Edit: well, given we only have 5.5/5.6 backward-compat types, this effectively means pre-5.5 users would get problems. (I really don't want to add more unless it's super duper required)

jakebailey:
Oh boy, I wonder how much this conflicts with the existing trusted types that people are using...

saschanaz:
Should certainly take a look at https://www.npmjs.com/package/@types/trusted-types?activeTab=code to reduce friction.

saschanaz:
Although the library's `private brand: true` effectively makes it exclusive to any other declaration AFAICT...

jakebailey:
> Although the library's `private brand: true` effectively makes it exclusive to any other declaration AFAICT...

Yeah, they are declared as classes, and that private brand means they are nominal. I think this is definitely something that needs to be rectified somehow...

PR #2075, opened 2025-07-19
Add `__brand` to make empty interfaces nominal

saschanaz:
Because who have thought that there are so many existing empty interfaces which are also used as parameter types. And thus this follows the pattern in https://www.typescriptlang.org/play/?#example/nominal-typing.

Preparing for #2074.

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -1165,10 +1165,12 @@ export function emitWebIdl(
       optionsType: string,
     ) {
       printer.printLine(
-        `${prefix}${addOrRemove}EventListener<K extends keyof ${iParent.name
+        `${prefix}${addOrRemove}EventListener<K extends keyof ${
+          iParent.name
         }EventMap>(type: K, listener: (this: ${nameWithForwardedTypes(
           i,
-        )}, ev: ${iParent.name
+        )}, ev: ${
+          iParent.name
         }EventMap[K]) => any, options?: boolean | ${optionsType}): void;`,
       );
     }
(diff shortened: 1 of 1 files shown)


saschanaz:
https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1246/files#r881972594 suggested emitting a class with a private member. While I'm all for emitting classes (yes please) but a private member would only raise the possibility to clash with other libraries. Introducing a public member would allow other libraries to match the brand.

src/build/emitter.ts
Bashamega, line 1: Maybe combine all the if statements into one if statement?

saschanaz: IMO complex expressions are better as a separate statement.

Bashamega (approved):

jakebailey:
These will end up being in completions and so on (I see people do like `" brand": any` to derank them); is this a pattern taken from the existing trusted types definition or something?

I don't know that we actually use type branding in any of the existing lib.d.ts files, so this would be a new thing I think.

saschanaz:
> These will end up being in completions and so on

This is only for empty interfaces so I think that part is less a problem, although putting a space is an interesting solution.

> is this a pattern taken from the existing trusted types definition or something?

https://www.typescriptlang.org/play/?#example/nominal-typing

As you saw in #2074 the most popular one (I don't think there's any other comparable library?) uses `private brand: true` 😞

jakebailey:
One thing I think we're going to have to try is to port this code into the main TS repo and run the extended suite.

Renegade334:
Perhaps a better approach would be to implement https://webidl.spec.whatwg.org/#ref-for-dfn-class-string, ie. resurrect the spirit of #1762 et al.?

This could only really be a TS 6.0+ change (for all lib targets to include es6's well-known symbols) but it seems better than introducing ghost properties across the entire library.

PR #2076, opened 2025-07-20
Restore GPU*Error from BCD 6.0.24

saschanaz:


jakebailey (approved):

saschanaz:
LGTM

kainino0x:
Hi, how did TypeScript (and BCD) end up with only this tiny subset of WebGPU? It creates a conflict with `@webgpu/types`, but since TypeScript doesn't define the rest of the API, it prevents using WebGPU at all.
https://github.com/gpuweb/types/issues/192

kainino0x:
https://github.com/mdn/browser-compat-data/pull/27372 is the PR that updated support for Firefox, but it added the whole API.

PR #2077, opened 2025-07-20
Fix incorrect extract of comments

saschanaz:
Apparently the regex was too greedy and ate more characters than it should.

Lesson learned again: Don't use complex regex until it's super duper needed. If you think it's super duper needed, think again.


jakebailey (approved):

Bashamega (approved):
Nice

saschanaz:
LGTM

PR #2078, opened 2025-07-20
Use plaintext part of the MDN template

saschanaz:
Depends on #2077.

### DIFF
--- a/src/build/mdn-comments.ts
+++ b/src/build/mdn-comments.ts
@@ -29,7 +29,7 @@ function extractSummary(markdown: string): string {
     // Keep link text but remove URLs
     .replace(/\[(.*?)\]\(.*?\)/g, "$1")
     .replace(/"/g, "'")
-    .trim()
+    .trim();
 
   // Extract the first sentence (ending in . ! or ?)
   const sentenceMatch = normalizedText.match(/(.*?[.!?])(?=\s|$)/);
--- a/src/build/mdn-comments.ts
+++ b/src/build/mdn-comments.ts
@@ -29,7 +29,7 @@ function extractSummary(markdown: string): string {
     // Keep link text but remove URLs
     .replace(/\[(.*?)\]\(.*?\)/g, "$1")
     .replace(/"/g, "'")
-    .trim();
+    .trim()
 
   // Extract the first sentence (ending in . ! or ?)
   const sentenceMatch = normalizedText.match(/(.*?[.!?])(?=\s|$)/);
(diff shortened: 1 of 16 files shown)


src/build/mdn-comments.ts
```diff
@@ -5,49 +5,67 @@
...
-      /\{\{\s*(?:event|jsxref|cssref|specname)\s*\|\s*([^}]+)\s*\}\}/gi,
-      "$1",
-    ) // Handle event, jsxref, cssref, etc.
-    .replace(/\{\{\s*([^}]+)\s*\}\}/g, (_, match) => `[MISSING: ${match}]`) // Catch any remaining unhandled templates
-    .replace(/\[(.*?)\]\(.*?\)/g, "$1") // Keep link text but remove URLs
-    .replace(/\s+/g, " ") // Normalize spaces
-    .replace(/\n\s*/g, "\n") // Ensure line breaks are preserved
+  const normalizedText = firstParagraph
+    // Extract first argument from multiple templates, handling escaped quotes & spaces
+    .replace(/\{\{ *(?:\w+)*([^}]*)\}\}/g, (_, arg) => {
```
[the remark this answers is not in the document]
saschanaz: I think the regex is now easy enough for CodeQL to start understanding and complaining 😛

We could replace `*` to something like `{0, 20}` but I don't think that's worth.

src/build/mdn-comments.ts
```diff
@@ -5,49 +5,67 @@ const subdirectories = [
...
+      }
+      const args = trimEnds(arg).split(",");
+      if (args.length > 2) {
+        // For whatever reason domxref gives the first argument when there are more than 3 parameters
+        return trimEnds(args[0]);
+      }
+      // For all other cases, the last argument is the plaintext
+      return trimEnds(args.at(-1)!);
+    })
+    // Keep link text but remove URLs
```
Bashamega, line 51: Technically you can add links in the comments. So why not include links? 
Just a suggestion

saschanaz: That part is from your PR 😛

We can do it, but not here.

Bashamega: > That part is from your PR 😛
> 
> We can do it, but not here.

Correct 😃
I will open another PR for it

Bashamega (approved):

jakebailey (approved):

saschanaz:
Looking it again, it should probably be wrapped as a code than plaintext.

saschanaz:
Superceded by #2171

PR #2082, opened 2025-07-25
Refactor mixin handling in KDL parser

Bashamega:
- Removed AbstractWorker event definitions from addedTypes.jsonc.
- Introduced abstractworker.kdl to define AbstractWorker mixin with error event.
- Updated patches.ts to handle mixin parsing and return mixins in the output.

Related #2053


inputfiles/patches/abstractworker.kdl
```diff
@@ -0,0 +1,3 @@
+mixin AbstractWorker event {
+    error ErrorEvent
```
saschanaz:
```suggestion
    event error type=ErrorEvent
```

inputfiles/patches/abstractworker.kdl
```diff
@@ -0,0 +1,3 @@
+mixin AbstractWorker event {
```
saschanaz:
```suggestion
interface-mixin AbstractWorker {
```

Bashamega:
Thank you @saschanaz 
Can you send me a snippet of how you want the KDL to look?

saschanaz:
Briefly something like in https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/2053#issuecomment-3018946812.

Could change during review in case something is off though.

```kdl
// This part is done, thank you 🎉
enum Foo {
  foo
  bar
}

interface Foo {
  // Event information should ideally be provided by MDN and @webref/events, this is for edge cases
  event show type=ShowEvent
  event close type=CloseEvent

  method fooOp returns=FooReturnType {
    param arg1 type=ArgType
  }
  property fooAttr type=AttrType
}

interface Foo2 {
  template T1 extends=FooFoo default=FooBar

  operation fooOp override-returns=TemplateType<T1>
}

dictionary Bar {
  template T

  member bar required=#true type=DOMString
  // but all types here are to override the IDL one, perhaps there could be a better word, like ts-type?
  member baz override-type=Foo2<T>
}

**Comment by Bashamega** [comment]
I have updated it @saschanaz.
I will do the rest in separate PRs

File: src/build/patches.ts [hunk 3]
```diff
@@ -45,6 +53,29 @@ function handleEnum(node: any, enums: Record<string, Enum>) {
...
 
+/**
+ * Handles a mixin node by extracting its name and associated events.
+ * Throws an error if the mixin name is missing.
+ * If the mixin node specifies "event" as its second value, it collects all child nodes as events,
+ * each with a name and type, and adds them to the mixins record under the mixin's name.
+ * @param node The mixin node to handle.
+ * @param mixins The record of mixins to update.
+ */
+export function handleMixin(node: any, mixins: Record<string, any>) {
```
```suggestion
function handleMixin(node: Node, mixins: Record<string, any>) {
```

(Something I missed last time)

inputfiles/patches/abstractworker.kdl
```diff
@@ -0,0 +1,3 @@
+interface-mixin AbstractWorker {
+    event error type=ErrorEvent
+}
```
saschanaz: Would be nice to have a final newline - if you use VSCode then "Files: Insert Final Newline" should do it for you.

inputfiles/patches/abstractworker.kdl
```diff
@@ -0,0 +1,3 @@
+interface-mixin AbstractWorker {
```
saschanaz: In this specific case, I'd prefer to put all the event things to a single file as `events.kdl`, so that I can see which events are not covered by automation and can ultimately remove this file.

In this PR just the filename can change, no addition is needed.

(Because all event data technically exists, we don't have a way to track that to assign automatically right now)

src/build/patches.ts
```diff
@@ -1,5 +1,5 @@
 import { parse } from "kdljs";
```
```suggestion
import { parse, type Node } from "kdljs";
```

saschanaz (commented):
A few nits and then this will be good to go:

Bashamega:
Thank you for your review @saschanaz 
I have updated it

saschanaz:
LGTM, thanks!

jakebailey (approved):

saschanaz:
LGTM

PR #2086, opened 2025-07-29
Add links to comments

Bashamega:
_Originally suggested by me in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2078#discussion_r2217972142_

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -257,7 +257,7 @@ declare var AbortSignal: {
 };
 
 /**
- * The **`AudioWorkletGlobalScope`** interface of the Web Audio API represents a global execution context for user-supplied code, which defines custom AudioWorkletProcessor-derived classes.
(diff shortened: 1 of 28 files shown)


baselines/audioworklet.generated.d.ts
```diff
@@ -889,7 +889,7 @@ declare var ReadableByteStreamController: {
 };
 
 /**
- * The `ReadableStream` interface of the Streams API represents a readable stream of byte data.
+ * The `ReadableStream` interface of the [Streams API](/en-US/docs/Web/API/Streams_API) represents a readable stream of byte data.
```
jakebailey: This URL is unrooted

Bashamega: Thanks. I will fix it

Bashamega: I have updated it @jakebailey

jakebailey:
Is this actually a good idea? It seems like a lot of text, and the docs already include a link.

Bashamega:
> Is this actually a good idea? It seems like a lot of text, and the docs already include a link.

I think people would benefit from it, maybe @saschanaz has an opinion

PR #2088, opened 2025-07-30
Handle Properties in kdl

Bashamega:
reference #2053 

I've added support for handling properties and refactored the mixin handling into smaller, more manageable functions.


src/build/patches.ts
```diff
@@ -66,14 +103,11 @@ function handleMixin(node: Node, mixins: Record<string, any>) {
...
-    (child: any) => child.name === "event",
-  );
-  const event: Event[] = rawEvents.map((child: any) => ({
-    name: child.values[0],
-    type: child.properties.type,
-  }));
-  mixins[name] = { name, events: { event } };
+
+  const event: Event[] = extractMixinEvents(node);
+  const property: Properties = extractMixinProperties(node);
```
saschanaz: I think we'd want to iterate on children rather than filtering, so that if there's an unknown child then it can throw. Just like how handleMixin is called from parseKDL.

Bashamega: Is it now better @saschanaz, I have updated it :)

src/build/patches.ts
```diff
@@ -66,14 +67,31 @@ function handleMixin(node: Node, mixins: Record<string, any>) {
...
+        event.push({
+          name,
+          type: child.properties.type as string,
+        });
+        break;
+      case "property":
+        property[name] = {
+          name,
+          exposed: child.properties?.exposed as string,
+        };
```
saschanaz: I think separate functions for each is better, as I'm sure this will become longer, even for property alone.

Bashamega: I have created smaller functions @saschanaz, would you prefer smaller files?

src/build/patches.ts
```diff
@@ -94,6 +88,30 @@ function handleMixin(node: Node, mixins: Record<string, any>) {
   mixins[name] = { name, events: { event }, properties: { property } };
 }
 
+/**
+ * Handles a child node of type "event" and adds it to the event array.
+ * @param child The child node to handle.
+ * @param event The event array to update.
+ */
+function handleEventChild(child: Node, event: Event[], name: string) {
```
saschanaz: Every event will be child so no need to have Child in the name. Same for property.

src/build/patches.ts
saschanaz: I think this wouldn't become super huge to justify extra modules, for now let's keep it this way and see how it goes.

src/build/patches.ts
Bashamega: I have renamed it @saschanaz

src/build/patches.ts
```diff
@@ -66,14 +67,49 @@ function handleMixin(node: Node, mixins: Record<string, any>) {
...
+  mixins[name] = { name, events: { event }, properties: { property } };
+}
+
+/**
+ * Handles a child node of type "event" and adds it to the event array.
+ * @param child The child node to handle.
+ * @param event The event array to update.
+ */
+function handleEvent(child: Node, event: Event[], name: string) {
+  event.push({
```
saschanaz: Can we return Event here and do `event.push` in the caller? Like `event.push(handleEvent(child))`.

Also probably better to get the name in each function, as I think not all members will have a name, e.g. constructor.

Bashamega: I have updated it for events and properties

src/build/patches.ts
```diff
@@ -54,10 +55,10 @@ function handleEnum(node: Node, enums: Record<string, Enum>) {
 }
 
 /**
- * Handles a mixin node by extracting its name and associated events.
+ * Handles a mixin node by extracting its name and associated events and properties.
  * Throws an error if the mixin name is missing.
- * If the mixin node specifies "event" as its second value, it collects all child nodes as events,
- * each with a name and type, and adds them to the mixins record under the mixin's name.
+ * Uses helper functions to collect events and properties.
```
saschanaz: This is not useful information for the caller, so I'd skip this line.

src/build/patches.ts
```diff
@@ -54,10 +55,10 @@ function handleEnum(node: Node, enums: Record<string, Enum>) {
 }
 
 /**
- * Handles a mixin node by extracting its name and associated events.
+ * Handles a mixin node by extracting its name and associated events and properties.
```
```suggestion
 * Handles a mixin node by extracting its name and associated members.
```

(Because there will be more)

src/build/patches.ts
```diff
@@ -1,7 +1,8 @@
 import { parse, type Node } from "kdljs";
-import type { Enum, Event } from "./types";
+import type { Enum, Event, Property } from "./types";
 import { readdir, readFile } from "fs/promises";
 import { merge } from "./helpers.js";
+type Properties = Record<string, Omit<Property, "type">>;
```
```suggestion
type Properties = Record<string, Partial<Property>>;
```

saschanaz (commented):
LGTM with nits

Bashamega:
Thank you @saschanaz,
I have updated it

src/build/patches.ts
```diff
@@ -66,14 +66,48 @@ function handleMixin(node: Node, mixins: Record<string, any>) {
...
+        const propName = child.values[0] as string;
+        property[propName] = handleProperty(child);
+        break;
+      }
+      default:
+        throw new Error(`Unknown node name: ${child.name}`);
+    }
+  }
+
+  mixins[name] = { name, events: { event }, properties: { property } };
```
saschanaz, line 88: Would be nice to return a value as handleProperty now does. But as handleEnum already does this, this can be done in a separate PR.

saschanaz:
LGTM

PR #2096, opened 2025-08-02
Feature: Extend for mixins

Bashamega:
related #2053

### DIFF
--- a/inputfiles/patches/childnode.kdl
+++ b/inputfiles/patches/childnode.kdl
@@ -1,0 +1,1 @@
+interface-mixin ChildNode extends="Node"
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -92,11 +92,15 @@ function handleMixin(node: Node): DeepPartial<Interface> {
     }
   }
 
-  return {
+  const result: any = {
     name,
     events: { event },
     properties: { property },
-  } as DeepPartial<Interface>;
+  };
+  if (node.properties.extends) {
+    result.extends = node.properties.extends;
+  }
+  return result as DeepPartial<Interface>;
 }
 
 /**
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -19,9 +19,6 @@
                     }
                 }
             },
-            "ChildNode": {
-                "extends": "Node"
-            },
             "ParentNode": {
                 "extends": "Node"
             },


inputfiles/patches/childnode.kdl
```diff
@@ -0,0 +1 @@
+interface-mixin ChildNode extends="Node"
```
saschanaz: (skip quote)

src/build/patches.ts
```diff
@@ -92,11 +92,15 @@ function handleMixin(node: Node): DeepPartial<Interface> {
     }
   }
 
-  return {
+  const result: any = {
     name,
     events: { event },
     properties: { property },
-  } as DeepPartial<Interface>;
+  };
```
```suggestion
  const result = {
name,
events: { event },
properties: { property },
  } as DeepPartial<Interface>;
```

It's not clear to me why `const result: DeepPartial<Interface>` doesn't work (the error is confusingly complex), but this should be ok.

saschanaz (commented):
The direction looks good with some changes

inputfiles/patches/childnode.kdl
saschanaz: Also I don't think it's worth having a separate file for a single line change, but then I'm not sure what would be the right name. Maybe misc.kdl is fine and we can put random simple changes here.

Bashamega: I think extensions is a good name, because there are a bunch of things that we extend

saschanaz: And please cover ParentNode too, and some comment like:

```
// ChildNode and ParentNode are actually defined as mixins, but because of their names they have historically been used as a sub-interface of Node.
```

... with some line wrapping.

saschanaz: > I think extensions is a good name, because there are a bunch of things that we extend

Actually I change my mind here, covering ChildNode/ParentNode with some context is worth a separate file

src/build/patches.ts

inputfiles/patches/childnode.kdl
Bashamega: Done

saschanaz: Now child-parent-node.kdl or such? With that LGTM.

Bashamega: Done @saschanaz

saschanaz:
LGTM

jakebailey (approved):

saschanaz:
LGTM

PR #2100, opened 2025-08-03
Feature: Add method handling to mixin processing in KDL

Bashamega:
related #2053

### DIFF
--- a/inputfiles/patches/dom.kdl
+++ b/inputfiles/patches/dom.kdl
@@ -1,0 +1,12 @@
+// Manually moved from Document
+// See https://github.com/w3c/csswg-drafts/issues/5886 and https://github.com/w3c/csswg-drafts/issues/556
+interface-mixin DocumentOrShadowRoot {
+    method elementFromPoint returns="Element | null" {
+        param x type=number
+        param y type=number
+    }
+    method elementsFromPoint returns="Element[]" {
+        param x type=number
+        param y type=number
+    }
+}
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -1,5 +1,5 @@
 import { parse, type Node } from "kdljs";
-import type { Enum, Event, Property, Interface, WebIdl } from "./types";
+import type { Enum, Event, Property, Interface, WebIdl, Method } from "./types";
 import { readdir, readFile } from "fs/promises";
 import { merge } from "./helpers.js";
 
@@ -76,6 +76,7 @@ function handleMixin(node: Node): DeepPartial<Interface> {
 
   const event: Event[] = [];
   const property: Record<string, Partial<Property>> = {};
+  const method: Record<string, Partial<Method>> = {};
 
   for (const child of node.children) {
     switch (child.name) {
@@ -87,6 +88,11 @@ function handleMixin(node: Node): DeepPartial<Interface> {
         property[propName] = handleProperty(child);
         break;
       }
+      case "method": {
+        const methodName = child.values[0] as string;
+        method[methodName] = handleMethod(child);
+        break;
+      }
       default:
         throw new Error(`Unknown node name: ${child.name}`);
     }
@@ -96,6 +102,7 @@ function handleMixin(node: Node): DeepPartial<Interface> {
     name,
     events: { event },
     properties: { property },
+    methods: { method },
   } as DeepPartial<Interface>;
   if (node.properties.extends) {
     result.extends = node.properties.extends as string;
@@ -123,6 +130,19 @@ function handleProperty(child: Node): Partial<Property> {
     name: child.values[0] as string,
     exposed: child.properties?.exposed as string,
   };
+}
+
+/**
+ * Handles a child node of type "method" and adds it to the method object.
+ * @param child The child node to handle.
+ */
+function handleMethod(child: Node): Partial<Method> {
+  const name = child.values[0] as string;
+  // Build the overrideSignatures array with the method signature string
+  const overrideSignatures = [
+    `${name}(${child.children.map((c) => `${c.values[0]}: ${c.properties.type}`).join(", ")}): ${child.properties.returns || "void"}`,
+  ];
+  return { name, overrideSignatures };
 }
 
 /**
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -1,28 +1,4 @@
 {
-    "mixins": {
-        "mixin": {
-            "DocumentOrShadowRoot": {
-                // Manually moved from Document
-                // See https://github.com/w3c/csswg-drafts/issues/5886 and https://github.com/w3c/csswg-drafts/issues/556
-                "methods": {
-                    "method": {
-                        "elementFromPoint": {
-                            "name": "elementFromPoint",
-                            "overrideSignatures": [
-                                "elementFromPoint(x: number, y: number): Element | null"
-                            ]
-                        },
-                        "elementsFromPoint": {
-                            "name": "elementsFromPoint",
-                            "overrideSignatures": [
-                                "elementsFromPoint(x: number, y: number): Element[]"
-                            ]
-                        }
-                    }
-                }
-            }
-        }
-    },
     "callbackInterfaces": {
         "interface": {}
     },


inputfiles/patches/dom.kdl
```diff
@@ -0,0 +1,12 @@
+// Manually moved from Document
```
saschanaz: cssom-view.kdl

src/build/patches.ts
```diff
@@ -125,6 +132,19 @@ function handleProperty(child: Node): Partial<Property> {
...
 }
 
+/**
+ * Handles a child node of type "method" and adds it to the method object.
+ * @param child The child node to handle.
+ */
+function handleMethod(child: Node): Partial<Method> {
+  const name = child.values[0] as string;
+  // Build the overrideSignatures array with the method signature string
+  const overrideSignatures = [
```
saschanaz: We should not use overrideSignatures unless it's explicitly used. Take a look at widlprocess.ts to see how to proceed: (this PR does not need to cover everything in that file)

https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/23819c7e552e9e2b81f8042fa4ea9cf0890acbb3/src/build/widlprocess.ts#L192-L204

inputfiles/patches/cssom-view.kd
```diff
@@ -0,0 +1,12 @@
+// Manually moved from Document
+// See https://github.com/w3c/csswg-drafts/issues/5886 and https://github.com/w3c/csswg-drafts/issues/556
+interface-mixin DocumentOrShadowRoot {
+    method elementFromPoint returns="Element | null" {
```
saschanaz:
```suggestion
    method elementFromPoint returns="Element?" {
```

We want IDL language for types so that the script can parse and detect the support.

inputfiles/patches/cssom-view.kd
```diff
@@ -0,0 +1,12 @@
...
+        param x type=number
+        param y type=number
+    }
+    method elementsFromPoint returns="Element[]" {
```
saschanaz:
```suggestion
    method elementsFromPoint returns="sequence<Element>" {
```

Bashamega:
I've updated it, @saschanaz.
I used `array` instead of wrapping the return value with a `sequence`. If you still prefer your version, I can make it work with a regex. That said, I believe the current implementation is better for the parser.

inputfiles/patches/cssom-view.kd
```diff
@@ -0,0 +1,12 @@
+// Manually moved from Document
```
saschanaz: .kdl, not .kd 😛

Also maybe 2-space indent to be consistent with the rest of the code 👀

inputfiles/patches/cssom-view.kd
```diff
@@ -0,0 +1,12 @@
...
+// See https://github.com/w3c/csswg-drafts/issues/5886 and https://github.com/w3c/csswg-drafts/issues/556
+interface-mixin DocumentOrShadowRoot {
+    method elementFromPoint returns=Element nullable=#true {
+        param x type=long
+        param y type=long
+    }
+    method elementsFromPoint returns="Element" array {
```
saschanaz: Would be nice to follow the syntax of https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/23819c7e552e9e2b81f8042fa4ea9cf0890acbb3/src/build/types.d.ts#L1-L8

But that doesn't seem possible as a property. Perhaps we should allow (but not force) type as a member for complex cases like this.

```kdl
method elementsFromPoint {
  type sequence {
     type Element
  }
  param x type=long
  param y type=long
}
```

inputfiles/patches/cssom-view.kd
Bashamega: I have fixed the name

saschanaz: but not the indent 👀

inputfiles/patches/cssom-view.kd
Bashamega: Sure

inputfiles/patches/cssom-view.kd
Bashamega: I forgot, I have updated them

Bashamega:
I have updated it @saschanaz

Bashamega:
Hello @saschanaz 
Do you have any more requests for this PR?

inputfiles/patches/cssom-view.kdl
```diff
@@ -0,0 +1,15 @@
+// Manually moved from Document
+// See https://github.com/w3c/csswg-drafts/issues/5886 and https://github.com/w3c/csswg-drafts/issues/556
+interface-mixin DocumentOrShadowRoot {
+  method elementFromPoint returns=Element nullable=#true {
```
saschanaz: `nullable` is logically a property of the return type rather than the method, so in this case it would be also nicer to use separate type node as below.

src/build/patches.ts
```diff
@@ -142,6 +149,37 @@ function handleProperty(child: Node): Partial<Property> {
...
+  const params = child.children
+    .filter((c) => c.values[0] != "sequence")
+    .map((c) => ({
+      name: c.values[0] as string,
+      type: c.properties.type as string,
+    }));
+
+  const signature: Method["signature"] = [
+    {
+      type: type.name == "type" ? (type.values[0] as string) : returnType,
```
saschanaz: We'll probably get similar situation on other things than methods, this should probably be done in a helper function so that others can reuse it, something like `handleTyped`.

src/build/patches.ts
```diff
@@ -142,6 +149,37 @@ function handleProperty(child: Node): Partial<Property> {
...
+ * @param child The child node to handle.
+ */
+function handleMethod(child: Node): Partial<Method> {
+  const name = child.values[0] as string;
+  const type = child.children[0];
+  const returnType = child.properties.returns as string;
+  const nullable = child.properties.nullable as boolean;
+
+  const params = child.children
+    .filter((c) => c.values[0] != "sequence")
```
saschanaz: It should filter in based on the type `param`, not filter out based on name?

src/build/patches.ts
Bashamega: I have made it a helper function

src/build/patches.ts

Bashamega:
I have updated it @saschanaz

Bashamega:
Hello @saschanaz
Do you have any more requests for this PR?

Bashamega:
Welcome back from vacation, @saschanaz! 
I hope you had a great time. I’m excited to get this PR merged whenever you’re ready.

src/build/patches.ts
```diff
@@ -142,6 +170,31 @@ function handleProperty(child: Node): Partial<Property> {
   };
 }
 
+/**
+ * Handles a child node of type "method" and adds it to the method object.
+ * @param child The child node to handle.
+ */
+function handleMethod(child: Node): Partial<Method> {
+  const name = child.values[0] as string;
```
saschanaz: Too many `as string`, would be nice to have something like

```ts
function string(arg: unknown): string {
  if (typeof arg !== "string") {
    throw new Error(`Expected a string but found ${typeof arg}`);
  }
  return arg;
}
```

src/build/patches.ts
```diff
@@ -142,6 +170,31 @@ function handleProperty(child: Node): Partial<Property> {
...
+  const type = child.children[0];
```
saschanaz: We should iterate and check types than assuming a position or using `filter()`, otherwise we can't throw on unexpected child. Just like line 120.

src/build/patches.ts
```diff
@@ -105,6 +127,11 @@ function handleMixin(node: Node): DeepPartial<Interface> {
         property[propName] = handleProperty(child);
         break;
       }
+      case "method": {
+        const methodName = child.values[0] as string;
+        method[methodName] = handleMethod(child);
```
saschanaz, line 140: (Fine for now but eventually this will need to change to support overloads)

src/build/patches.ts
```diff
@@ -25,6 +32,20 @@ function optionalMember<const T>(prop: string, type: T, value?: Value) {
   };
 }
 
+function handleTyped(type: Node, returnType: string) {
+  const isTyped = type.name == "type";
+  const name = isTyped ? (type.values[0] as string) : returnType;
+  const subType =
+    type.children.length > 0
+      ? { type: type.children[0].values[0] as string }
```
saschanaz: This should be a recursive call to handleTyped

Bashamega:
No problem :)
I will take a look at the review today

src/build/patches.ts

Bashamega:
I have updated it @saschanaz 😃

Bashamega:
Hello @saschanaz,
Is there anything else that is missing here?

src/build/patches.ts
```diff
@@ -135,13 +160,58 @@ function handleEvent(child: Node): Event {
...
   };
 }
 
+/**
+ * Handles a child node of type "method" and adds it to the method object.
+ * @param child The child node to handle.
+ */
+function handleMethod(child: Node): Partial<Method> {
+  const name = string(child.values[0]);
+  const returnType = child.properties.returns;
```
saschanaz: This is not used here, maybe add back when it's actually used but not now

src/build/patches.ts
```diff
@@ -135,13 +160,58 @@ function handleEvent(child: Node): Event {
...
+
+  let typeNode: Node | undefined;
+  const params: { name: string; type: string }[] = [];
+
+  for (const c of child.children) {
+    switch (c.name) {
+      case "type":
+        if (typeNode) {
+          throw new Error(`Method "${name}" has multiple type nodes (invalid)`);
```
saschanaz, line 195: I actually want to allow this for union types, but for now this is ok as we don't have union types right now

src/build/patches.ts
```diff
@@ -135,13 +160,58 @@ function handleEvent(child: Node): Event {
...
+    throw new Error(`Method "${name}" is missing a return type`);
+  }
+
+  const signature: Method["signature"] = [
+    {
+      param: params,
+      ...handleTyped(typeNode, returnType),
+    },
+  ];
+  return { name, signature };
```
saschanaz, line 222: (This is fine for now but eventually we'll hit a problem when dealing with overloads)

saschanaz (commented):
Let's remove `returns` for now and then this should be ok. Thanks!

BTW I think we got a lot more conversion boilerplate than I expected... I'll think about how to simplify.

Bashamega:
Thank you for your review @saschanaz !
I have removed `returns`

Bashamega:
Hello @saschanaz 
Is there anything left here?

src/build/patches.ts
```diff
@@ -25,6 +33,28 @@ function optionalMember<const T>(prop: string, type: T, value?: Value) {
...
 }
 
+function string(arg: unknown): string {
+  if (typeof arg !== "string") {
+    throw new Error(`Expected a string but found ${typeof arg}`);
+  }
+  return arg;
+}
+
+function handleTyped(type: Node, returnType?: Value): Typed {
```
saschanaz: Still have unused parameter

Bashamega: We still use it for the subtype:
https://github.com/Bashamega/TypeScript-DOM-lib-generator/blob/0693d5d3d913903e66e808162f3dcee397681e9c/src/build/patches.ts#L50

saschanaz: Is it needed though? subtype would also be a type, so `isTyped` should be true, and then the second parameter won't be used, right?

Bashamega: I will update it

Bashamega:
Done @saschanaz

saschanaz:
LGTM

saschanaz:
LGTM (I feel like we are going to hit an anniversary of double LGTMing 😂)

jakebailey:
Well, I'm at least merging them

PR #2103, opened 2025-08-04
remove redundant language in MDN links

### LINKED ISSUE #2102: Redundant language in MDN links
Affected property - `textContent`

Bashamega:
fixes #2102 
I have removed the `en-US/` part in the URL since it is redundant, and also this makes it consistent with the auto-generated comments.


saschanaz:
This looks right. Thanks! LGTM

(It's ideal to omit the locale so that the automatic redirection to the browser locale can happen)

saschanaz:
LGTM

PR #2104, opened 2025-08-04
Migrate to Typescript-go

Bashamega:
Just a test PR, to see if this can move to Typescript-go

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,15 +12,15 @@
     "worklet"
   ],
   "scripts": {
-    "build": "tsc && node ./lib/build.js",
+    "build": "tsgo && node ./lib/build.js",
     "baseline-accept": "cpx \"generated\\**\" baselines\\",
-    "lint": "eslint --max-warnings 0 src deploy/*.js && tsc -p deploy/jsconfig.json",
+    "lint": "eslint --max-warnings 0 src deploy/*.js && tsgo -p deploy/jsconfig.json",
     "lint-fix": "eslint --max-warnings 0 src deploy/*.js --fix",
     "test": "npm run build && npm run lint && node ./lib/test.js && node ./unittests/index.js",
-    "changelog": "tsc && node ./lib/changelog.js",
+    "changelog": "tsgo && node ./lib/changelog.js",
     "ts-changelog": "node ./deploy/versionChangelog.js",
     "migrate": "node ./deploy/migrate.js",
-    "version": "npm i && tsc && node ./lib/version.js",
+    "version": "npm i && tsgo && node ./lib/version.js",
     "generate": "npm run build && npm run baseline-accept"
   },
   "author": {
@@ -68,5 +68,8 @@
   },
   "overrides": {
     "typescript@*": "$typescript"
+  },
+  "dependencies": {
+    "@typescript/native-preview": "^7.0.0-dev.20250804.1"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
       "name": "@typescript/dom-lib-generator",
       "version": "0.0.1",
       "license": "Apache-2.0",
+      "dependencies": {
+        "@typescript/native-preview": "^7.0.0-dev.20250804.1"
+      },
       "devDependencies": {
         "@eslint/js": "^9",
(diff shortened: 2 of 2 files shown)


saschanaz:
Seems the answer is "not yet"

jakebailey:
First the problem is to fix this repo's imports; the repo is still using the old deprecated `node10` resolution, which we did not implement in TS7.

saschanaz:
(I want my maintainer permission to push to this PR and see how it goes 🙈)

jakebailey:
Also, you definitely do not want to accept this PR as is, the native preview is just a preview and should not "replace" tsc at the time being. A new workflow to check it seems fine.

package.json
```diff
@@ -68,5 +68,8 @@
   },
   "overrides": {
     "typescript@*": "$typescript"
+  },
+  "dependencies": {
```
jakebailey: Should be a dev dep

jakebailey:
> (I want my maintainer permission to push to this PR and see how it goes 🙈)

Me too but I haven't gotten that settled yet ☹️

package.json
Bashamega: I have used dev dep

Bashamega:
The build error should be fixed when the next version of BCD is published

Bashamega:
I have updated it to use the latest bcd and typescript syntax @jakebailey 
I think the current issue is a problem with typescript-go.
It says that `optionsOrLoose` in semver is required. But it is actually optional:
image
Or is there something that I am missing?

jakebailey:
It's using a jsconfig and then looking into the JS types for the package instead of the dts file, which is interesting. I'm not sure any of us have actually tested jsconfig.

Worth a bug and a more minimal repro.

jakebailey:
Oh, the types for semver aren't even installed in the repo. Yeah, I think this might be expected.

Bashamega:
I will create a minimal repro and open an issue today.
Thanks for your reply

saschanaz:
Why did we go ahead to fix BCD instead of investigating why tsgo fails on BCD even though ts5 is okay with it?

Bashamega:
> First the problem is to fix this repo's imports; the repo is still using the old deprecated `node10` resolution, which we did not implement in TS7.

I think this answers your question @saschanaz

saschanaz:
But this PR doesn't use node10 anymore?

jakebailey:
We didn't implement node10 in tsgo; right now it silently maps itself to bundler resolution for easier testing, but will error in the future. The errors shown in this PR (before the dep fixes) would appear with tsc in newer resolution modes.

saschanaz:
Hmm actually, ts5 is not okay with it with node16 👀

saschanaz:
Understood, thanks for fixing it 👍🏻

We should probably open new PR to switch from node10 and land it.

Bashamega:
> Understood, thanks for fixing it 👍🏻
> 
> We should probably open new PR to switch from node10 and land it.

I will open a separate PR for it

jakebailey:
The build failure here should be fixed in the latest tsgo nightly.

Bashamega:
The tsgo test is much faster 

Any ways I think this is ready @jakebailey

HolgerJeromin:
Perhaps the PR title should be changed.
This adds tsgo but not removed ts.

.github/workflows/tsgo.yml
```diff
@@ -0,0 +1,22 @@
...
+          submodules: true  # Ensures submodules are cloned
+
+      
+      - uses: actions/setup-node@v5
+        with:
+          node-version: "lts/*"
+          cache: npm
+
+      - run: npm ci
+      - run: npx tsgo && node ./lib/build.js
```
jakebailey: I'd rather this be a job in the regular workflow if anything

Bashamega:
I have updated it @jakebailey

jakebailey (approved):
In general this looks good. My main concern now is just this version getting stale, given how often tsgo is being updated and fixed, so would be good to be rechecked. But maybe dependabot/etc will be sufficient.

Bashamega:
I think dependabot is fine, or I can create a script to update it.

PR #2108, opened 2025-08-06
Use node16

Bashamega:
Split #2104


inputfiles/mdn
jakebailey, line 1: Seems mistakenly changed.

Bashamega: Yes, I am trying to fix it

Bashamega:
I have fixed it @jakebailey :)

inputfiles/mdn
jakebailey: Usually it's just something like

```
git fetch origin
git checkout origin/main -- input files/mdn
```

Bashamega: Chat GPT helped me 😄

jakebailey:
To be clear this is fine, but I probably should have suggested just using nodenext

Bashamega:
> To be clear this is fine, but I probably should have suggested just using nodenext

If you want, I can open another PR

src/build/patches.ts
```diff
@@ -1,5 +1,5 @@
 import { parse, type Value, type Node } from "kdljs";
-import type { Enum, Event, Property, Interface, WebIdl } from "./types";
+import type { Enum, Event, Property, Interface, WebIdl } from "./types.js";
```
saschanaz, line 2: Too late but this should really be d.ts

src/build/bcd.ts
```diff
@@ -1,4 +1,4 @@
-import * as Browser from "./types";
+import * as Browser from "./types.js";
```
saschanaz, line 1: Same

src/build/bcd/mapper.ts
```diff
@@ -5,7 +5,7 @@ import {
   SimpleSupportStatement,
 } from "bcd-idl-mapper";
 import api from "bcd-idl-mapper";
-import * as Browser from "../types";
+import * as Browser from "../types.js";
```
saschanaz, line 8: Same

src/build/patches.ts
jakebailey: No, definitely not; this is correct.

jakebailey:
> > To be clear this is fine, but I probably should have suggested just using nodenext
> 
> If you want, I can open another PR

If you do I'll approve it; node16 is likely going to be deprecated in 6.0.

src/build/patches.ts
jakebailey: Note that the PR moved the file to .ts; you can't import a .d.ts file straight up like that without also enabling flags that allow importing TS files.

PR #2111, opened 2025-08-07
Use KDL for nonce

Bashamega:
I have migrated nonce to KDL and used the optional property instead of renaming it.

### DIFF
--- a/baselines/ts5.5/dom.generated.d.ts
+++ b/baselines/ts5.5/dom.generated.d.ts
@@ -16385,7 +16385,7 @@ interface HTMLOrSVGElement {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/dataset) */
     readonly dataset: DOMStringMap;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/nonce) */
-    nonce?: string;
+    nonce?: string | undefined;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/tabIndex) */
     tabIndex: number;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/blur) */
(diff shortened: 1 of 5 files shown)


baselines/ts5.6/dom.generated.d.ts
```diff
@@ -16403,7 +16403,7 @@ interface HTMLOrSVGElement {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/dataset) */
     readonly dataset: DOMStringMap;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/nonce) */
-    nonce?: string;
```
Bashamega, line 16406: This is how the emitter handles optional properties, and I don't see an issue with this

baselines/ts5.5/dom.generated.d.ts
```diff
@@ -16385,7 +16385,7 @@ interface HTMLOrSVGElement {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/dataset) */
     readonly dataset: DOMStringMap;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/nonce) */
-    nonce?: string;
+    nonce?: string | undefined;
```
Bashamega, line 16388: This is how the emitter handles optional properties, and I don't see an issue with this

saschanaz:
The original PR introduced it is https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/626

Which links to https://github.com/microsoft/TypeScript/issues/28452

At this point everyone supports `nonce` and thus we should just remove it from `overridingTypes.jsonc`.

(This is why additional comments with context would be helpful for each patching)

Bashamega:
Thank you for your review @saschanaz 
I have updated it

saschanaz:
(but it did not remove nonce, could be nicer with some different message)

jakebailey:
Ah, true; it removed nonce from the overrides. Could have renamed it, sorry.

PR #1947, opened 2025-03-24
fix: `LockManager.request` promises the return of its callback parameter

turbocrime:
https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1946

PR #2112, opened 2025-08-08
Correct `LockManager.prototype.request` return type with async callback

Renegade334:
Resolves microsoft/TypeScript#62231, which results from #1947.

refs: https://www.w3.org/TR/web-locks/#api-lock-manager-request


saschanaz:
Interesting, can you add a unit test too? It's in `unittests/files`.

saschanaz:
LGTM, thanks!

saschanaz:
LGTM

PR #2124, opened 2025-08-25
Add SecurityPolicyViolationEvent & securitypolicyviolation to WorkerG…

SkyZeroZx:
# Add missing `securitypolicyviolation` event to WorkerGlobalScope

## Summary
This PR adds the missing `securitypolicyviolation` event API to `WorkerGlobalScope` in the TypeScript DOM lib generator, completing the CSP violation reporting support for Web Workers.

## Specifications Compliance
W3C CSP Level 3: [Report Violation](https://w3c.github.io/webappsec-csp/#report-violation)
MDN Documentation: [WorkerGlobalScope securitypolicyviolation event](https://developer.mozilla.org/docs/Web/API/WorkerGlobalScope/securitypolicyviolation_event)

## Implementation Details
The implementation adds:
```typescript
interface WorkerGlobalScopeEventMap {
    // ... existing events
    "securitypolicyviolation": SecurityPolicyViolationEvent;
}

interface WorkerGlobalScope extends EventTarget {
    // ... existing properties
    onsecuritypolicyviolation: ((this: WorkerGlobalScope, ev: SecurityPolicyViolationEvent) => any) | null;
}
```

### DIFF
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -10779,6 +10779,7 @@ interface WorkerGlobalScopeEventMap {
     "offline": Event;
     "online": Event;
     "rejectionhandled": PromiseRejectionEvent;
+    "securitypolicyviolation": SecurityPolicyViolationEvent;
     "unhandledrejection": PromiseRejectionEvent;
 }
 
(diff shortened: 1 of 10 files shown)


SkyZeroZx:
@microsoft-github-policy-service agree [company="{your company}"]

SkyZeroZx:

saschanaz:
It looks like MDN is wrong, ``new Worker(URL.createObjectURL(new Blob([`console.log("onsecuritypolicyviolation" in globalThis)`])))`` gives false on Firefox and Chrome. Maybe file a bug in MDN?

SkyZeroZx:
> It looks like MDN is wrong, `` new Worker(URL.createObjectURL(new Blob([`console.log("onsecuritypolicyviolation" in globalThis)`]))) `` gives false on Firefox and Chrome. Maybe file a bug in MDN?
Firefox return true for me 
image
image

Chrome the same
image
image

saschanaz:
Your code is not exactly running the console.log in worker environment, maybe copypaste my code and try again?

SkyZeroZx:
If I misinterpreted it, I just ran it in a sandbox with https/localhost , and the result is false 
So it would be an error in the documentation or a bug in the browsers?

saschanaz:
It is MDN issue, the spec doesn't say it should be available in workers.

Renegade334:
The issue is with the HTML IDL – see w3c/webappsec-csp#568 for context. The event was removed from the HTML spec but the intent was for the `onsecuritypolicyviolation` IDL attribute to remain on the relevant interfaces, and it doesn't look like this occurred for `WorkerGlobalScope`.

saschanaz:
"could in theory also fire" in the worker. But that did not happen and browsers do not implement it. File a spec bug maybe?

saschanaz:
Closing for now, but please feel free to file a spec bug to proceed.

@github-actions close

PR #2140, opened 2025-09-05
Remove Workflow trigger documentation

Bashamega:
It doesn't exist and goes to a 404 page


HolgerJeromin (approved):

saschanaz:
It's removed in #1859, LGTM

saschanaz:
Actually the removal is not enough, it's not removing the whole section but only the initial part of it.

Bashamega:
> Yeah we have to do this by hand now 

Can you fix this?
I don't think I can, because I am not a maintainer

jakebailey:
Aghhh that's what I get for merging from my phone

saschanaz:
Now README is malformed..

PR #2141, opened 2025-09-06
Update README.md

Bashamega:
I have fixed it @jakebailey, @saschanaz


Bashamega:
Also added a comment for KDL

README.md
```diff
@@ -86,12 +86,6 @@ If you are familiar with Web IDL, you may also want to check whether the upstrea
 - `addedTypes.jsonc`: types that should exist but are missing from the spec data.
 - `overridingTypes.jsonc`: types that are defined in the spec but have TypeScript-friendly modifications in the json files.
 - `removedTypes.jsonc`: types that are defined in the spec but should be removed.
+- `patches/*.kdl`: KDL types
 - `comments.json`: comment strings to be embedded in the generated .d.ts files.
 - `deprecatedMessage.json`: the reason why one type is deprecated.
-
-    npm run ts-changelog -- @types/web 0.0.2 0.0.23
```
saschanaz, line 92: This still exists and probably part of the manual job? IDK what people actually do, though.

saschanaz: And should we document how the manual job goes?

jakebailey (dismissed):

README.md
Bashamega: I think no one uses it ... I never used it
Maybe it is only for maintainers, Idk even what it does, I never had to generate a change log.

jakebailey: I'm honestly not sure it matters

jakebailey: Yeah, I looked and I don't think we have used this in recent memory

jakebailey (approved):
Sorry for the earlier mismerge, I noticed the syntax error too late.

I'll merge this in now, as I'm pretty certain we don't use the script, and it's probably worth relooking at that later. But at the moment I don't want the readme to be broken 😑

PR #2147, opened 2025-09-12
Selector in `Element.matches()` should type narrow

voxpelli:
Adds TypeScript overloads for the [`matches`](https://developer.mozilla.org/docs/Web/API/Element/matches) method.

The overloads leverages the `HTMLElementTagNameMap` similar to how overloads are handled for methods like `querySelector`, but here for type narrowing by making it eg. `this is HTMLElementTagNameMap[K]`.

Tests have been run locally and are passing.

### Changes:

* Added a new function `emitMatchesOverloads` to generate overload signatures for the `matches` method.
* Updated the main emitter logic to call `emitMatchesOverloads` when processing the `matches` method.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -11291,6 +11291,9 @@ interface Element extends Node, ARIAMixin, Animatable, ChildNode, NonDocumentTyp
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Element/matches)
      */
+    matches<K extends keyof HTMLElementTagNameMap>(selectors: K): this is HTMLElementTagNameMap[K];
+    matches<K extends keyof SVGElementTagNameMap>(selectors: K): this is SVGElementTagNameMap[K];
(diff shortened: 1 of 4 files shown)


voxpelli:
@microsoft-github-policy-service agree company="040 Internet AB"

saschanaz:
LGTM for now, but I think this should ultimately move to a patch rather than in the emitter itself.\

voxpelli:
@saschanaz Weird error: "HttpError: Repository rule violations found New changes require approval from someone other than the last pusher. "

jakebailey:
It's unfortunately normal, the infra was written before new rules required approval from someone with merge perms (defeating the purpose more or less).

I wasn't sure if the comment was an actual ask to merge or not, but I can just approve it.

jakebailey (approved):

saschanaz:
LGTM

MartinJohns:
This change breaks code: https://github.com/microsoft/TypeScript/issues/63497

PR #2148, opened 2025-09-14
feat: add the ability override extended interfaces, fixes HTMLElement onerror type

soda-sorcery:
## Description
This PR started as a fix for [this TypeScript issue](https://github.com/Microsoft/TypeScript/issues/62075). To handle it, I needed to update how the `emitter.ts` handled generating interfaces.

The `HTMLElement` extends the `GlobalEventHandlers.onerror` property. I didn't want to blow away the whole type to handle this fix. I thought something more flexible would make fixes like this scalable. 
I was looking at 
```ts
const extendConflictsBaseTypes: Record<
  string,
  { extendType: string[]; memberNames: Set<string> }
> = {
  HTMLCollection: {
    extendType: ["HTMLFormControlsCollection"],
    memberNames: new Set(["namedItem"]),
  },
};
```
and thought that is where I could add the flexibity by adding a new key but I didn't want to break any existing functionality or misuse a type.

I added 
```ts
const extendConflictsInterfaces: Record<
  string,
  { overrideMap: Record<string, string> }
> = {
  HTMLElement: {
    overrideMap: {
      GlobalEventHandlers: 'Omit<GlobalEventHandlers, "onerror">',
    },
  },
};
```

It allows you to define a type and override its interfaces when unique cases arise.

I've included fixes for `HTMLElement` and `SvgElement`. If these changes are approved but we'd like to split the PR up into discrete chunks that's no problem. I'm also open to other approaches so if there's a more obvious place to bring this in I can update the PR.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -29789,6 +29789,12 @@ interface SVGSVGElement extends SVGGraphicsElement, SVGFitToViewBox, WindowEvent
     unsuspendRedraw(suspendHandleID: number): void;
     /** @deprecated */
     unsuspendRedrawAll(): void;
+    /**
+     * The addEventListener() method of the EventTarget interface sets up a function that will be called whenever the specified event is delivered to the target.
+     *
+     * [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)
+     */
+    addEventListener(type: string, listener: ((event: Event) => void) | ((event: UIEvent) => void)): void;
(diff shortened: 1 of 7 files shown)


cosmic-pixel-painter:

saschanaz:
Where did UIEvent come from? Looks like a bug.

I think we should instead add an interface that inherits from GlobalEventHandlers (maybe name it DocumentOrElementEventHandlers or such) and make others inherit from it, that way we don't need to repeat onerror for each interface and no need to touch emitter logic (as overridingTypes.jsonc should be able to handle it).

cosmic-pixel-painter:
That's a good call. Let me see what I can do there. 

I have a question out curiosity: is there ever a situation where using something like that `omit` in the emitter would be necessary or was I misunderstanding some of the architecture of `overridingTypes.jsonc`?

PR #2150, opened 2025-09-20
Migrate ImageOrientation to KDL

Bashamega:
#2053

### DIFF
--- a/inputfiles/patches/dom.kdl
+++ b/inputfiles/patches/dom.kdl
@@ -4,3 +4,10 @@ enum InsertPosition {
     afterbegin
     afterend
 }
+
+enum ImageOrientation {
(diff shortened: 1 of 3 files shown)


inputfiles/patches/dom.kdl
```diff
@@ -4,3 +4,10 @@ enum InsertPosition {
     afterbegin
     afterend
 }
+
+enum ImageOrientation {
```
saschanaz: This is not dom but html, so html.kdl

saschanaz (commented):
Looks good with a new file

inputfiles/patches/dom.kdl
Bashamega: Done

saschanaz:
LGTM

jakebailey (approved):

PR #2153, opened 2025-09-21
Support KDL Interfaces

Bashamega:
#2053

### DIFF
--- a/inputfiles/patches/html.kdl
+++ b/inputfiles/patches/html.kdl
@@ -5,6 +5,6 @@ enum ImageOrientation {
     none
 }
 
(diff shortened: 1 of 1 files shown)


src/build/patches.ts
```diff
@@ -68,6 +68,7 @@ function parseKDL(kdlText: string): DeepPartial<WebIdl> {
   const nodes = output!;
   const enums: Record<string, Enum> = {};
   const mixin: Record<string, DeepPartial<Interface>> = {};
+  const interfaces: Record<string, DeepPartial<Interface>> = {};
```
Bashamega, line 71: I had to name it interfaces because interface is a reserved word

Bashamega:
What do you think @saschanaz

PR #2155, opened 2025-09-24
Move most HTML related interfaces to KDL

Bashamega:
#2053 

NOTE: I didn't migrate methods, because there is still some missing pieces in the parser

### DIFF
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -238,7 +238,6 @@ function handleMethod(child: Node): Partial<Method> {
   const name = string(child.values[0]);
 
   const typeNode = findTypeNode(child.children, `Property "${name}"`);
-  if (!typeNode) throw new Error(`Method "${name}" is missing a return type`);
   const params: { name: string; type: string }[] = [];
 
   for (const c of child.children) {


Bashamega:
I have rebased it @saschanaz

src/build/patches.ts
```diff
@@ -210,6 +210,7 @@ function handleProperty(child: Node): Partial<Property> {
     ...optionalMember("optional", "boolean", child.properties?.optional),
     ...optionalMember("overrideType", "string", child.properties?.overrideType),
     ...optionalMember("type", "string", child.properties?.type),
+    ...optionalMember("nullable", "boolean", child.properties?.nullable),
```
saschanaz: `nullable` belongs to type rather than property itself, so I think this should go to type node (and thus use type node in the patch file too)

Bashamega:
I have updated it to use the type and also refactored the code @saschanaz

Bashamega:
Can you review this @saschanaz ?

src/build/patches.ts
```diff
@@ -55,6 +55,19 @@ function handleTyped(type: Node): Typed {
   };
 }
 
+function findTypeNode(children: Node[], context: string): Node | undefined {
```
saschanaz, line 58: Why the split? I don't find this easier to understand. The code is longer now while the logic is still same, right?

Bashamega: This is actually shorter @saschanaz, since we now use it for both the method and the property. I factored it out into its own function to keep things DRY.

Bashamega:
Hello @saschanaz 
Can you please review this?

Bashamega:
Can you check this @saschanaz

PR #2160, opened 2025-09-24
Support overrideSignatures for methods

Bashamega:
#2053


Bashamega:
I forgot that we should not use this

PR #2162, opened 2025-09-25
Migrate interface that we are forced to declare

Bashamega:
#2053 

I love how neat it is


saschanaz:
Doing multiple things are fine if converter change is not needed, but if it's needed it complicates review process. Can we do it one by one, for each converter change?

Bashamega:
> Doing multiple things are fine if converter change is not needed, but if it's needed it complicates review process. Can we do it one by one, for each converter change?

Sure

PR #2163, opened 2025-09-25
🤖 Update core dependencies

Bashamega:
Closes #2149


saschanaz:
Tag and TagType should go into WebAssembly, not the other way.

Bashamega:
Done @saschanaz

PR #2168, opened 2025-09-26
Move iterables into main d.ts files in 6.0, reference es2015 and es2018.asynciterable

jakebailey:
For https://github.com/microsoft/TypeScript/pull/62111


saschanaz:
LGTM

petamoriken:
Can I add `Symbol.toStringTag` to `*.generated.d.ts`? #1762

jakebailey:
That has nothing to do with this PR

saschanaz:
But it is effectively blocked by this, so maybe?

jakebailey:
Hm, maybe? It wouldn't work for older TS versions, so I'm not sure where we'd put it other than to not generate it at all for older TS?

saschanaz:
I mean we already make things different for different TS versions. But anyway this discussion can continue in #1762 after this PR.

jakebailey:
Yeah, sorry, I did not realize that they were referring to the side effect change of the types referencing es2015

PR #2169, opened 2025-09-27
Support methods without names in interfaces

Bashamega:
#2053 

This PR adds support for methods without explicit names.

I got the name "callable", from the jsonc


inputfiles/patches/forced-declare.kdl
```diff
@@ -0,0 +1,7 @@
+// Keeping EventListener and EventListenerObject isn't the most elegant way to handle the event listeners, but we need to keep the EventListener as an extendable interface for libraries like angular.
+interface EventListener noInterfaceObject=#true {
+    method callable {
+        type void
```
saschanaz: This is a good reason to use `returns=void`

inputfiles/patches/forced-declare.kdl
```diff
@@ -0,0 +1,7 @@
+// Keeping EventListener and EventListenerObject isn't the most elegant way to handle the event listeners, but we need to keep the EventListener as an extendable interface for libraries like angular.
+interface EventListener noInterfaceObject=#true {
+    method callable {
```
saschanaz: We already have `interface.anonymouseMethods` so use it with something like `anonymous-method {}`?

Bashamega: Sure, I will update it

Bashamega:
I have updated @saschanaz 
But anonymousMethods is not working, can you tell me what is wrong?

saschanaz:
Ohno, I forgot that it was only for getters and stringifiers... sorry 🫠

But maybe `method ""` works? `""` is empty string in KDL.

Bashamega:
No issues @saschanaz 
Is it now better?

src/build/patches.ts
```diff
@@ -244,17 +244,18 @@ function handleMethod(child: Node): Partial<Method> {
...
       param: params,
-      ...handleTyped(typeNode),
+      ...(typeNode
+        ? handleTyped(typeNode)
+        : { type: string(child.properties?.returns) }),
     },
   ];
-  return { name, signature };
+  return {
+    ...optionalMember("name", "string", name),
```
saschanaz: Why should this be optionalMember?

Bashamega: I forgot to update it, done

saschanaz:
LGTM

inputfiles/patches/forced-declare.kdl
```diff
@@ -0,0 +1,6 @@
+// Keeping EventListener and EventListenerObject isn't the most elegant way to handle the event listeners, but we need to keep the EventListener as an extendable interface for libraries like angular.
+interface EventListener noInterfaceObject=#true {
```
saschanaz: Erm actually this is dom, let's put this in dom.kdl

Bashamega:
I have moved it @saschanaz

saschanaz:
LGTM2

jakebailey (approved):

PR #2170, opened 2025-09-27
Regroup patches into each spec

saschanaz:
It's not useful to have separate files for single line patches.


saschanaz:
In overridingTypes.jsonc yes because there's no topic-based grouping. But each topic-based patch should be short enough to find things.

Bashamega:
Okay, then LGTM😃

saschanaz:
LGTM

PR #2171, opened 2025-09-29
MDN description V2

### LINKED ISSUE #2151: Migrate to MDN metadata.json
MDN has an API endpoint that gives summaries for each document: https://developer.mozilla.org/en-US/metadata.json

The data looks sufficient for this repository, and we can save a submodule (and prevent pain like #2137, also makes #2078 redundant)

### LINKED ISSUE #2209: don't include js-nolint in tooltips
### Expected behaviour

The `js-nolint` is not part of the editor tooltips and tooltip matches the sentence is like on MDN:

Image

```
The read() method of the FileSystemSyncAccessHandle interface reads the content of the file associated with the handle into a specified buffer, optionally at a given offset.
```

### Actual behaviour

The `js-nolint` is part of the editor tooltips:

Image

```
The read() method of the js-nolint read(buffer, options) - buffer - : An ArrayBuffer or ArrayBufferView (such as a DataView) representing the buffer that the file content should be read into.
```

Bashamega:
Migrate to be using the json API
closes #2151 
closes #2086 
closes #2078 
closes #2209


baselines/dom.iterable.generated.d.ts
```diff
@@ -4,7 +4,10 @@
 
 interface AudioParam {
     /**
-     * The **`setValueCurveAtTime()`** method of the AudioParam interface schedules the parameter's value to change following a curve defined by a list of values.
+     * The
+     * setValueCurveAtTime() method of the
+     * AudioParam interface schedules the parameter's value to change
+     * following a curve defined by a list of values.
```
saschanaz: newline-to-space normalization seems to be needed.

src/build/mdn-comments.ts
```diff
@@ -1,63 +1,20 @@
-import fs from "fs/promises";
-const basePath = new URL("../../inputfiles/mdn/files/en-us/", import.meta.url);
+import { readFile } from "fs/promises";
+const inputFile = new URL("../../inputfiles/mdn/mdnData.json", import.meta.url);
```
saschanaz: Maybe just `inputfiles/mdn.json`? And you can do `import * as data from "(path)" with { type: "json" }`

src/build/mdn-comments.ts
```diff
@@ -1,63 +1,20 @@
...
+
+interface MetaDataEntry {
```
saschanaz: I don't think we should treat meta and data separately, let's do Metadata.

Bashamega:
I have updated it @saschanaz,
But the import JSON file didn't work for some reason. 
This is the error: https://github.com/microsoft/TypeScript-DOM-lib-generator/actions/runs/18124591077/job/51576830796?pr=2171

baselines/ts5.5/audioworklet.generated.d.ts
```diff
@@ -1745,127 +1717,51 @@ declare namespace WebAssembly {
...
+    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/console/timeLog_static) */
     timeLog(label?: string, ...data: any[]): void;
     timeStamp(label?: string): void;
-    /**
-     * The **`console.trace()`** static method outputs a stack trace to the console.
-     *
-     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/console/trace_static)
-     */
+    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/console/trace_static) */
     trace(...data: any[]): void;
```
jakebailey, line 1929: Seems like these docs are missing?

.github/workflows/update-core-deps.yml
```diff
@@ -12,11 +12,18 @@ jobs:
...
+        working-directory: ./inputfiles
+        run: |
+          curl -fL 'https://developer.mozilla.org/en-US/metadata.json' \
+            | jq '{ data: [ .[]
+              | select(
+                (.mdn_url | ascii_downcase
+                | (startswith("/en-us/docs/web/api/") or startswith("/en-us/docs/webassembly/reference/javascript_interface/")))
+              )
+              | { mdn_url, pageType, summary } ] }' \
+            > mdn.json
```
saschanaz: Why do bash magic when we can filter it in the script instead?

saschanaz (commented):
I think you reverted too early. The module error doesn't happen locally.

Bashamega:
The module error also occurs on my local machine (Ubuntu), @saschanaz.

I used the bash magic because it significantly reduces the output size — most of the data isn’t needed.
For comparison:

Current (filtered and formatted): 2.5 MB

Original (unfiltered, unformatted): 28.7 MB

That’s over 11× larger (28.7 MB vs 2.5 MB), so I think the extra shell step is worth it.
Plus, the command is still quite readable.

Bashamega:
I’ve added support for _static, so there is no missing comments now, @jakebailey.

saschanaz:
> The module error also occurs on my local machine (Ubuntu), @saschanaz.

Are you seeing the exact same module error? Because that error says you don't have lib/build.js, which means typescript silently failed to compile anything at all. Can you check what's in lib/ in that case?
 
> I used the bash magic because it significantly reduces the output size — most of the data isn’t needed. For comparison:
> 
> Current (filtered and formatted): 2.5 MB
> 
> Original (unfiltered, unformatted): 28.7 MB
> 
> That’s over 11× larger (28.7 MB vs 2.5 MB), so I think the extra shell step is worth it. Plus, the command is still quite readable.

Hmm, fair. I think we should still fetch and filter in typescript instead, as bash usage would break on Windows.

saschanaz:
Or well, in JS, to save a build step.

Bashamega:
> Hmm, fair. I think we should still fetch and filter in typescript instead, as bash usage would break on Windows.

I don't think that this is an issue, since no one would actually run it, but sure, I will write a script to generate the file

Bashamega:
> Are you seeing the exact same module error? Because that error says you don't have lib/build.js, which means typescript silently failed to compile anything at all. Can you check what's in lib/ in that case?
Yes, it is the same error

saschanaz:
Can you check what's in lib/ in that case?

Bashamega:
image
Here

Bashamega:
I have removed the bash magic @saschanaz

.github/workflows/update-core-deps.yml
```diff
@@ -31,6 +27,8 @@ jobs:
       # This prevents annoying change when contributors run `npm i` on their local machine.
       # Example: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1463
       - run: npm i
+      - name: Update MDN Data
+        run: node scripts/generateMDN
```
saschanaz: We are not really generating but more fetching.

really not a fan of camelCasing on file paths though 🫠

I do recognize we have deploy/ which is mostly written by Orta than me, and we do have jsonc files that preexisted, but any new files have been hyphen-cased...

src/build/mdn-comments.ts
```diff
@@ -115,35 +45,31 @@ function insertComment(
   }
 }
 
+function extractSlug(slug: string): string[] {
```
saschanaz: Why reorder the function? Almost thought this is an entirely new function with the same name.. Keeping its position should help reducing the diff.

saschanaz (commented):
Looks good, but this will need to update with the latest main branch.

Bashamega:
I have renamed it, and rebased it @saschanaz

baselines/ts5.5/dom.iterable.generated.d.ts
```diff
@@ -281,7 +281,7 @@ interface PluginArray {
 
 interface RTCRtpTransceiver {
     /**
-     * The **`setCodecPreferences()`** method of the RTCRtpTransceiver interface is used to set the codecs that the transceiver allows for decoding _received_ data, in order of decreasing preference.
+     * The setCodecPreferences() method of the RTCRtpTransceiver interface is used to set the codecs that the transceiver allows for decoding received data, in order of decreasing preference.
```
jakebailey: Are we just generally losing all markup?

Bashamega: Yes, because the JSON doesn't have Markdown

Bashamega: Yes, because the JSON doesn't have Markdown

saschanaz: Yeah. We could hack around by matching the first occurrence of the name though...

Bashamega: This PR is big as is, maybe a separate PR

saschanaz: Technically hacking around would reduce the size of the PR.

saschanaz: (You can open PR on my fork and then I can try it)

Bashamega: Okay, I will hack it today

Bashamega:
Hello @saschanaz 
I have restored most of the markdown, and updated the branch
Do you have any other requests?

src/build/mdn-comments.ts
```diff
@@ -117,37 +45,50 @@ function insertComment(
...
+function generateComment(summary: string, name: string): string {
+  // Escape special regex characters in the name
+  const escapedName = name.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+  return summary
+    .replace(/\n/g, " ") // remove newlines
+    .replace(
+      // Match optional preceding identifier + dot OR just the name itself
+      new RegExp(`(?:\\b\\w+\\.)?${escapedName}(\\(\\))?`),
+      (match) => `**\`${match}\`**`,
```
saschanaz: Would be happier with less regexp as possible, something like this should work:

```js
// `parent` can be `slugArr.at(-2)` from the caller
.replace(` ${parent}.${name}\(\) `, `**\`$1\`**`)
.replace(` ${parent}.${name} `, `**\`$1\`**`)
.replace(` ${name}\(\) `, `**\`$1\`**`)
.replace(` ${name} `, `**\`$1\`**`)
```

Bashamega:
Hello @saschanaz 
I wasn't able to remove the regex because it would change other things. I have opened a PR on your fork, so you can try it. When you are ready to merge, just notify me so I can rebase this repo

Bashamega:
Hello @saschanaz 

any updates?

jakebailey:
Can you merge main and update again?

Bashamega:
I have updated it @jakebailey

src/build/mdn-comments.ts
jakebailey: @Bashamega are you looking into this?

Bashamega: I wasn't able to remove the regex because it would change other things.
@jakebailey

jakebailey: Honestly, I feel like we don't even need this. Was the style originally only this way because that's what the markdown in MDN did?

I'm going through the diff and there are places we are now _making_ bold that did not used to be. It seems like we'd save space and probably extra formatting noise by not bothering, and also avoid mistakenly causing issues with replacement.

jakebailey: Example:

image

jakebailey:
Randomly looking at parts of the diff, I see a few issues:

```
    /**
     * The **`track`** read-only property of the HTMLTrackElement interface returns a TextTrack object corresponding to the text track of the <track> element.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLTrackElement/track)
     */
```

Says `<track>`. Same sort of thing happens on `HTMLVideoElement` too:

image


image

image


Technically, these need to be escaped. At some level it feels like backticks would be easier to deal with. Not sure.

Reading the input file, it doesn't seem like there's much one can do to improve this, given it seems to literally just be strings? Not consistent enough to parse as HTML or something either.

src/build/mdn-comments.ts
saschanaz: > I'm going through the diff and there are places we are now making bold that did not used to be.

Most likely that's MDN mistakenly not styling the pages properly. This way we get less diffs so that we can be sure we are not breaking anything.

src/build/mdn-comments.ts
```diff
@@ -117,37 +66,43 @@ function insertComment(
...
 
+function generateComment(summary: string, name: string): string {
+  // Escape special regex characters in the name
+  const escapedName = name.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+  let result = summary.replace(/\n/g, " "); // remove newlines
+
+  for (const [word, formatted] of Object.entries(syntax)) {
+    result = result.replace(new RegExp(`\\b${word}\\b`, "g"), formatted);
+  }
```
saschanaz, line 68: But I'd skip this one, as they are rarer.

Bashamega: Done

Bashamega:
Done @saschanaz 
Is it ready now?

src/build/mdn-comments.ts
```diff
@@ -117,37 +58,37 @@ function insertComment(
...
 
+function generateComment(summary: string, name: string): string {
+  // Escape special regex characters in the name
+  const escapedName = name.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+  return summary
+    .replace(/\n/g, " ") // remove newlines
+    .replace(
+      // Match optional preceding identifier + dot OR just the name itself
+      new RegExp(`(?:\\b\\w+\\.)?${escapedName}(\\(\\))?`),
```
jakebailey: I'm honestly wondering if any of these replacements are worth it. (Have not tested locally, though, might look at this tomorrow, not asking for a change yet)

src/build/mdn-comments.ts
```diff
@@ -117,37 +58,37 @@ function insertComment(
...
+      (match) => `**\`${match}\`**`,
+    )
+    .replace(/"/g, "'")
```
saschanaz: Why the quote replacement?

src/build/mdn-comments.ts
```diff
@@ -117,37 +58,37 @@ function insertComment(
   }
 }
 
+function generateComment(summary: string, name: string): string {
+  // Escape special regex characters in the name
+  const escapedName = name.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
saschanaz: Do we ever get these characters in the name? Can we throw instead if those are included?

src/build/mdn-comments.ts
Bashamega: This way, it will remove the markdown, which you are the one who asked for btw

jakebailey: I'm just trying to find the balance between consistency and diff size 😞

src/build/mdn-comments.ts
```diff
@@ -117,37 +58,37 @@ function insertComment(
...
-  );
+  // metadata is an array of objects, each with at least: slug, page-type, summary
+  for (const entry of mdn.data) {
+    const mdnUrl = entry.mdn_url.split("/en-US/docs/")[1];
+    const slugArr = extractSlug(mdnUrl);
+    const path = paths[entry.pageType];
+    const comment = generateComment(entry.summary, slugArr[slugArr.length - 1]);
+    if (!slugArr.length || !path) {
+      continue;
+    }
```
```suggestion
    if (!slugArr.length || !path) {
      continue;
    }
    const comment = generateComment(entry.summary, slugArr.at(-1)!);
```

slugArr length check should happen first, and let's use the modern `at()`.

src/build/mdn-comments.ts
saschanaz: I see dot is sometimes included but not others, so something like:

```suggestion
  if ([ ... list of invalid characters ...]).some(c => name.includes(c))) {
    throw new Error(`Invalid name: ${name}`);
  }
  const escapedName = name.replaceAll(".", "\\.");
```

src/build/mdn-comments.ts
Bashamega: I think it good option is to file a PR to fix the mismatch, If you like this idea, I can draft a Pr for MDN

jakebailey:
FWIW this all seems good to me. Whenever @saschanaz's comments are resolve and the PR is up to date with main (we stop changing main...) I think this can go in so we can drop the submodule.

src/build/mdn-comments.ts

src/build/mdn-comments.ts

Bashamega:
I have updated @jakebailey 


Can you merge it?

jakebailey (approved):
This missed the submodule line from deploy.yml but it's not a big deal

scripts/fetch-mdn.js
```diff
@@ -0,0 +1,35 @@
...
+  .map(({ mdn_url, pageType, summary }) => ({
+    mdn_url,
+    pageType,
+    summary,
+  }));
+
+// Save to file
+await fs.writeFile(
+  new URL("../inputfiles/mdn.json", import.meta.url),
+  JSON.stringify({ data: filtered }, null, 2),
```
saschanaz: Can we skip `data` and do:

```suggestion
  JSON.stringify(filtered, null, 2),
```

?

package.json
```diff
@@ -14,8 +14,8 @@
   "scripts": {
     "build": "tsc && node ./lib/build.js",
     "baseline-accept": "cpx \"generated\\**\" baselines\\",
-    "lint": "eslint --max-warnings 0 src deploy/*.js && tsc -p deploy/jsconfig.json",
-    "lint-fix": "eslint --max-warnings 0 src deploy/*.js --fix",
+    "lint": "eslint --max-warnings 0 src deploy/*.js scripts/*.js && tsc -p deploy/jsconfig.json",
+    "lint-fix": "eslint --max-warnings 0 src deploy/*.js scripts/* --fix",
```
saschanaz:
```suggestion
    "lint": "eslint --max-warnings 0 src deploy/*.js scripts && tsc -p deploy/jsconfig.json",
    "lint-fix": "eslint --max-warnings 0 src deploy/*.js scripts --fix",
```

src/build/mdn-comments.ts
```diff
@@ -117,37 +58,41 @@ function insertComment(
...
-      if (!slug.length || !generatedPath) {
-        return;
-      }
-
-      const summary = extractSummary(content);
-      insertComment(results, slug, summary, generatedPath);
-    }),
-  );
+  // metadata is an array of objects, each with at least: slug, page-type, summary
+  for (const entry of mdn.data) {
```
```suggestion
  for (const entry of mdn) {
```

saschanaz:
Err, can't update the json file as I have no push permission. Can either of you update the file?

src/build/mdn-comments.ts
```diff
@@ -1,65 +1,11 @@
-import fs from "fs/promises";
-const basePath = new URL("../../inputfiles/mdn/files/en-us/", import.meta.url);
+import { readFile } from "fs/promises";
+const inputFile = new URL("../../inputfiles/mdn.json", import.meta.url);
```
saschanaz, line 3:
```suggestion

const inputFile = new URL("../../inputfiles/mdn.json", import.meta.url);
```

src/build/mdn-comments.ts
```diff
@@ -1,65 +1,12 @@
-import fs from "fs/promises";
-const basePath = new URL("../../inputfiles/mdn/files/en-us/", import.meta.url);
+import { readFile } from "fs/promises";
+
+const inputFile = new URL("../../inputfiles/mdn.json", import.meta.url);
+// These are the subdirectory prefixes we care about for slugs
```
```suggestion

// Valid subdirectories for our use case
```

src/build/mdn-comments.ts
```diff
@@ -74,20 +22,15 @@ const paths: Record<string, string[]> = {
...
-  const pageType = content.match(/\npage-type: (.+)\n/)!;
-  const type = pageType[1];
-  return paths[type];
-}
-
-function extractSlug(content: string): string[] {
-  const match = content.match(/\nslug: (.+)\n/)!;
-  const url = match[1].split(":").pop()!;
-  const normalized = url.endsWith("_static") ? url.slice(0, -7) : url;
+function extractSlug(slug: string): string[] {
```
```suggestion
function extractSlug(mdnUrl: string): string[] {
```

src/build/mdn-comments.ts
```diff
@@ -74,20 +22,15 @@ const paths: Record<string, string[]> = {
...
   for (const subdirectory of subdirectories) {
-    if (normalized.toLowerCase().startsWith(subdirectory)) {
-      return normalized.slice(subdirectory.length).split("/");
+    if (!slug.toLowerCase().startsWith(subdirectory)) {
```
```suggestion
    if (!mdnUrl.toLowerCase().startsWith(subdirectory)) {
```

src/build/mdn-comments.ts
```diff
@@ -74,20 +22,15 @@ const paths: Record<string, string[]> = {
...
+      continue;
     }
+    return slug
```
```suggestion
    return mdnUrl
```

jakebailey:
How are you able to apply changes to the PR?

saschanaz:
> How are you able to apply changes to the PR?

With a hack - @Bashamega also opened https://github.com/saschanaz/types-web/pull/241 and that grants me some sort of access. But only through GitHub UI and no direct git push.

jakebailey:
Interesting, did not know that GitHub allowed a branch to be open to two PRs even across repos.

Bashamega:
Thanks for helping @saschanaz, I will pull and fix the code right now, give me 5-10 minutes

src/build/mdn-comments.ts
```diff
@@ -117,37 +60,41 @@ function insertComment(
   }
 }
 
+function generateComment(summary: string, name: string): string | undefined {
+  // Escape special regex characters in the name
+  if (["."].some((c) => name.includes(c))) {
```
```suggestion
  // Ban any non-alphanumeric characters in the name for safe regex
  // For now the only known exception is `RTCStatsReport/Symbol.iterator`.
  if (name.match(/\W/)) {
```

saschanaz:
> Thanks for helping @saschanaz, I will pull and fix the code right now, give me 5-10 minutes

Cool, I'll stop making changes until then.

Bashamega:
I have updated the MDN file @saschanaz 
Is there anything else you want me to update?

Bashamega:
> > How are you able to apply changes to the PR?
> 
> With a hack - @Bashamega also opened [saschanaz#241](https://github.com/saschanaz/types-web/pull/241) and that grants me some sort of access. But only through GitHub UI and no direct git push.

I think the best thing to solve this issue with permission to invite @saschanaz as a collaborator for this repository :)

saschanaz:
Apparently Microsoft didn't like that idea 🥺

jakebailey:
I have a plan to fix the old workflow; I think I just over-classified the repo which locked out some things. (Will take a bit for the system to update repo rules.)

Bashamega:
I hope you can gain access

Bashamega:
Also, for the future if you don't mind @saschanaz 
Do you have a faster way of communicating for example, Discord or Slack?

jakebailey:
There's a dom lib channel on the discord, fwiw

jakebailey:
Everything in order now?

(This is a maintained tagged PR so I have to merge it anyway, just let me know)

Bashamega:
> There's a dom lib channel on the discord, fwiw

I know but I think @saschanaz is not on discord or inactive

saschanaz:
I think mostly fine but I'll check again in 30min

saschanaz:
Actually, LGTM

Bashamega:
Thanks

jakebailey:
I'll check back then if I'm still awake (my fire alarm went off in the middle of the night, I'm not normally around at this time...)

jakebailey:
Oh, good

PR #2173, opened 2025-09-30
Remove unneeded command maps

Bashamega:
These are extras that we don't need


saschanaz:
Those are not IDL files but command maps. Please change the title when merging. Otherwise fine for me.

jakebailey (approved):
Updated the title. Merge at will 😄

saschanaz:
LGTM, thanks!

PR #2174, opened 2025-09-30
Introduce CSSStyleDeclarationBase

saschanaz:
This should reduce DT breakage. But note that this would make `CSSPageDescriptors extends CSSStyleDeclaration` to `false`. Authors would need to use CSSStyleDeclarationBase for that.

### DIFF
--- a/inputfiles/patches/cssom.kdl
+++ b/inputfiles/patches/cssom.kdl
@@ -1,0 +1,7 @@
+// CSSStyleDeclaration spec change broke a bunch of DT packages.
+// This tries to make existing usages work while still keeping the new
+// CSSStyleProperties.
+
+interface CSSStyleDeclaration \
+  forward=CSSStyleDeclarationBase \
+  forwardExtends=CSSStyleProperties
--- a/inputfiles/patches/cssom.kdl
+++ b/inputfiles/patches/cssom.kdl
@@ -5,3 +5,5 @@
 interface CSSStyleDeclaration \
   forward=CSSStyleDeclarationBase \
   forwardExtends=CSSStyleProperties
+
+interface CSSStyleProperties replaceReference=CSSStyleDeclaration
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -765,7 +765,7 @@ export function emitWebIdl(
   }
 
   function emitCallBackInterface(i: Browser.Interface) {
-    const methods = mapToArray(i.methods.method);
+    const methods = mapToArray(i.methods?.method);
     const m = methods[0];
     const overload = m.signature[0];
     const paramsString = overload.param ? paramsToString(overload.param) : "";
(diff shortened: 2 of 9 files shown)


jakebailey:
CSSStyleDeclaration doesn't add anything to CSSStyleProperties so I think it technically still works?

saschanaz:
With this PR it's CSSStyleDeclarationBase -> CSSStyleProperties -> CSSStyleDeclaration, so yeah it can't add anything to it's now-parent.

image

saschanaz:
@jakebailey 🤞🏻

inputfiles/patches/cssom.kdl
```diff
@@ -0,0 +1,9 @@
+// CSSStyleDeclaration spec change broke a bunch of DT packages.
+// This tries to make existing usages work while still keeping the new
+// CSSStyleProperties.
```
Bashamega, line 3: Why is it on a new line?

src/build/emitter.ts
```diff
@@ -472,15 +472,18 @@ export function emitWebIdl(
...
+    }
+    // Name of an alias / callback / enum / dict. Just return itself
     if (
-      allInterfacesMap[objDomType] ||
       allLegacyWindowAliases.includes(objDomType) ||
       allCallbackFunctionsMap[objDomType] ||
       allDictionariesMap[objDomType] ||
       allEnumsMap[objDomType]
-    )
+    ) {
```
Bashamega, line 484: Why the extra brackets?

src/build/types.ts
```diff
@@ -151,11 +151,17 @@ export interface Interface {
...
+  // "extends" for the original interface, in case `forward: true`.
+  forwardExtends?: string;
+  // replaces references from methods and properties
+  replaceReference?: string;
   comment?: string;
   constants?: {
     constant: Record<string, Constant>;
   };
-  methods: {
+  methods?: {
```
Bashamega, line 164: I don't see why we need this update, and in the widlprocess.ts file, specifically in this PR. Maybe in a separate PR

Bashamega:
Just a few nits that I found

inputfiles/patches/cssom.kdl
saschanaz: Wrapping has always been industrial standard

src/build/emitter.ts
saschanaz: Because somebody omitted it and I don't like it. Can even add an eslint rule...

src/build/types.ts
saschanaz: Because otherwise the emitter should pass `methods: { method: {} }` for no good reason.

inputfiles/patches/cssom.kdl
Bashamega: Sure, I didn't know

saschanaz: Text editors automatically wrap too, so it's not strictly required.

src/build/emitter.ts
Bashamega: Well, it is also in line 488, so maybe a separate PR, to update it everywhere, with an eslint rule

saschanaz: I'd accept such PR.

src/build/types.ts
Bashamega: k

src/build/emitter.ts
Bashamega: I will open a PR for it this week, so you can undo it here

saschanaz: This is a single line change nearby the actual change, which is fine here

jakebailey (approved):
Based on #62496, it seems like this is working fine. If you're happy, then I think this is okay.

saschanaz:
LGTM, let's go 🙂

PR #2183, opened 2025-10-01
Make URLPatternResult properties mandatory

Renegade334:
One of those cases where a "return value" dictionary is being generated as an interface with optional properties.

These aren't explicitly marked as `required` in the IDL, but are mandatorily assigned in the URLPattern spec. Having these as optional properties in TS is just going to be a usability headache.

Refs: https://urlpattern.spec.whatwg.org/#url-pattern-match
Refs: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73799#discussion_r2396093936

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2384,15 +2384,15 @@ interface URLPatternOptions {
 }
 
 interface URLPatternResult {
-    hash?: URLPatternComponentResult;
-    hostname?: URLPatternComponentResult;
-    inputs?: URLPatternInput[];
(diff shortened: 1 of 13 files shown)


jakebailey:
Does `URLPatternComponentResult` need to also have this done to it?

Renegade334:
[Yarp](https://urlpattern.spec.whatwg.org/#create-a-component-match-result). Good spot.

saschanaz:
LGTM, thanks!

jakebailey (approved):

PR #2187, opened 2025-10-04
Enhance optionalMember function to accept multiple types

Bashamega:
#2053

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -400,7 +400,6 @@
                 "name": "StyleMedia",
                 "exposed": "Window",
                 "noInterfaceObject": true,
-                "deprecated": true,
                 "properties": {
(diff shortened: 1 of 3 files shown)


inputfiles/addedTypes.jsonc
saschanaz, line 398: So this was added in https://github.com/saschanaz/types-web/pull/113, back then I accepted it to make things easier, but not sure this is still needed. Can we update React to get their own definition at least? @jakebailey 

For deprecated=true, given this is a manual deprecation, we should probably figure out some actual deprecation message than just `true`.

jakebailey: I think the react types forward declare this interface, so it may not be that big of a deal to remove. If you send a PR here with your proposed change, I can pull that into a TS PR to check on the extended test suite.

PR #2189, opened 2025-10-05
Add curly rule to ESLint configuration

Bashamega:


saschanaz:
I'm happy with this.

HolgerJeromin (approved):

jakebailey:
I thought these rules had been removed from eslint proper? I guess this repo is very behind on linting versions?

saschanaz:
Hmm? It's a thing? https://eslint.org/docs/latest/rules/curly

HolgerJeromin:
> I thought these rules had been removed from eslint

`curly` is not part of the stylistics rules which were migrated to another team.

https://eslint.style/rules

jakebailey:
Interesting choice to only get rid of _some_ style rules...

jakebailey:
Personally I think doing prettier or dprint would be better solutions, but, it's all the same

jakebailey (approved):

HolgerJeromin:
ref:
https://www.joshuakgoldberg.com/blog/the-blurry-line-between-formatting-and-style/

saschanaz:
We do use prettier already, except it didn't say anything here.

HolgerJeromin:
We could add the plugin:
https://github.com/JoshuaKGoldberg/prettier-plugin-curly

saschanaz:
If we used prettier only, yeah. Not much point if we have eslint IMO.

PR #2192, opened 2025-10-07
Remove redundant SVGStyleElement override

Renegade334:
This is now in the IDL, and the emitter is warning about the duplicate property.


jakebailey:
Er, were there supposed to be d.ts changes? Where are you getting the error?

Renegade334:
> Er, were there supposed to be d.ts changes? Where are you getting the error?

The property is getting emitted correctly, so no change to the output, but the build script is warning about the duplication.

PR #2193, opened 2025-10-08
Emit `Symbol.toStringTag` properties in TS 6.0

### LINKED ISSUE #1641: `Symbol.toStringTag` is missing on `File`, `Blob` and `FormData`

### LINKED ISSUE #2172: [Web API type definition issue] WebAssembly.Module interface is empty, allowing invalid assignments
### Summary

The `WebAssembly.Module` interface is currently emitted as an empty interface, making it structurally compatible with any object in TypeScript

### Expected vs. Actual Behavior

```ts
const abc: WebAssembly.Module = 0;
```

I would expect the above to be a compilation error.

But `WebAssembly.Module` gets emitted as an empty interface allowing all assignments.

```ts
interface Module { }
```

### Playground Link

https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAhgI2ALhgdQKYIIIQhgWwQBsBPAOgFkQATAV2IxgF4YAGAbiA

### Browser Support

- [x] This API is supported in at least two major browser engines (not two Chromium-based browsers).

### Have Tried The Latest Releases

- [x] This issue applies to the latest release of TypeScript.
- [x] This issue applies to the latest release of `@types/web`.

### Additional Context

Do we need to make `WebAssembly.Module` nominally typed? 🫣 

I'm not aware of any other interfaces with only constructors/static methods but no instance members

Renegade334:
Fixes #1641.
Fixes #2172.
Alternative to #1762.
Supersedes #2075.

Reference: https://webidl.spec.whatwg.org/#dfn-class-string

This amends the builder to append `Symbol.toStringTag` properties to interfaces and iterators in the ts6.0 core libraries. (ES6's well-known symbols are universally defined in ts6.0.)

Because adding a property with a literal type affects inheritance chains, only "final" classes (ie. those that do not have any inheritance children within the current global) can have these properties typed as string literals. (For example, it's not possible for `Blob#[Symbol.toStringTag]` to be typed as `"Blob"`, because File inherits from Blob, and a property of type `"File"` can't override a parent property of type `"Blob"`.)

For those interfaces that have one or more inheritance children, there are two possible approaches.
- One is not to emit the property at all – but it's inconsistent to have some interfaces with `Symbol.toStringTag` properties, and others (including some very common ones) without them.
- The alternative is to emit properties for these as well, but type them permissively as `string`, so that they can be correctly overridden by child interfaces. This is the approach taken here.

The toStringTag emitter does not add to any interface marked `noInterfaceObject`, since these are usually pseudo-types that are not true IDL interfaces, nor to namespaces converted to interfaces (ie. `console`).

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -315,6 +315,7 @@ interface AudioWorkletProcessor {
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/AudioWorkletProcessor/port)
      */
     readonly port: MessagePort;
+    readonly [Symbol.toStringTag]: string;
 }
 
 declare var AudioWorkletProcessor: {
@@ -678,6 +679,7 @@ interface Event {
     readonly CAPTURING_PHASE: 1;
     readonly AT_TARGET: 2;
     readonly BUBBLING_PHASE: 3;
+    readonly [Symbol.toStringTag]: string;
 }
 
 declare var Event: {
@@ -721,6 +723,7 @@ interface EventTarget {
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/EventTarget/removeEventListener)
      */
     removeEventListener(type: string, callback: EventListenerOrEventListenerObject | null, options?: EventListenerOptions | boolean): void;
+    readonly [Symbol.toStringTag]: string;
 }
 
 declare var EventTarget: {
@@ -1440,6 +1443,7 @@ declare var URLSearchParams: {
  * [MDN Reference](https://developer.mozilla.org/docs/Web/API/WorkletGlobalScope)
  */
 interface WorkletGlobalScope {
+    readonly [Symbol.toStringTag]: string;
 }
 
 declare var WorkletGlobalScope: {
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2831,6 +2831,7 @@ interface AbortController {
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/AbortController/abort)
      */
     abort(reason?: any): void;
+    readonly [Symbol.toStringTag]: string;
 }
 
 declare var AbortController: {
@@ -2872,6 +2873,7 @@ interface AbortSignal extends EventTarget {
     addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
     removeEventListener<K extends keyof AbortSignalEventMap>(type: K, listener: (this: AbortSignal, ev: AbortSignalEventMap[K]) => any, options?: boolean | EventListenerOptions): void;
     removeEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | EventListenerOptions): void;
+    readonly [Symbol.toStringTag]: string;
 }
 
 declare var AbortSignal: {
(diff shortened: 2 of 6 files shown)


inputfiles/overridingTypes.jsonc
```diff
@@ -1725,7 +1725,8 @@
                 }
             },
             "DOMException": {
-                "extends": "Error"
+                "extends": "Error",
```
Bashamega: Move DOMException to a .kdl file, maybe move it to `dom.kdl`

Bashamega (changes requested):

inputfiles/overridingTypes.jsonc
Renegade334: Moving the existing override is really beyond the scope of this changeset; would be better suited to a follow-up PR.

Bashamega: > Moving the existing override is really beyond the scope of this changeset; would be better suited to a follow-up PR.

Then remove the changes in patches.ts

saschanaz: But why noToStringTag at all?

Renegade334: `Error` doesn't have a toStringTag property in the TS libs, and DOMException is a common target for userland inheritance, so the concern is that this will do more harm than good.

saschanaz: Perhaps we should test it in TS repo without noToStringTag - and if it gives some trouble then we should consider it. Right now it feels arbitrary.

(idk much about testing in TS repo, @jakebailey knows more)

jakebailey: It's pretty easy to check these variants (though a TS team member has to send the PR as we auto close anyone else trying to send a PR updating these files due to people not understanding they shouldn't touch generated files in the repo 😄)

Renegade334: I've rebased and re-enabled tagging for `DOMException` – this should be in a position to go through the user tests.

saschanaz: I guess Jake needs to make such PR then?

jakebailey: Yes, getting there 😄

jakebailey: https://github.com/microsoft/TypeScript/pull/62682

jakebailey: The results on the above are pretty bad.

Renegade334: A significant proportion are intentional errors that are now differentiating between discrete DOM interfaces that are empty or identically-shaped, but looks like most of the remainder are assignability errors between the lib.dom interfaces and userland DOM-like interfaces that are unlikely to go down well. (It's worth saying that those same errors would also be triggered by those interfaces having _any_ new props/methods added, not just toStringTag, but adding an additional key to every single interface would turbocharge the breakage.)

saschanaz: I think we can start small and merge the logic in #2075 to add it only for empty interfaces for now. Thoughts?

jakebailey: When I'm back in town in the new year I'll be happy to try another variant

Renegade334: I'm agnostic. I don't think there'll be much appetite to roll this out universally. Having this property on some interfaces and not others would be a tad inconsistent, but better than defining a non-existent property to brand the object...

saschanaz:
@Renegade334 do you plan to revive this per https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2193#discussion_r2636444845 ? 👀

Renegade334:
I think that the verdict from microsoft/TypeScript#62682 was "ouch".

saschanaz:
Yeah but that doesn't mean we can't start small 👀

PR #2204, opened 2025-10-25
Update messageerror event's type in service worker to ExtendableMessageEvent

dhurlburtusa:
According to https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/messageerror_event, the type of the `messageerror` event in a service worker is supposed to be `ExtendableMessageEvent` instead of just `MessageEvent`. That is, it is missing the `waitUntil` method.


dhurlburtusa:

saschanaz:
This looks correct, thanks! LGTM

saschanaz:
LGTM

PR #2205, opened 2025-10-25
Add securitypolicyviolation event to WorkerGlobalScope

dhurlburtusa:
According to https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/securitypolicyviolation_event, the `WorkerGlobalScope` should have a `securitypolicyviolation` event of type `SecurityPolicyViolationEvent`.

I'd rather not continue adding to the `addedTypes.jsonc` file address issues.

I took a look at https://github.com/mdn/browser-compat-data. I don't see anything obvious in:
- https://github.com/mdn/browser-compat-data/blob/main/api/WorkerGlobalScope.json#L549
- https://github.com/mdn/browser-compat-data/blob/main/api/SecurityPolicyViolationEvent.json
Maybe I should be looking elsewhere.

I don't see it in `removedTypes.jsonc` or `overridingTypes.jsonc`.

For some reason, other events like the `error` event of the `WorkerGlobalScope` show up in the generated .d.ts files and there is nothing in `addedTypes.jsonc` to explicitly add it.

Note: The `SecurityPolicyViolationEvent` type already exists.

Please guide me if there is a better way to implement this fix.

And in the previous PR I submitted (https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2204), I noticed that there seems to be a bunch of stuff in `overridingTypes.jsonc` that no longer seem to be necessary. That is, I was able to remove some entries, regenerate the .d.ts files, and Git would report no changes to the .d.ts files.


saschanaz:
This is wrong, see also #2124 for the context.

dhurlburtusa:
Thanks. I'll take a look at that.

dhurlburtusa:
I guess MDN is probably correct most of the time but sometimes they aren't.

PR #2206, opened 2025-10-26
Fix overrides for non-shared ArrayBuffer views

Renegade334:
Fixes a couple of divergences from the specifications:

- `WebSocket#send()` was previously overridden back in 2018 to accept shared array buffers and array buffer views. The parameter isn't `[AllowShared]` in the specification, so this is simply inaccurate.
- The `extends` clause of a couple of interface methods is currently set to `ArrayBufferView` when it should be `ArrayBufferView<ArrayBuffer>`. Because `extends` is a manually defined field, it doesn't honour useGenericTypedArrays, so this needs to be indirectly resolved for compatibility with TS 5.6 and earlier.

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -959,7 +959,7 @@ interface ReadableStreamBYOBReader extends ReadableStreamGenericReader {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ReadableStreamBYOBReader/read)
      */
-    read<T extends ArrayBufferView>(view: T, options?: ReadableStreamBYOBReaderReadOptions): Promise<ReadableStreamReadResult<T>>;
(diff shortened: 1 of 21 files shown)


saschanaz:
Interesting, LGTM

jakebailey (approved):
Hope this doesn't break people even more 😄

saschanaz:
LGTM

saschanaz:
What? 🙄 lgtm

jakebailey:
I rotated a bunch of tokens, maybe I broke this one. Will fix soon

jakebailey:
Can you try again?

saschanaz:
LGTM

jakebailey:
Okay, well that's quite confusing.

jakebailey:
The token I rotated isn't even the one used by this workflow, which is just the regular github actions token. Weird. Maybe they broke yet another thing in the workflow :(

I'll just merge it.

saschanaz:
This is why I need a real permission rather than a workflow workaround

PR #2217, opened 2025-11-01
Refactor CSS utility functions and remove unused files

Bashamega:


src/build/webref/css.ts
```diff
@@ -1,17 +0,0 @@
-function hyphenToCamelCase(name: string) {
-  const camel = name
-    .replace(/^-(\w)/, (_, c) => c)
-    .replace(/-(\w)/g, (_, c) => c.toUpperCase());
-  return camel === "float" ? "_float" : camel;
-}
-
-export function generateWebIdlFromCssProperties(properties: string[]): string {
```
saschanaz, line 8: This is webref specific and I'd rather keep it here. Maybe import the case conversion function?

Bashamega: Done

saschanaz: Eh, actually, the utils/css module is totally unused. Let's remove that and keep this one.

Bashamega: I have updated it @saschanaz

src/build/webref/css.ts
```diff
@@ -1,10 +1,15 @@
-function hyphenToCamelCase(name: string) {
+export function hyphenToCamelCase(name: string): string {
```
saschanaz: No need to export, right? 
```suggestion
function hyphenToCamelCase(name: string): string {
```

src/build/webref/css.ts
```diff
@@ -1,10 +1,15 @@
...
   const camel = name
     .replace(/^-(\w)/, (_, c) => c)
     .replace(/-(\w)/g, (_, c) => c.toUpperCase());
   return camel === "float" ? "_float" : camel;
 }
 
+export function camelToHyphenCase(name: string): string {
```
saschanaz: Doesn't seem like this is used anywhere.

Bashamega: Done

jakebailey (dismissed):

saschanaz:
#2218 started to use hyphenToCamelCase, merging this would break it

Bashamega:
> #2218 started to use hyphenToCamelCase, merging this would break it

I will later update it in the other PR. You can merge this now

saschanaz:
Why remove it when we need it?

Bashamega:
I have added it @saschanaz

src/build/webref/css.ts
```diff
@@ -1,4 +1,4 @@
-function hyphenToCamelCase(name: string) {
+export function hyphenToCamelCase(name: string): string {
```
saschanaz: I thought only webref would use it, thus the build/webref/css. If not it should be in build/utils/css.

Bashamega: done

Bashamega:
Is there anything missing here @saschanaz

src/build/webref/css.ts
```diff
@@ -1,8 +1,7 @@
-function hyphenToCamelCase(name: string) {
-  const camel = name
-    .replace(/^-(\w)/, (_, c) => c)
-    .replace(/-(\w)/g, (_, c) => c.toUpperCase());
-  return camel === "float" ? "_float" : camel;
+import { hyphenToCamelCase } from "../utils/css.js";
+export function camelToHyphenCase(name: string): string {
```
saschanaz: We don't use this one, right?

Bashamega: Yup, I have updated it

saschanaz:
LGTM

Bashamega:
Did you forget this @jakebailey

PR #2218, opened 2025-11-01
Comments for css

Bashamega:


src/build/mdn-comments.ts
```diff
@@ -1,14 +1,19 @@
 import { readFile } from "fs/promises";
+import { hyphenToCamelCase } from "./utils/css.js";
 
 const inputFile = new URL("../../inputfiles/mdn.json", import.meta.url);
 
 // Valid subdirectories for our use case
 const subdirectories = [
   "web/api/",
   "webassembly/reference/javascript_interface/",
+  "web/",
```
saschanaz: Did you intend to say web/css?

Bashamega: No, I intended for web/ because the script needs the CSS part

baselines/dom.generated.d.ts
```diff
@@ -6065,1163 +6065,2933 @@ declare var CSSStyleDeclaration: {
 };
 
 /**
- * The **`CSSStyleProperties`** interface of the CSS Object Model (CSSOM) represents inline or computed styles available on an element, or the styles associated with a CSS style rule.
+ * The CSS interface holds useful CSS-related methods. No objects with this interface are implemented: it contains only static methods and is therefore a utilitarian interface.
```
saschanaz: This looks wrong?

src/build/mdn-comments.ts
saschanaz: That's `web/css/`, no? fetch-mdn filters web/css after all.

src/build/mdn-comments.ts
```diff
@@ -27,10 +32,20 @@ function extractSlug(mdnUrl: string): string[] {
...
     }
-    return mdnUrl
+    let slugArr = mdnUrl
       .slice(subdirectory.length)
       .replace(/_static/g, "")
-      .split("/");
+      .split("/")
+      .map((part) =>
+        hyphenToCamelCase(
+          part.toLowerCase() === "css" ? "CSSStyleProperties" : part,
```
saschanaz: This looks wrong, CSS is not CSSStyleProperties.

baselines/dom.generated.d.ts
```diff
@@ -38277,13 +40131,8 @@ declare var XSLTProcessor: {
     new(): XSLTProcessor;
 };
 
-/** The **`CSS`** interface holds useful CSS-related methods. No objects with this interface are implemented: it contains only static methods and is therefore a utilitarian interface. */
 declare namespace CSS {
-    /**
-     * The static, read-only **`highlights`** property of the CSS interface provides access to the HighlightRegistry used to style arbitrary text ranges using the CSS Custom Highlight API.
```
saschanaz, line 38285: And that caused loss of comments.

src/build/mdn-comments.ts
```diff
@@ -1,14 +1,19 @@
...
 // Valid subdirectories for our use case
 const subdirectories = [
   "web/api/",
   "webassembly/reference/javascript_interface/",
+  "web/",
 ];
 
 const paths: Record<string, string[]> = {
+  "css-property": ["properties", "property"],
+  "css-type": ["properties", "property"],
```
saschanaz: We probably don't need this one, it's for CSS syntax. We don't handle this down there after all.

src/build/mdn-comments.ts
```diff
@@ -27,10 +32,20 @@ function extractSlug(mdnUrl: string): string[] {
     if (!mdnUrl.toLowerCase().startsWith(subdirectory)) {
```
saschanaz: We should stop lowercasing, btw. Should be a separate PR though.

baselines/dom.generated.d.ts
Bashamega: Fixed

src/build/mdn-comments.ts
Bashamega: What do you recommend using instead @saschanaz

baselines/dom.generated.d.ts
Bashamega: Fixed

test
```diff
@@ -0,0 +1,12 @@
+
```
saschanaz: (remove this)

Bashamega: I added it by mistake. I have removed it

src/build/mdn-comments.ts
```diff
@@ -27,10 +32,18 @@ function extractSlug(mdnUrl: string): string[] {
...
       .slice(subdirectory.length)
       .replace(/_static/g, "")
-      .split("/");
+      .split("/")
+      .map((part) =>
+        hyphenToCamelCase(part === "Reference" ? "CSSStyleProperties" : part),
+      );
+    if (slugArr[0] === "CSSStyleProperties") {
+      // Drop the second itmem
+      slugArr.splice(1, 1);
```
saschanaz: I was confused why this is needed at all, it seems it's because it's now `Web/CSS/Reference/Properties/` as we saw in #2216. Then maybe the subdirectory should be that. (And maybe we should stop lowercasing before doing that)

src/build/mdn-comments.ts

src/build/mdn-comments.ts
```diff
@@ -96,6 +108,22 @@ export async function generateDescriptions(): Promise<{
       continue;
     }
     insertComment(results, slugArr, comment, path);
+    // Check if this is a CSS property and its slug is in the webkit list
```
saschanaz: I'd skip this, as this way:

1. If MDN has a separate page for -webkit- property, we lose that
2. Otherwise we add a potentially wrong comment, e.g. we say "The flex property ..." for -webkit-flex, which is wrong. And there's a risk that the webkit property may have a different syntax than the standard one.

saschanaz (commented):
I have a few changes I'd like to make, if you don't mind can you open a PR on my fork?

src/build/mdn-comments.ts
Bashamega: But this way we will loose a lot of comments.
I will open a PR on your fork right now

Bashamega: I have removed them @saschanaz

saschanaz: That's technically not a "loss" as they would be incorrect and they don't exist right now.

scripts/fetch-mdn.js
```diff
@@ -17,7 +17,10 @@ const filtered = Object.values(data)
     const path = entry.mdn_url.toLowerCase();
     return (
       path.startsWith("/en-us/docs/web/api/") ||
-      path.startsWith("/en-us/docs/webassembly/reference/javascript_interface/")
+      path.startsWith(
+        "/en-us/docs/webassembly/reference/javascript_interface/",
+      ) ||
+      path.startsWith("/en-us/docs/web/css/")
```
saschanaz:
```suggestion
      path.startsWith("/en-us/docs/web/css/reference/properties/")
```

src/build/mdn-comments.ts
```diff
@@ -1,14 +1,18 @@
 import { readFile } from "fs/promises";
+import { hyphenToCamelCase } from "./utils/css.js";
 
 const inputFile = new URL("../../inputfiles/mdn.json", import.meta.url);
 
 // Valid subdirectories for our use case
 const subdirectories = [
   "web/api/",
   "webassembly/reference/javascript_interface/",
+  "web/css/",
```
```suggestion
  "web/css/reference/properties/",
```

src/build/mdn-comments.ts
```diff
@@ -1,14 +1,18 @@
...
 ];
 
 const paths: Record<string, string[]> = {
+  "css-property": ["properties", "property"],
+  "css-shorthand-property": ["properties", "property"],
```
```suggestion
  "css-property": ["CSSStyleProperties", "properties", "property"],
  "css-shorthand-property": ["CSSStyleProperties", "properties", "property"],
```

saschanaz:
@Bashamega can you regenerate?

Bashamega:
Done @saschanaz

saschanaz:
LGTM

jakebailey:
Needs a merge with main and update after the dep bump

Bashamega:
I have updated it @jakebailey

saschanaz:
LGTM

PR #2226, opened 2025-11-02
Change Dependabot versioning strategy to 'increase'

Bashamega:
Suggested by @jakebailey 
> Wish this thing would update the actual package.json instead of just the lock file...

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,8 @@ updates:
     # Check the npm registry for updates every month
     schedule:
       interval: "monthly"
-    # Usually only bump versions in package-lock.json, but update package.json for major version updates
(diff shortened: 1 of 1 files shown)


saschanaz:
Well, that was intentional because otherwise every deps update would get maintainers tag?

Bashamega:
But the action is broken anyway, @saschanaz.
So you’ll need the maintainer to merge it regardless.
Also, I think you can just add package.json to the CODEOWNERS file.

jakebailey:
I'm hoping the action starts working again in a few days. But I guess it's true that many dep updates are actual content modifying ones and I forgot about that gotcha.

Bashamega:
Should I close this PR?

PR #2227, opened 2025-11-02
Support for multiple method signatures

Bashamega:
#2053

### DIFF
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -7,7 +7,6 @@ import type {
   WebIdl,
   Method,
   Typed,
-  Signature,
   Param,
   Dictionary,
   Member,


baselines/ts5.5/dom.generated.d.ts
```diff
@@ -36146,6 +36146,7 @@ interface WebGLRenderingContextBase {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/WebGLRenderingContext/getError) */
     getError(): GLenum;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/WebGLRenderingContext/getExtension) */
+    getExtension(name: string): any;
```
Bashamega, line 37975: Any idea why this got moved up? @saschanaz

saschanaz: Because this PR is now merging `signature` while originally this was using `additionalSignatures`, which has ordering difference. But the different order doesn't seem to affect TS behavior, while I thought it would matter, or at least it did at some point (https://github.com/microsoft/TypeScript/issues/1860#issuecomment-72154737). Maybe that changed.

@jakebailey, any idea?

src/build/emitter.ts
```diff
@@ -509,6 +509,11 @@ export function emitWebIdl(
       return objDomType;
     }
 
+    // If it's a quoted string literal (has " " inside), just return it
+    if (typeof objDomType === "string" && objDomType.includes('"')) {
```
saschanaz: This implies that something went wrong... This shouldn't be needed.

inputfiles/patches/webgl.kdl
```diff
@@ -1,3 +1,143 @@
 interface-mixin WebGLRenderingContextBase {
   property canvas exposed=Window
+  method getExtension {
+    signatures {
+      _ {
```
saschanaz: This is not great, we should be able to do:

```kdl
method getExtension {
  param extensionName type="\"ANGLE_instanced_arrays\""
  type ANGLE_instanced_arrays nullable=#true
}
method getExtension {
  param extensionName type="\"EXT_blend_minmax\""
  type EXT_blend_minmax nullable=#true
}
```

And that should merge the signatures. One question is that whether it's okay to just use `signature` or it needs to use `additionalSignatures`.

src/build/emitter.ts
Bashamega: This is because the JSON never used quotes in the signature, so this was a missing feature.

inputfiles/patches/webgl.kdl
Bashamega: Sure, I can rewrite it to handle multiple methods, but should I use signature or `additionalSignatures`?

saschanaz: Depends on what Jake says above. For now I don't see a reason to not use `signature` though.

baselines/ts5.5/dom.generated.d.ts
Bashamega: We could flip things around in the emitter.

inputfiles/patches/webgl.kdl
```diff
@@ -1,3 +1,143 @@
...
+        param extensionName type="\"ANGLE_instanced_arrays\""
```
saschanaz: BTW you can also do

```suggestion
        param extensionName type=#""ANGLE_instanced_arrays""#
```

src/build/emitter.ts
saschanaz: Perhaps we should use overrideType for that?

inputfiles/patches/webgl.kdl
Bashamega: Which one would you prefer @saschanaz?

inputfiles/patches/webgl.kdl
Bashamega: I have update it to the new format.

src/build/patches.ts
```diff
@@ -229,6 +234,7 @@ function handleProperty(child: Node): Partial<Property> {
 /**
  * Handles a child node of type "method" and adds it to the method object.
  * @param child The child node to handle.
+ * @param name Optional name for the method.
```
saschanaz: This change is now redundant

src/build/patches.ts
```diff
@@ -265,7 +271,10 @@ function handleMethod(child: Node): Partial<Method> {
         : { type: string(child.properties?.returns) }),
     },
   ];
-  return { name, signature };
+  return {
```
saschanaz, line 353: This too

src/build/patches.ts
```diff
@@ -161,7 +161,12 @@ function handleMixinandInterfaces(
       }
       case "method": {
         const methodName = string(child.values[0]);
-        method[methodName] = handleMethod(child);
+        const m = handleMethod(child);
+        if (method[methodName]) {
+          method[methodName]!.signature!.push(m.signature![0]);
```
saschanaz: Is `!` after `method[methodName]` needed here? TS should know that it's non-null...

src/build/emitter.ts
saschanaz: This is still valid

Bashamega: Why not just add this @saschanaz

saschanaz: Why add this when we have already existing things that work 🤔

`type` is for IDL types, non-IDL types can go to `overrideType`.

saschanaz (commented):
LGTM, assuming @jakebailey is okay with the overload ordering.

Bashamega:
I have fixed the ordering issue @saschanaz

saschanaz:
I don't like the last change and canceling my approval. Let's ask Jake first?

Bashamega:
Sure @saschanaz 
I can revert it if you want

saschanaz:
That will bring my approval back 👍

Bashamega:
I have reverted it @saschanaz

Bashamega:
Hey @saschanaz 
I  wrote some unit tests, and all came back working. We can merge now, I think

saschanaz:
> Let's ask Jake first?

This.

unittests/files/webgl.ts
```diff
@@ -0,0 +1,107 @@
+function assertType<T>(_x: T) {}
+
+const canvas = document.getElementById("canvas") as HTMLCanvasElement;
+const gl = canvas.getContext("webgl")!;
+
+const ext_ANGLE_instanced_arrays = gl.getExtension("ANGLE_instanced_arrays");
+// TypeScript should infer: ANGLE_instanced_arrays | null
+assertType<ANGLE_instanced_arrays | null>(ext_ANGLE_instanced_arrays);
+
+const ext_EXT_blend_minmax = gl.getExtension("EXT_blend_minmax");
```
saschanaz: Meh, one is enough, we don't have to cover every overload.

unittests/files/webgl.ts
```diff
@@ -0,0 +1,107 @@
+function assertType<T>(_x: T) {}
+
+const canvas = document.getElementById("canvas") as HTMLCanvasElement;
+const gl = canvas.getContext("webgl")!;
+
+const ext_ANGLE_instanced_arrays = gl.getExtension("ANGLE_instanced_arrays");
+// TypeScript should infer: ANGLE_instanced_arrays | null
+assertType<ANGLE_instanced_arrays | null>(ext_ANGLE_instanced_arrays);
```
saschanaz, line 8: This won't work if `ext_` is `any`

unittests/files/webgl.ts
Bashamega: Done

src/build/patches.ts
```diff
@@ -264,7 +274,7 @@ function handleMethod(child: Node): Partial<Method> {
         ? handleTyped(typeNode)
         : { type: string(child.properties?.returns) }),
     },
-  ];
+  ] as Method["signature"];
```
saschanaz: Why is this suddenly needed?

Bashamega: Because the types say that type is mandatory, but it is optional

unittests/files/webgl.ts
Bashamega: Can you clarify?
It is never any

src/build/patches.ts
saschanaz: Hmm, you may want to use `DeepPartial<Method>` for handleMethod and use `DeepPartial<Signature>` here (it should be Signature instead of `Method["signature"]`)

unittests/files/webgl.ts
saschanaz: The intention of the test is to check it's not any. But if it becomes any this still won't fail.

saschanaz: Try giving `"ANGLE_illegal"` instead for getExtension.

src/build/patches.ts
Bashamega: Sure

Bashamega:
I have added more types @saschanaz

saschanaz:
Cool, thanks!

jakebailey:
Isn't it critical that the overloads are ordered most specific to least specific?

Bashamega:
> Isn't it critical that the overloads are ordered most specific to least specific?

I think that changed since I wrote tests, and they worked

unittests/files/webgl.ts
jakebailey: Yeah, this isn't a correct way to test this, unfortunately. If this is any, it's legal to cast to anything.

jakebailey:
I would be very surprised if so, given we are supposed to walk top to bottom.

How much work is it to avoid this? Is it challenging? (I am not asking for it to be changed yet)

Bashamega:
I tried to avoid it but @saschanaz didn't like the proposed solution 
This was the solution https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/ae72c10fe7904d708de2dfbea36f086fd272d41c

Another solution is to just use overrideSignature like the original

saschanaz:
> Isn't it critical that the overloads are ordered most specific to least specific?

I thought so but [it works nonetheless](https://www.typescriptlang.org/play/?ssl=4&ssc=27&pln=1&pc=1#code/CYUwxgNghgTiAEAzArgOzAFwJYHtVJxwAoBnALnhIxi1QHMBKCqVATwG4BYAKFElgQp02PAWJgKAIgDKAFQBKkppWq06Xbj0SEiMhUvbwA9EZU16AGngALZNaA). I thought you would know why...

jakebailey:
Utterly weird 
image

jakebailey:
Anyway I guess this PR is fine, but needs a main merge

PR #2235, opened 2025-11-04
Migrated device-motion to KDL

martymcnutt1:
Created with @Bashamega

### DIFF
--- a/inputfiles/patches/device-motion.kdl
+++ b/inputfiles/patches/device-motion.kdl
@@ -1,0 +1,2 @@
+interface DeviceMotionEventAcceleration noInterfaceObject=#true
+interface DeviceMotionEventRotationRate noInterfaceObject=#true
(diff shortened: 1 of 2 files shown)


martymcnutt1:

inputfiles/patches/device-motion.kdl
```diff
@@ -0,0 +1,2 @@
+interface DeviceMotionEventAcceleration noInterfaceObject=#true
```
saschanaz, line 1: Technically it should be device-orientation.kdl or something per the spec name https://w3c.github.io/deviceorientation/, but I think we don't have any other patch for this spec, and the spec name is a bit weird, so all good.

saschanaz (commented):
LGTM

jakebailey (approved):

saschanaz:
LGTM

PR #2237, opened 2025-11-05
override scroll methods to return void until browser implementation changes

### LINKED ISSUE #2197: [Web API type definition issue] Async `scroll...()` methods are non-async in fact
### Summary

Scroll methods are non-async currently in latest browsers

### Expected vs. Actual Behavior

Expected:
```typescript
scroll(options?: ScrollToOptions): void;
scroll(x: number, y: number): void;
scrollBy(options?: ScrollToOptions): void;
scrollBy(x: number, y: number): void;
scrollIntoView(arg?: boolean | ScrollIntoViewOptions): void;
scrollTo(options?: ScrollToOptions): void;
scrollTo(x: number, y: number): void;

// or with optional async support

scroll(options?: ScrollToOptions): Promise<void> | void;
scroll(x: number, y: number): Promise<void> | void;
scrollBy(options?: ScrollToOptions): Promise<void> | void;
scrollBy(x: number, y: number): Promise<void> | void;
scrollIntoView(arg?: boolean | ScrollIntoViewOptions): Promise<void> | void;
scrollTo(options?: ScrollToOptions): Promise<void> | void;
scrollTo(x: number, y: number): Promise<void> | void;
```

Web Types `0.0.280`
```typescript
scroll(options?: ScrollToOptions): Promise<void>;
scroll(x: number, y: number): Promise<void>;
scrollBy(options?: ScrollToOptions): Promise<void>;
scrollBy(x: number, y: number): Promise<void>;
scrollIntoView(arg?: boolean | ScrollIntoViewOptions): Promise<void>;
scrollTo(options?: ScrollToOptions): Promise<void>;
scrollTo(x: number, y: number): Promise<void>;
```

### Playground Link

_No response_

### Browser Support

- [x] This API is supported in at least two major browser engines (not two Chromium-based browsers).

### Have Tried The Latest Releases

- [x] This issue applies to the latest release of TypeScript.
- [x] This issue applies to the latest release of `@types/web`.

### Additional Context

[Commit](https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/41c31e6db889628ab3c5c6942f1ed07fc3f8df32)

#### MDN

https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollBy
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo

Bashamega:
fixes #2197 

Related #2053

### DIFF
--- a/inputfiles/patches/cssom-view.kdl
+++ b/inputfiles/patches/cssom-view.kdl
@@ -15,7 +15,7 @@ interface-mixin DocumentOrShadowRoot {
   }
 }
 
-// No implementation of Promise return values as of 2025-11"
+// No implementation of Promise return values as of 2025-11
 interface Element {
   method scroll returns=void signatureIndex=0
   method scroll returns=void signatureIndex=1


inputfiles/patches/scroll.kdl
```diff
@@ -0,0 +1,9 @@
+interface Element {
+    method scroll overrideType=void
```
saschanaz: type=undefined should do it as an IDL type

saschanaz: As you see in #2196 it's more than just Element but also Window.

saschanaz:
Title should be "override scroll methods to return void until browser implementation changes" or something.

src/build/patches.ts
```diff
@@ -168,7 +169,39 @@ function handleMixinandInterfaces(
       }
       case "method": {
         const methodName = string(child.values[0]);
-        method[methodName] = handleMethod(child);
+        const m = handleMethod(child);
+        if (method[methodName]) {
+          // ts: The goal here is to merge multiple method signatures together for methods with the same name.
```
saschanaz: ~~we already have a better solution in another PR, no? This is weird.~~

I take this back and added another comment below.

src/build/patches.ts
```diff
@@ -265,15 +298,24 @@ function handleMethod(child: Node): Partial<Method> {
...
-        ? handleTyped(typeNode)
-        : { type: string(child.properties?.returns) }),
-    },
-  ];
-  return { name, signature };
+  const signature:
+    | DeepPartial<Signature>[]
+    | Record<string, DeepPartial<Signature>> = child.properties?.overrideType
+    ? {
+        "0": {
```
saschanaz: This certainly is the key but I don't think we should do it this way, What if we only want to override the second signature but leave the first one as-is?

What about something like:

```kdl
method name signatureIndex=0 type=undefined
```

And then we return object as signature list instead of an array if `signatureIndex` exists?

Bashamega:
I have updated it @saschanaz 

I couldn't update window, because it gave me this error
file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:766
            method.signature.forEach((sig) => emitSignature(sig, prefix, name, printLine, shouldResolvePromise));
                             ^

TypeError: method.signature.forEach is not a function
    at emitSignatures (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:766:30)
    at emitMethod (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:751:9)
    at file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:790:33
    at Array.forEach (<anonymous>)
    at emitMethods (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:790:18)
    at emitMembers (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:830:9)
    at emitInterface (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:1063:9)
    at emitNonCallbackInterfaces (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:1092:17)
    at emit (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:1184:9)
    at emitWebIdl (file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:156:21)

src/build/patches.ts
```diff
@@ -17,6 +18,10 @@ type DeepPartial<T> = T extends object
   ? { [K in keyof T]?: DeepPartial<T[K]> }
   : T;
 
+interface Method extends Omit<originalMethod, "signature"> {
+  signature: DeepPartial<Signature>[] | Record<string, DeepPartial<Signature>>;
```
saschanaz: It's more like `Record<number, ...>`, and maybe name the interface something like OverridableMethod, so that we don't have to rename the pure Method

src/build/patches.ts
```diff
@@ -265,14 +273,24 @@ function handleMethod(child: Node): Partial<Method> {
...
-    },
-  ];
+  // Determine the actual signature object
+  const signatureObj: DeepPartial<Signature> = {
+    param: params,
+    ...(typeNode
+      ? handleTyped(typeNode)
+      : {
+          type: string(child.properties?.returns),
+          subtype: undefined,
```
saschanaz, line 283: why subtype?

src/build/patches.ts
```diff
@@ -265,14 +273,24 @@ function handleMethod(child: Node): Partial<Method> {
...
+        }),
+  };
+
+  let signature: Method["signature"];
+  const signatureIndex = child.properties?.signatureIndex;
```
saschanaz, line 288: time to add `number()`?

saschanaz:
> I couldn't update window, because it gave me this error file:///home/adam/os/TypeScript-DOM-lib-generator/lib/build/emitter.js:766 method.signature.forEach((sig) => emitSignature(sig, prefix, name, printLine, shouldResolvePromise)); ^

Meaning you may have touched a non-existent method. signatureIndex would work only for an existing method. Note that not all Element methods exist in Window and try again based on that?

src/build/patches.ts
Bashamega: Why? Sometimes it is undefined

src/build/patches.ts

src/build/patches.ts
Bashamega: With out it it will become `undefind<undefind>`

src/build/patches.ts
saschanaz: Hmm, fair.

src/build/patches.ts
saschanaz: Hmm, fair. if one needs subtype they should use nested types.

saschanaz:
Much more concise than the initial PR. Thanks, LGTM

src/build/patches.ts
Bashamega: Okay

Bashamega:
I am still working on the rest

saschanaz:
Oh right, the window methods.

Bashamega:
Done @saschanaz

inputfiles/patches/scroll.kdl
```diff
@@ -0,0 +1,18 @@
...
+}
+
+interface Window {
+    method scroll returns=void signatureIndex=0
+    method scroll returns=void signatureIndex=1
+    method scrollBy returns=void signatureIndex=0
+    method scrollBy returns=void signatureIndex=1
+    method scrollTo returns=void signatureIndex=0
+    method scrollTo returns=void signatureIndex=1
+}
```
saschanaz: Lastly let's put this in cssom-view.kdl

saschanaz: With some comments like "No implementation of Promise return values as of 2025-11"

Bashamega:
I have updated it @saschanaz

inputfiles/patches/cssom-view.kdl
```diff
@@ -14,3 +14,23 @@ interface-mixin DocumentOrShadowRoot {
     param y type=long
   }
 }
+
+// No implementation of Promise return values as of 2025-11"
```
saschanaz:
```suggestion
// No implementation of Promise return values as of 2025-11
```

saschanaz:
LGTM2

Bashamega:
Can you review this @jakebailey 
Thank you 😊

turansky:
Can we use `Promise<void> | void` as return type?
It's expected baseline for nearest 2 years.

saschanaz:
We will migrate to `Promise<void>` if at least two browser engines are ready.

saschanaz:
LGTM

PR #2238, opened 2025-11-05
Migrate CSSFontFaceRule and CSSStyleProperties

Bashamega:
#2053

### DIFF
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -232,7 +232,6 @@ function handleProperty(child: Node): Partial<Property> {
     ...optionalMember("type", "string", child.properties?.type),
     ...optionalMember("readonly", "boolean", child.properties?.readonly),
     ...optionalMember("deprecated", "boolean", child.properties?.deprecated),
-    ...optionalMember("mdnUrl", "string", child.properties?.mdnUrl),
   };
 }
 


inputfiles/patches/cssom.kdl
```diff
@@ -6,4 +6,13 @@ interface CSSStyleDeclaration \
...
-interface CSSStyleProperties replaceReference=CSSStyleDeclaration
+interface CSSStyleProperties replaceReference=CSSStyleDeclaration {
+  property imageOrientation deprecated=#true
+  property wordWrap deprecated=#true
+  
+  // The corresponding standardized property is not supported by
+  // anyone as of 2024-10.
+  property webkitLineClamp deprecated=#false
+
+  property cssFloat mdnUrl="https://developer.mozilla.org/docs/Web/API/CSSStyleDeclaration/cssFloat"
```
saschanaz: Do we lose MDN URL if we skip this one? We have https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleProperties/cssFloat so hopefully not?

inputfiles/patches/cssom.kdl
```diff
@@ -6,4 +6,13 @@ interface CSSStyleDeclaration \
   forward=CSSStyleDeclarationBase \
   forwardExtends=CSSStyleProperties
 
-interface CSSStyleProperties replaceReference=CSSStyleDeclaration
+interface CSSStyleProperties replaceReference=CSSStyleDeclaration {
+  property imageOrientation deprecated=#true
```
saschanaz: We should probably skip this one. I did this for some reason in 0506730e1d1ca9260620189c4f6c8ca19716f0ba without a comment (this is why every patch should ideally have a comment 🥹, but presumably it was because of https://github.com/w3c/csswg-drafts/commit/600fd752153edbe15326a04762334ede1872ad84). But in 2025 all browsers have it and I don't think anyone would unship this.

Bashamega:
Done @saschanaz

Bashamega:
Anything left here @saschanaz

src/build/patches.ts
```diff
@@ -231,6 +231,8 @@ function handleProperty(child: Node): Partial<Property> {
     ...optionalMember("overrideType", "string", child.properties?.overrideType),
     ...optionalMember("type", "string", child.properties?.type),
     ...optionalMember("readonly", "boolean", child.properties?.readonly),
+    ...optionalMember("deprecated", "boolean", child.properties?.deprecated),
+    ...optionalMember("mdnUrl", "string", child.properties?.mdnUrl),
```
saschanaz: Now we don't use this line 👀

inputfiles/patches/cssom.kdl
```diff
@@ -6,4 +6,11 @@ interface CSSStyleDeclaration \
...
   forwardExtends=CSSStyleProperties
 
-interface CSSStyleProperties replaceReference=CSSStyleDeclaration
+interface CSSStyleProperties replaceReference=CSSStyleDeclaration {
+  property wordWrap deprecated=#true
+  
+  // The corresponding standardized property is not supported by
+  // anyone as of 2024-10.
+  property webkitLineClamp deprecated=#false
+
```
saschanaz:
```suggestion
```

src/build/patches.ts
saschanaz:
```suggestion
```

saschanaz:
LGTM

PR #2240, opened 2025-11-05
Refactor emitFlavor function to streamline file generation

Bashamega:
This way it is more maintainable and readable

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -51,7 +51,7 @@ async function emitFlavor(
   mergeNamesakes(exposed);
   exposed.events = webidl.events;
 
-  // Helper to emit and write multiple iterators
+  // Iterator types in separate files as the default target doesn't understand iterators (for TS 6.0-)
   const outputs = [
     {
       suffix: ".generated.d.ts",


src/build.ts
```diff
@@ -51,40 +51,35 @@ async function emitFlavor(
...
-      `${options.name}.asynciterable.generated.d.ts`,
-      options.outputFolder,
-    ),
-    asyncIterators,
+  await Promise.all(
+    outputs.map(async ({ suffix, signature }) => {
+      const content = emitWebIdl(
+        exposed,
+        options.global[0],
+        signature,
```
saschanaz: The parameter name is `iterator`, maybe the argument name should also be the same 👀

Bashamega:
Done @saschanaz

src/build.ts
```diff
@@ -51,40 +51,35 @@ async function emitFlavor(
...
-    exposed,
-    options.global[0],
-    "sync",
-    options.compilerBehavior,
-  );
-  await fs.writeFile(
-    new URL(`${options.name}.iterable.generated.d.ts`, options.outputFolder),
-    iterators,
-  );
+  // Helper to emit and write multiple iterators
```
saschanaz:
```suggestion
  // Iterator types in separate files as the default target doesn't understand iterators (for TS 6.0-)
```

Probably should be formatted though.

saschanaz:
LGTM

jakebailey (approved):

PR #2245, opened 2025-11-06
Move the generate comments function down after all the merges

Bashamega:
This way all the interfaces have comments, even overwritten ones.

### DIFF
--- a/baselines/ts5.5/audioworklet.generated.d.ts
+++ b/baselines/ts5.5/audioworklet.generated.d.ts
@@ -347,7 +347,9 @@ declare var ByteLengthQueuingStrategy: {
  * [MDN Reference](https://developer.mozilla.org/docs/Web/API/CompressionStream)
  */
 interface CompressionStream extends GenericTransformStream {
+    /** The **`readable`** read-only property of the CompressionStream interface returns a ReadableStream. */
     readonly readable: ReadableStream<Uint8Array>;
+    /** The **`writable`** read-only property of the CompressionStream interface returns a WritableStream. */
     readonly writable: WritableStream<BufferSource>;
 }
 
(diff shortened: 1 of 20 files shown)


saschanaz (commented):
This way we can't override comments, but we probably won't ever do that. LGTM thus.

baselines/ts5.5/audioworklet.generated.d.ts
```diff
@@ -347,7 +347,9 @@ declare var ByteLengthQueuingStrategy: {
  * [MDN Reference](https://developer.mozilla.org/docs/Web/API/CompressionStream)
  */
 interface CompressionStream extends GenericTransformStream {
+    /** The **`readable`** read-only property of the CompressionStream interface returns a ReadableStream. */
```
jakebailey, line 350: Wow, what useless comments... Who is providing these?

Bashamega: MDN, they are from MDN

jakebailey: It's a shame we are wasting parsing time and memory on these comments that just restate the annotation 😞

Bashamega: But some comments are useful

jakebailey: Some are absolutely, I am just lamenting comments that are just:

```ts
interface Foo {
    // The **`bar`** read-only property of the Foo interface returns a string
    readonly bar: string;
}
```

saschanaz:
LGTM. Yeah some comments are basically not anything more than placeholders, but we don't have a great way to filter them without adding some magic, which I don't want to. 😔

PR #2247, opened 2025-11-06
Migrate most html interfaces

Bashamega:
#2053

### DIFF
--- a/inputfiles/patches/html.kdl
+++ b/inputfiles/patches/html.kdl
@@ -17,10 +17,10 @@ interface HTMLInputElement {
 
 interface HTMLIFrameElement {
     property referrerPolicy type=ReferrerPolicy
-    property loading overrideType="\"eager\" | \"lazy\""
+    property loading overrideType=#""eager" | "lazy""#
 }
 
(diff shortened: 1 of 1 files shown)


Bashamega:
Can you check this @saschanaz

inputfiles/patches/html.kdl
```diff
@@ -10,3 +10,17 @@ interface HTMLTableHeaderCellElement extends=HTMLTableCellElement exposed=Window
...
+}
+
+interface HTMLIFrameElement {
+    property referrerPolicy type=ReferrerPolicy
+    property loading overrideType="\"eager\" | \"lazy\""
+}
+
+interface HTMLImageElement {
+    property decoding overrideType="\"async\" | \"sync\" | \"auto\""
+    property loading overrideType="\"eager\" | \"lazy\""
```
saschanaz: You probably want to try the alternative string literal: `#" (inside this you don't have to escape quotes) "#`

Bashamega: Done @saschanaz

Bashamega:
What do you think @saschanaz

PR #2250, opened 2025-11-07
KDL support for removals

Bashamega:
#2053 

Added support for removing items using KDL


Bashamega:
What do you think @saschanaz

saschanaz:
A bit busy... hopefully this weekend.

Bashamega:
No problem

src/build/patches.ts
```diff
@@ -375,14 +375,37 @@ export async function readPatch(fileUrl: URL): Promise<any> {
   return parseKDL(text);
 }
 
+function removeNamesDeep(obj: unknown): unknown {
```
saschanaz: So this basically removes the `name` field of the generated objects, right?

I think I'd ultimately prefer not generating `name` for removal purpose, but that would require rather large refactoring (as basically all functions would get an additional parameter, or refactored to make a class, etc.). I'm planning to do a refactoring when we finish the migration anyway, so maybe we can do this for now and pick it up later.

src/build/patches.ts
```diff
@@ -375,14 +375,37 @@ export async function readPatch(fileUrl: URL): Promise<any> {
...
+  if (Array.isArray(obj)) {
+    return obj.map(removeNamesDeep);
+  } else if (obj && typeof obj === "object") {
```
```suggestion
  }
  if (obj && typeof obj === "object") {
```

src/build/patches.ts
```diff
@@ -375,14 +375,37 @@ export async function readPatch(fileUrl: URL): Promise<any> {
...
+  }
+  return obj;
+}
+
 /**
  * Read, parse, and merge all KDL files under the input folder.
  */
-export default async function readPatches(): Promise<any> {
-  const patchDirectory = new URL("../../inputfiles/patches/", import.meta.url);
+export default async function readPatches(
```
saschanaz: It would be sad to put removal files separately. Would be nice if we can:

```kdl
// ... somenormal patches ...

// and then removals in the same file
removals {
  enum foo { bar }
}
```

We can extract all `removals` node from the result of `readPatch` and then separately return all extracted removals.

src/build/patches.ts
saschanaz:
```suggestion
/**
 * Remove all name fields from the object and its children as we don't want
 * the names to be part of the removal.
 */
function removeNamesDeep(obj: unknown): unknown {
```

The parameter and the return types are sad, but I'll not complain for now...

src/build/patches.ts
```diff
@@ -91,8 +114,13 @@ function parseKDL(kdlText: string): DeepPartial<WebIdl> {
   const mixin: Record<string, DeepPartial<Interface>> = {};
   const interfaces: Record<string, DeepPartial<Interface>> = {};
   const dictionary: Record<string, DeepPartial<Dictionary>> = {};
+  let removals: DeepPartial<WebIdl> = {};
 
   for (const node of nodes) {
+    if (node.name === "removals") {
+      removals = parseKDL(node.children);
+      continue;
```
saschanaz: I think the logic change should rather happen in readPatches.

1. `parse(kdl)` can happen in `readPatches`
2. `readPatches` then extracts `removals` from the resulting `Document`.
3. Now `readPatches` has two Document for each file - the normal and the removals.
4. Pass that to `parseKDL()` (which maybe a new name, because it doesn't parse anymore).
5. Return the results as `{ patches, removalPatches }`. (Then the call can happen once rather than twice.)

src/build/patches.ts
```diff
@@ -77,22 +93,21 @@ function handleTypeParameters(value: Value) {
 }
 
 /**
- * Converts patch files in KDL to match the [types](types.d.ts).
+ * Converts parsed KDL Document nodes to match the [types](types.d.ts).
  */
-function parseKDL(kdlText: string): DeepPartial<WebIdl> {
-  const { output, errors } = parse(kdlText);
+function convertKDLNodes(nodes: Node[] | Document): DeepPartial<WebIdl> {
```
saschanaz: Hmm? `Document` is equivalent to `Node[]`, this shouldn't be needed.

src/build/patches.ts
```diff
@@ -113,10 +128,10 @@ function parseKDL(kdlText: string): DeepPartial<WebIdl> {
   }
 
   return {
-    enums: { enum: enums },
-    mixins: { mixin },
-    interfaces: { interface: interfaces },
-    dictionaries: { dictionary },
+    ...optionalMember("enums.enum", "object", enums),
```
saschanaz: Do we need this? Why?

src/build/patches.ts
```diff
@@ -368,21 +383,87 @@ async function getAllFileURLs(folder: URL): Promise<URL[]> {
...
+ */
+function removeNamesDeep(obj: unknown): unknown {
+  if (Array.isArray(obj)) {
+    return obj.map(removeNamesDeep);
+  }
+  if (obj && typeof obj === "object") {
+    const newObj: { [key: string]: unknown } = {};
+    for (const [key, value] of Object.entries(obj as Record<string, unknown>)) {
+      if (key !== "name") {
+        newObj[key] = removeNamesDeep(value);
```
saschanaz: Do you have a plan how to make `null`s work to remove other things than enum values?

src/build/patches.ts

Bashamega:
I have fixed it, and added support for dictionaries @saschanaz

src/build/patches.ts
saschanaz: I still don't understand why this change is needed

src/build/patches.ts
```diff
@@ -77,22 +93,21 @@ function handleTypeParameters(value: Value) {
...
- * Converts patch files in KDL to match the [types](types.d.ts).
+ * Converts parsed KDL Document nodes to match the [types](types.d.ts).
  */
-function parseKDL(kdlText: string): DeepPartial<WebIdl> {
-  const { output, errors } = parse(kdlText);
+function convertKDLNodes(nodes: Node[]): DeepPartial<WebIdl> {
+  // Accept either Document or array of nodes
+  const actualNodes: Node[] = Array.isArray(nodes)
+    ? nodes
+    : nodes;
```
saschanaz: nodes or nodes is just nodes

src/build/patches.ts
```diff
@@ -368,21 +385,92 @@ async function getAllFileURLs(folder: URL): Promise<URL[]> {
...
+ * replace any empty objects ({} or []) with null.
+ */
+function sanitizeRemovals(obj: unknown): unknown {
+  if (Array.isArray(obj)) {
+    const result = obj.map(sanitizeRemovals).filter((v) => v !== undefined);
+    return result.length === 0 ? null : result;
+  }
+  if (obj && typeof obj === "object") {
+    const newObj: { [key: string]: unknown } = {};
+    for (const [key, value] of Object.entries(obj as Record<string, unknown>)) {
```
```suggestion
    for (const [key, value] of Object.entries(obj)) {
```

This should just work without it: https://www.typescriptlang.org/play/?ssl=4&ssc=1&pln=5&pc=1#code/DYUwLgBA9gXBCuA7A1oqB3REC8EDeAvgNwBQJAlgGYQAUUEAZAxGAJ4AOIU192fEAIigAjAFYgAxmAEBKfCQgQA8mMlgAdCERgATuRABnOnID0JlgAtyBiKPgHI6KDuQGSBIA

src/build/patches.ts
```diff
@@ -368,21 +385,92 @@ async function getAllFileURLs(folder: URL): Promise<URL[]> {
...
+ * Recursively remove all 'name' fields from the object and its children, and
+ * replace any empty objects ({} or []) with null.
+ */
+function sanitizeRemovals(obj: unknown): unknown {
+  if (Array.isArray(obj)) {
+    const result = obj.map(sanitizeRemovals).filter((v) => v !== undefined);
+    return result.length === 0 ? null : result;
+  }
+  if (obj && typeof obj === "object") {
+    const newObj: { [key: string]: unknown } = {};
```
saschanaz: Can use Record here

Bashamega:
Done @saschanaz

src/build/patches.ts
```diff
@@ -113,10 +123,14 @@ function parseKDL(kdlText: string): DeepPartial<WebIdl> {
...
-    mixins: { mixin },
-    interfaces: { interface: interfaces },
-    dictionaries: { dictionary },
+    enums: {
+      enum: enums,
+    },
+    dictionaries: {
+      dictionary,
+    },
+    ...optionalMember("mixins.mixin", "object", mixin),
```
saschanaz: Same for this and below, and generally the dot accessor thing as a whole.

Bashamega:
Is it better now @saschanaz

src/build/patches.ts
```diff
@@ -22,7 +22,6 @@ type DeepPartial<T> = T extends object
 interface OverridableMethod extends Omit<Method, "signature"> {
   signature: DeepPartial<Signature>[] | Record<number, DeepPartial<Signature>>;
 }
-
```
saschanaz, line 25: (this should be restored)

src/build/patches.ts
```diff
@@ -368,21 +361,96 @@ async function getAllFileURLs(folder: URL): Promise<URL[]> {
...
+  const removalObjs = removalsNodeGroups.map((nodes) => convertKDLNodes(nodes));
+
+  const patches = patchObjs.reduce((acc, cur) => merge(acc, cur), {});
+  const removalPatches = sanitizeRemovals(
+    removalObjs.reduce((acc, cur) => merge(acc, cur), {}),
+  ) as DeepPartial<WebIdl>;
+
+  // Only because we don't use them
+  removalPatches.mixins = undefined;
+  removalPatches.interfaces = undefined;
```
saschanaz: Does this matter - if we omit this does the test fail?

src/build/patches.ts
```diff
@@ -368,21 +361,96 @@ async function getAllFileURLs(folder: URL): Promise<URL[]> {
 }
 
 /**
- * Read and parse a single KDL file.
+ * Read and parse a single KDL file into its KDL Document structure.
  */
-export async function readPatch(fileUrl: URL): Promise<any> {
+async function readPatchDocument(fileUrl: URL): Promise<Node[]> {
```
saschanaz: It says Document, so maybe Document fits better for return type?

```suggestion
async function readPatchDocument(fileUrl: URL): Promise<Document> {
```

src/build/patches.ts
```diff
@@ -368,21 +361,96 @@ async function getAllFileURLs(folder: URL): Promise<URL[]> {
...
+      cause: errors,
+    });
+  }
+  return output!;
+}
+/**
+ * Recursively remove all 'name' fields from the object and its children, and
+ * replace any empty objects ({} or []) with null.
+ */
+function sanitizeRemovals(obj: unknown): unknown {
```
saschanaz: convertForRemovals? It's not really sanitizing - normalizing random user input, but rather converting into a different format.

src/build/patches.ts
```diff
@@ -368,21 +361,96 @@ async function getAllFileURLs(folder: URL): Promise<URL[]> {
...
+  }
+
+  // Stage 3: Merge all main patches and removals separately using convertKDLNodes
+  const patchObjs = patchNodeGroups.map((nodes) => convertKDLNodes(nodes));
+  const removalObjs = removalsNodeGroups.map((nodes) => convertKDLNodes(nodes));
+
+  const patches = patchObjs.reduce((acc, cur) => merge(acc, cur), {});
+  const removalPatches = sanitizeRemovals(
+    removalObjs.reduce((acc, cur) => merge(acc, cur), {}),
+  ) as DeepPartial<WebIdl>;
```
```suggestion
  // Stage 2: Group by patches or removals
  const merged = documents.flat();
  const patchNodes = merged.filter((node) => node.name !== "removals");
  const removalNodes = merged.filter((node) => node.name === "removals").map((node) => node.children)).flat();

  // Stage 3: Convert the nodes for patches and removals respectively
  const patches = convertKDLNodes(patchNodes);
  const removalPatches = sanitizeRemovals(convertKDLNodes(removalNodes));
```

src/build/patches.ts
Bashamega: Yes without it the tests fail

Bashamega: Yes without it the tests fail

Bashamega: Yes without it the tests fail

saschanaz: Ah because `interface: {}` becomes `interface: null`. Hmmmmm I think this may be a footgun later when some patches become redundant and are removed...

Bashamega:
Thanks @saschanaz 
Is there anything else left?

saschanaz:
LGTM, thanks!

PR #2263, opened 2025-12-02
Migrate to use OOP

Bashamega:
I am very sorry @saschanaz 🤗
I rewrote the file to use OOP, as you suggested, plus it is more readable this way


saschanaz:
I didn't intend to suggest this, I was more like I don't want this kind of big change at this point 😅🙈

Bashamega:
No problem @saschanaz

PR #2269, opened 2025-12-04
Migrate Media Track

Bashamega:


inputfiles/patches/media-track.kdl
```diff
@@ -0,0 +1,124 @@
+removals {
+  dictionary MediaTrackCapabilities {
+    member latency // Blink only as of 2022-11
+    member resizeMode // Blink only as of 2022-11
+
+    // Screen Share extensions
+    // https://w3c.github.io/mediacapture-screen-share/#mediadevices-additions
```
saschanaz: This is huge. Would be nice to split into mediacapture-screen and mediacapture-image based on this

Bashamega:
I have updated it @saschanaz

Bashamega:
When will you merge this @saschanaz

Bashamega:
@saschanaz

saschanaz:
Consider requesting review from the GitHub UI rather than manually pinging repeatedly. That way I can track the requested reviews better.

LGTM

saschanaz:
Actually I can't even make a request for myself, the feature is not available at all here. Is that something blocked in the repository configuration? @jakebailey

saschanaz:
Or maybe I need write access to use that actually? That's annoying.

jakebailey:
I invited you as a triager, which I think should work.

PR #2271, opened 2025-12-05
Make sure deprecated is always a string

Bashamega:


inputfiles/patches/cssom.kdl
```diff
@@ -7,9 +7,9 @@ interface CSSStyleDeclaration \
   forwardExtends=CSSStyleProperties
 
 interface CSSStyleProperties replaceReference=CSSStyleDeclaration {
-  property wordWrap deprecated=#true
+  property wordWrap deprecated="The 'wordWrap' property is non-standard and has been replaced by 'overflow-wrap' in modern CSS."
   
   // The corresponding standardized property is not supported by
   // anyone as of 2024-10.
-  property webkitLineClamp deprecated=#false
+  property webkitLineClamp supported=#true
```
saschanaz: A better way would be to remove the field through `removals`, although it's not supported yet. Maybe that should be the first step.

Bashamega: Can you share the syntax for it @saschanaz I can't imagine it

Bashamega:
Yay @saschanaz 
I can re-request your review
Can you please tell me the syntax for it?

inputfiles/patches/cssom.kdl
saschanaz: Something like 

```kdl
removals {
  interface CSSStyleProperties {
    property webkitLineClamp deprecated=#null
  }
}
```

Or perhaps it's easier to keep using each property's own type, like `deprecated=""` and let it converted to null via the removal conversion function. Either way would be fine for me.

Bashamega:
I have updated it @saschanaz 
But now this is being logged
removedTypes.json has a redundant field events in {"events":{"event":null},"properties":{"property":{"webkitLineClamp":{"deprecated":null}}},"methods"

src/build/patches.ts
```diff
@@ -256,7 +256,8 @@ function handleProperty(child: Node): Partial<Property> {
       ? handleTyped(typeNode)
       : optionalMember("type", "string", child.properties?.type)),
     ...optionalMember("readonly", "boolean", child.properties?.readonly),
-    ...optionalMember("deprecated", "boolean", child.properties?.deprecated),
+    ...optionalMember("deprecated", "string", child.properties?.deprecated),
+    ...(child.properties?.supported && { deprecated: false }),
```
```suggestion
```

Bashamega:
You can push directly Btw @saschanaz

saschanaz:
LGTM

PR #2272, opened 2025-12-06
Html element onerror eventhandler override

cosmic-pixel-painter:
This PR fixes https://github.com/Microsoft/TypeScript/issues/62075.

Originally, I created this https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2148#issuecomment-3289858315, but closed it as the that approach wasn't ideal.

I started with `Omit<GlobalEventHandlers, 'onerror'>` as an approach. However, I quickly realized it isn't the correct path and creates a whole slew of new problems due to how event handlers are mapped to different elements during the emit phase. (if you're curious to see what exactly I ran into, I can push up a separate PR).

What this change does:

Introduces a new type `DocumentOrGlobalOnErrorEventHandler` which extends the `OnErrorEventHandlerNonNull` type with HTMLElement specific handlers. This type then overrides GlobalEventHandlers onerror property. This allow us to be backwards compatible and fix the bug.

This did require a change to both `types.ts` and `emitter.ts` so that exposed was accounted for when creating a new typedef. I needed the new `DocumentOrGlobalOnErrorEventHandler` to only be exposed on the dom library.

The `onerror` of `GlobalEventHandlers` cannot be overridden by the `overrideType.ts` file. I tried to a few times, in both mixin-ins and interfaces. It turns out it has to be overidden in the `patches/events.kdl `file.

Let me know if I missed anything or another approach would be better. I think this threads the needle nicely as it works only on dom elements and still provides both the global and HTMLElement specific onerror handling types.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -29370,6 +29370,12 @@ interface SVGElement extends Element, ElementCSSInlineStyle, GlobalEventHandlers
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/SVGElement/viewportElement)
      */
     readonly viewportElement: SVGElement | null;
+    /**
+     * The addEventListener() method of the EventTarget interface sets up a function that will be called whenever the specified event is delivered to the target.
+     *
+     * [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)
+     */
+    addEventListener(type: string, listener: ((event: Event) => void) | ((event: UIEvent) => void)): void;
     addEventListener<K extends keyof SVGElementEventMap>(type: K, listener: (this: SVGElement, ev: SVGElementEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
     addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
     removeEventListener<K extends keyof SVGElementEventMap>(type: K, listener: (this: SVGElement, ev: SVGElementEventMap[K]) => any, options?: boolean | EventListenerOptions): void;
@@ -31734,6 +31740,12 @@ interface SVGSVGElement extends SVGGraphicsElement, SVGFitToViewBox, WindowEvent
     unsuspendRedraw(suspendHandleID: number): void;
     /** @deprecated */
     unsuspendRedrawAll(): void;
+    /**
+     * The addEventListener() method of the EventTarget interface sets up a function that will be called whenever the specified event is delivered to the target.
+     *
+     * [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)
+     */
+    addEventListener(type: string, listener: ((event: Event) => void) | ((event: UIEvent) => void)): void;
     addEventListener<K extends keyof SVGSVGElementEventMap>(type: K, listener: (this: SVGSVGElement, ev: SVGSVGElementEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
     addEventListener(type: string, listener: EventListenerOrEventListenerObject, options?: boolean | AddEventListenerOptions): void;
     removeEventListener<K extends keyof SVGSVGElementEventMap>(type: K, listener: (this: SVGSVGElement, ev: SVGSVGElementEventMap[K]) => any, options?: boolean | EventListenerOptions): void;
--- a/baselines/ts5.5/dom.generated.d.ts
+++ b/baselines/ts5.5/dom.generated.d.ts
@@ -29345,6 +29345,12 @@ interface SVGElement extends Element, ElementCSSInlineStyle, GlobalEventHandlers
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/SVGElement/viewportElement)
      */
     readonly viewportElement: SVGElement | null;
+    /**
+     * The addEventListener() method of the EventTarget interface sets up a function that will be called whenever the specified event is delivered to the target.
+     *
+     * [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)
+     */
+    addEventListener(type: string, listener: ((event: Event) => void) | ((event: UIEvent) => void)): void;
     addEventListener<K extends keyof SVGElementEventMap>(type: K, listener: (this: SVGElement, ev: SVGElementEventMap[K]) => any, options?: boolean | AddEventListenerOptions): void;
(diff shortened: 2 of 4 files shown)


cosmic-pixel-painter:
This one is ready for review @saschanaz.

baselines/dom.generated.d.ts
```diff
@@ -29370,6 +29370,12 @@ interface SVGElement extends Element, ElementCSSInlineStyle, GlobalEventHandlers
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/SVGElement/viewportElement)
      */
     readonly viewportElement: SVGElement | null;
+    /**
+     * The addEventListener() method of the EventTarget interface sets up a function that will be called whenever the specified event is delivered to the target.
+     *
+     * [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)
+     */
+    addEventListener(type: string, listener: ((event: Event) => void) | ((event: UIEvent) => void)): void;
```
saschanaz, line 29378: This is not quite right as this affects all other events. And still not sure about UIEvent:

1. MDN's description is a copypaste of what UI Events spec says in https://w3c.github.io/uievents/#event-type-error:
   > UIEvent if generated from a user interface, Event otherwise.
2. But then the same spec says otherwise in https://w3c.github.io/uievents/#event-types-list, where it only says Error for error event. So does HTML in https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers-on-elements,-document-objects,-and-window-objects.
3. Nothing in UI Events says exactly when UIEvent is used for error event.
4. UI Events also argues load event uses UIEvent, which is untrue.
5. UI Events hasn't been well maintained, see also the "Future of UI Events" discussion in https://docs.google.com/document/d/1fOjXiGUFf_krkfYaWf5drbqzY7CTxn7EcDYr5ZWjzFg/preview.

So I'd like to suggest excluding UIEvent in this PR.

baselines/dom.generated.d.ts
```diff
@@ -15071,7 +15071,7 @@ interface GlobalEventHandlers {
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLMediaElement/ended_event) */
     onended: ((this: GlobalEventHandlers, ev: Event) => any) | null;
     /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/error_event) */
-    onerror: OnErrorEventHandler;
+    onerror: DocumentOrGlobalOnErrorEventHandler;
```
saschanaz, line 15074: This also affects all other interfaces so it partially repeats the same issue of OnErrorEventHandler, and also not quite I suggested in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2148#issuecomment-3289921974...

I was thinking about:

```ts
interface DocumentOrElementEventHandlers extends GlobalEventHandlers {
  onerror: ((this: GlobalEventHandlers, ev: Event) => any) | null;
};
```

And make others inherit this instead of GlobalEventHandlers when appropriate. Wdyt?

saschanaz: But that as-is has an assignability issue. Hmm.

saschanaz: We do have `extendConflictsBaseTypes` in emitter.ts which can help here... it's sad to add more there but it does make sense here.

saschanaz: Doing so would make subinterfaces to fail to assign to a variable typed as GlobalEventHandlers, which is a bit worrisome.

Hmmmmm. Maybe a better way is to make it reversed?


```ts
interface GlobalEventHandlers {
  // remove OnErrorEventHandler and make it general here  
  onerror: ((this: GlobalEventHandlers, ev: Event) => any) | null;
}

interface GlobalErrorEventHandlers extends GlobalEventHandlers {
  onerror: OnErrorEventHandler;
}

// Make global interfaces extend GlobalErrorEventHandlers instead of GlobalEventHandlers.
// Ideally this should be replaced automatically by emitter than overridingTypes, as
// the latter wouldn't support any upcoming new globals.
// Checking `global` property for each interface should allow it. 
interface Window extends GlobalErrorEventHandlers {
  // ...
}
```

saschanaz: Err. Actually if we are going to check globals then we don't need GlobalErrorEventHandlers at all. The emitter can just add an overridden `onerror` property directly to each global interface.


```ts
interface GlobalEventHandlers {
  // Remove OnErrorEventHandler and make it general here.
  // Replacing it to a general EventHandler would do this automatically.
  onerror: ((this: GlobalEventHandlers, ev: Event) => any) | null;
}

interface Window extends GlobalEventHandlers {
  onerror: OnErrorEventHandler;
}
```

cosmic-pixel-painter: Hey! Zero worries about the delay I know how busy it gets. This bug is heady. First off, agreed on `UIEvent`. I'll drop that from the PR.

For the bug itself. I think the following things are true:
- We want this override to be specific to `HTMLElement` and its subtypes.
- We don't want to override window.
- We don't want to break backwards compatibility. 

The context I've found while attempting to fix this bug:
`HTMLElement` extends the `GlobalEventHandlers` interface which implements `OnErrorEventHandler`
(a type alias `OnErrorEventHandlerNonNull`) for for the onerror method. It's a callable interface. 

```
interface OnErrorEventHandlerNonNull {
    (event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error): any;
}
```

If I extend `OnErrorEventHandlerNonNull` I can't override the call signature when it's implemented on `HTMLElement`. It throws a type error of 
```// errors Type 'Event | string' is not assignable to type 'string'. Type 'Event' is not assignable to type 'string'```

```
interface DocumentOnErrorEventHandler extends OnErrorEventHandlerNonNull {
 (event: Event): any;
}
interface HTMLElement {  
 onerror: DocumentOnErrorEventHandler; // errors Type 'Event | string' is not assignable to type 'string'. Type 'Event' is not assignable to type 'string'
}
```

To solve this, I wonder if having `HTMLElement` not extend `GlobalEventHandlers` but instead extend a new interface `DocumentEventHandlers` ? Both `GlobalEventHandlers` and `DocumentEventHandlers` could inherit from the same base event handlers interface (it would need created here) and then each interface, `GlobalEventHandlers` and `DocumentEventHandlers`, could implement specific handlers when necessary?

That seems like a lot for this bug, but I'm not sure how else to meet the `HTMLElement` signature and `GlobalEventHandlers` signature requirements. 

Let me know what you think and if there's a better direction I'm all up for it. 

I'll fiddle a bit and see what it feels like.

PR #2274, opened 2025-12-07
Adds a rule to make sure that optional params can't come before required

Bashamega:


inputfiles/overridingTypes.jsonc
```diff
@@ -288,22 +288,6 @@
...
-            "GPUPipelineError": {
-                // Web IDL optional argument can be followed by a required argument, but not in TS
-                // TODO: Maybe do this in the emitter?
-                "constructor": {
-                    "signature": {
-                        "0": {
-                            "param": [
-                                {
-                                    "name": "message",
-                                    "optional": false
```
saschanaz, line 300: This was ok because it was one-off, but it's not great to automate this nonideal workaround.

Given the IDL parameters of "required, optional, and then required" pattern, both form of the function call should be allowed:

```
func(required);
func(required, undefined, required);
```

This can be only done with overloads in current TS. I'm not sure it's worth right now given it's very rare to have this problem.

src/build/widlprocess.ts
```diff
@@ -340,15 +340,29 @@ function convertCallbackFunctions(
   };
 }
 
-function convertArgument(arg: webidl2.Argument): Browser.Param {
+function hasRequiredArgumentAfter(arr: webidl2.Argument[], i: number): boolean {
```
saschanaz, line 343: Also I think this is a wrong layer to solve the problem. widlprocess should be about conversion to the format that the build script understands, and not about changing anything. That should be part of the emitter, as the comment above says.

Bashamega:
Okay

PR #2279, opened 2025-12-08
Support additionalType

Bashamega:


src/build/patches.ts
```diff
@@ -63,6 +63,20 @@ function handleTyped(type: Node): Typed {
   };
 }
 
+function handleAdditionalTypes(node: Node) {
+  const additionalTypes = [];
+  for (const child of node.children) {
+    if (child.name === "additionalType") {
```
saschanaz: This is assuming to repeat like

```
additionalType foo
additionalType bar
```

right? But maybe just `additionalTypes foo bar`, because it's always just a simple string?

Bashamega:
I have updated it @saschanaz, and added a sample for multiple additionaltypes

inputfiles/patches/webcrypto.kdl
```diff
@@ -13,6 +13,14 @@ interface CryptoKey {
   }
 }
 
+interface SubtleCrypto {
+  method decrypt signatureIndex=0 {
+    param algorithm {
+     additionalTypes RsaOaepParams AesCtrParams AesCbcParams AesGcmParams
```
saschanaz:
```suggestion
      additionalTypes RsaOaepParams AesCtrParams AesCbcParams AesGcmParams
```

src/build/patches.ts
```diff
@@ -63,6 +63,15 @@ function handleTyped(type: Node): Typed {
   };
 }
 
+function handleAdditionalTypes(node: Node) {
+  for (const child of node.children) {
+    if (child.name === "additionalTypes") {
+      const additionalTypes = child.values.map((v) => string(v));
```
```suggestion
      const additionalTypes = child.values.map(string);
```

Bashamega:
I have fixed the broken code @saschanaz

saschanaz:
LGTM, thanks!

PR #2284, opened 2025-12-09
Support removing enums

Bashamega:
We got rid of all enums in the JSON files 🥳

### DIFF
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -144,7 +144,7 @@ function handleEnum(node: Node): Partial<Enum> {
 
   return {
     name,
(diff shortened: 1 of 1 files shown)


src/build/patches.ts
```diff
@@ -144,7 +144,7 @@ function handleEnum(node: Node): Enum {
 
   return {
     name,
-    value: values,
+    ...(values.length > 0 ? { value: values } : {}),
```
saschanaz: optionalNestedMember should do this job

saschanaz:
LGTM, thanks!

saschanaz:
Err but actually, why do we remove that one? And that wasn't HTML...

PR #2288, opened 2025-12-14
Add support for type parameters for interfaces and nullable property

Bashamega:


inputfiles/patches/indexeddb.kdl
```diff
@@ -0,0 +1,5 @@
+interface IDBRequest {
+   property source nullable=#false
```
saschanaz: this should use `type` node instead

Bashamega: I don't understand @saschanaz 
Can you clarify?
This is how it is structured in the json file

Bashamega:
If you meant line 3 @saschanaz, then it doesn't work because it is not a known type

inputfiles/patches/indexeddb.kdl
saschanaz: And not in KDL, because it's "property with a nullable type" rather than "nullable property"...

Bashamega:
The issue is that it is an empty string  what do you want me to do @saschanaz 

> Sorry for all the commits I am working from my mobile phone right now and use the Github app

inputfiles/patches/indexeddb.kdl
```diff
@@ -1,5 +1,5 @@
 interface IDBRequest {
    property source nullable=#false
-   property result overrideType=T
+   property result type=T
```
saschanaz: It should be overrideType as T is not Web IDL type

Maybe overrideType needs a better name...

Bashamega: This is outdated

Bashamega: This is outdated

Bashamega: This is outdated

inputfiles/patches/indexeddb.kdl
```diff
@@ -0,0 +1,7 @@
+interface IDBRequest {
+   property source {
+      type "" nullable=#false
```
saschanaz: We can probably skip the type name in this case (and modify the code to accept that)

inputfiles/patches/indexeddb.kdl
```diff
@@ -0,0 +1,7 @@
...
+   }
+   property result overrideType=T
+   typeParameters T default=any
```
saschanaz: Move this to the top?

Bashamega:
I have fixed it

saschanaz:
LGTM, thanks!

PR #2289, opened 2025-12-15
Migrate definitions for UnderlyingByteSource and UnderlyingDefaultSource

Bashamega:
I love how clean it is


saschanaz:
> I love how clean it is

Exactly.

LGTM

PR #2292, opened 2025-12-16
Migrate most Index-DB overrides

Bashamega:
The only addition in the patches file is to support multiple types and nullable

### DIFF
--- a/inputfiles/addedTypes.jsonc
+++ b/inputfiles/addedTypes.jsonc
@@ -408,10 +408,6 @@
                     }
                 }
             },
-            "IDBOpenDBRequest": {
-                "name": "IDBOpenDBRequest",
-                "extends": "IDBRequest<IDBDatabase>"
-            },
             "HTMLBodyElement": {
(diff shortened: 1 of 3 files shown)


src/build/patches.ts
```diff
@@ -246,22 +246,27 @@ function handleEvent(child: Node): Event {
...
  */
 function handleProperty(child: Node): DeepPartial<Property> {
-  let typeNode: Node | undefined;
-  for (const c of child.children) {
-    if (c.name === "type") {
-      typeNode = c;
-      break;
-    }
+  // Collect all child type nodes
+  const typeNodes = child.children.filter((c) => c.name === "type");
```
saschanaz: This should look more like handleParam and throw when unknown child node exists

Also it would be nice to split this patch to things that need this logic change and things that do not, because it's too big right now.

Bashamega:
I have made it much smaller

Bashamega:
What do you think @saschanaz

PR #2296, opened 2025-12-17
Restores Variation Settings

Bashamega:
It is now supported in blink
Blink: https://issues.chromium.org/issues/40398871


saschanaz:
LGTM, thanks!

PR #2304, opened 2025-12-21
Support multiple types & update toJSON

### LINKED ISSUE #2114: `.toJSON()` on `PublicKeyCredential` effectively returns `any`
Consider the following code:

```ts
async function register() {
  const publicKey = PublicKeyCredential.parseCreationOptionsFromJSON({ /* … */ });
  const credential = (await navigator.credentials.create({publicKey})) as PublicKeyCredential;
  return credential.toJSON();
}
```

The type of the returned code is `any` because:

- `.toJSON()` returns `PublicKeyCredentialJSON` and
- `PublicKeyCredentialJSON` is defined as `any`.

This is fine for some use cases, but it completely removes type safety for accessing any fields on the returned value.

[`@github/webauthn-json` is winding down](https://github.com/github/webauthn-json/pull/95), and developers who transition to the built-in functions will gain some benefits at the expense of losing these types.

I'd be happy to contribute full types for these, since the actual types can be ported and adapted from `@github/webauthn-json` somewhat directly. However, I'm not clear on how to integrate this into `TypeScript-DOM-lib-generator`, as it would be a lot to add to `overridingTypes.jsonc`.

Bashamega:
fixes #2114

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -217,7 +217,6 @@ interface AuthenticationExtensionsClientOutputs {
 
 interface AuthenticationExtensionsClientOutputsJSON {
     appid?: boolean;
-    appidExclude?: boolean;
     credProps?: CredentialPropertiesOutput;
     largeBlob?: AuthenticationExtensionsLargeBlobOutputsJSON;
     prf?: AuthenticationExtensionsPRFOutputsJSON;
--- a/baselines/ts5.5/dom.generated.d.ts
+++ b/baselines/ts5.5/dom.generated.d.ts
@@ -214,7 +214,6 @@ interface AuthenticationExtensionsClientOutputs {
 
 interface AuthenticationExtensionsClientOutputsJSON {
     appid?: boolean;
-    appidExclude?: boolean;
     credProps?: CredentialPropertiesOutput;
     largeBlob?: AuthenticationExtensionsLargeBlobOutputsJSON;
     prf?: AuthenticationExtensionsPRFOutputsJSON;
--- a/baselines/ts5.6/dom.generated.d.ts
+++ b/baselines/ts5.6/dom.generated.d.ts
@@ -214,7 +214,6 @@ interface AuthenticationExtensionsClientOutputs {
 
 interface AuthenticationExtensionsClientOutputsJSON {
     appid?: boolean;
-    appidExclude?: boolean;
     credProps?: CredentialPropertiesOutput;
     largeBlob?: AuthenticationExtensionsLargeBlobOutputsJSON;
     prf?: AuthenticationExtensionsPRFOutputsJSON;
--- a/baselines/ts5.9/dom.generated.d.ts
+++ b/baselines/ts5.9/dom.generated.d.ts
@@ -214,7 +214,6 @@ interface AuthenticationExtensionsClientOutputs {
 
 interface AuthenticationExtensionsClientOutputsJSON {
     appid?: boolean;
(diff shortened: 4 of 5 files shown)


Bashamega:
What do you think @lgarron

src/build/patches.ts
```diff
@@ -62,6 +62,28 @@ function handleTyped(type: Node): DeepPartial<Typed> {
   };
 }
 
+function handleMultipleTypes(
+  typeNodes: Node[],
+  returns?: Value,
+): DeepPartial<Typed> | DeepPartial<Typed>[] | undefined {
```
saschanaz: It's `DeepPartial<Typed> | undefined`, no? It never returns an array.

And this function should be called handleTyped and the singular version should be called handleSingleTypeNode or something.

src/build/patches.ts
```diff
@@ -260,7 +284,7 @@ function handleProperty(child: Node): DeepPartial<Property> {
     ...optionalMember("optional", "boolean", child.properties?.optional),
     ...optionalMember("overrideType", "string", child.properties?.overrideType),
     ...(typeNode
-      ? handleTyped(typeNode)
+      ? handleSingleTypeNode(typeNode)
```
saschanaz: Why? properties can have union types too. handleSingleTypeNode should not be called outside handleTyped.

src/build/patches.ts
```diff
@@ -48,20 +48,44 @@ function string(arg: unknown): string {
...
-function handleTyped(type: Node): DeepPartial<Typed> {
+function handleSingleTypeNode(type: Node): DeepPartial<Typed> {
   const isTyped = type.name == "type";
   if (!isTyped) {
     throw new Error("Expected a type node");
   }
   const subType =
-    type.children.length > 0 ? handleTyped(type.children[0]) : undefined;
+    type.children.length > 0
+      ? handleSingleTypeNode(type.children[0])
```
saschanaz: And subtypes can be union too

Bashamega:
Done

src/build/patches.ts
```diff
@@ -48,20 +48,42 @@ function string(arg: unknown): string {
...
   return {
     ...optionalMember("type", "string", type.values[0]),
     subtype: subType,
     ...optionalMember("nullable", "boolean", type.properties?.nullable),
   };
 }
 
+function handleTyped(
+  typeNodes: Node[],
+  returns?: Value,
```
saschanaz: returns? you mean property?

Bashamega: Yes, I forgot to include it i will fix it

Bashamega:
Better?

Bashamega:
Is there any thing missing here @saschanaz

src/build/patches.ts
saschanaz: You still forgot this

Bashamega: I have renamed it

Bashamega:
Is there any thing missing here @saschanaz

baselines/dom.generated.d.ts
```diff
@@ -215,6 +215,14 @@ interface AuthenticationExtensionsClientOutputs {
     prf?: AuthenticationExtensionsPRFOutputs;
 }
 
+interface AuthenticationExtensionsClientOutputsJSON {
+    appid?: boolean;
+    appidExclude?: boolean;
```
saschanaz: Looks like this has no implementation, otherwise seems good

saschanaz:
LGTM, thanks!

PR #2305, opened 2025-12-21
Support KDL constructor

Bashamega:
related #2053

### DIFF
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -190,10 +190,8 @@ function handleMixinAndInterfaces(
         break;
       }
       case "constructor": {
-        if (constructor) {
-          throw new Error(`Multiple constructors found in ${type} "${name}"`);
-        }
-        constructor = handleMethodAndConstructor(child, true);
+        const c = handleMethodAndConstructor(child, true);
+        constructor = merge(constructor, c);
         break;
       }
       case "typeParameters": {


src/build/patches.ts
```diff
@@ -182,12 +183,19 @@ function handleMixinAndInterfaces(
...
-        const m = handleMethod(child);
+        const m = handleMethodAndConstructor(child);
         method = merge(method, {
           [methodName]: m,
         });
         break;
       }
+      case "constructor": {
+        if (constructor) {
+          throw new Error(`Multiple constructors found in ${type} "${name}"`);
```
saschanaz: It's fine when there are multiple signatures

saschanaz: (although there seems to be no multiple constructor use in patch file...)

Bashamega:
Hello @saschanaz 
I don't understand your comment in the construct PR.
constructor in the interface types is never an array
So my code is correct
Or is there something that I am missing?

Bashamega:
I have updated it

Bashamega:
Can you check this @saschanaz

saschanaz:
Didn't we agree that the request review should be sufficient and you shouldn't ping me every day

Bashamega:
Sorry

saschanaz:
LGTM, thanks!

PR #2306, opened 2025-12-21
Adds AudioEncoderConfig's aac

Bashamega:


saschanaz:
I'm not sure "restore" is the right word for this kind of things, because it's never been in the libs. Maybe Add is the better word.

inputfiles/removedTypes.jsonc
```diff
@@ -126,14 +126,6 @@
                     }
                 }
             },
-            "AudioEncoderConfig": {
-                "members": {
-                    "member": {
-                        "aac": null, // Blink only as of 2024-09
-                        "flac": null // Blink only as of 2024-09
```
saschanaz, line 133: Hmm actually the previous comment was wrong, Blink doesn't have flac config 🤔 and thus is now WebKit only

Bashamega: LOLLLLLLLLLLLLLLLL, I will close this PR 😂

saschanaz: Hmm? aac still has multiple implementations.

saschanaz: You can go ahead in https://searchfox.org/wubkat/search?q=AudioEncoderConfig&path=.idl&case=false&regexp=false and https://source.chromium.org/search?q=AudioEncoderConfig%20file:idl%20path:third_party%2Fblink%2Frenderer&sq= to check it yourself

Bashamega:
I have fixed it @saschanaz

inputfiles/removedTypes.jsonc
Bashamega: Thank you very much

saschanaz:
(methinks people should just use opus but who am I to argue)

LGTM

PR #2307, opened 2025-12-22
Adds MARGIN_RULE

Bashamega:
It is now widely supported
Blink: [source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/css_rule.idl?q=MARGIN_RULE%20file:idl%20path:third_party%2Fblink%2Frenderer](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/css_rule.idl?q=MARGIN_RULE%20file:idl%20path:third_party%2Fblink%2Frenderer) Since  2024-07-01

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -5875,6 +5875,7 @@ interface CSSRule {
     readonly MEDIA_RULE: 4;
     readonly FONT_FACE_RULE: 5;
     readonly PAGE_RULE: 6;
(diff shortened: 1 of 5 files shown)


saschanaz:
"Widely supported" except WebKit never had any actual support for CSSMarginRule... But it does have the constant for whatever reason, so I guess this is fine.

saschanaz:
LGTM

PR #2309, opened 2025-12-22
Migrate WebTransportSendStreamOptions to KDL

Bashamega:
I couldn't find it in Blink, maybe they removed it

### DIFF
--- a/inputfiles/patches/webtransport.kdl
+++ b/inputfiles/patches/webtransport.kdl
@@ -6,6 +6,7 @@ removals {
     // WebKit has it in a different name datagramsReadableMode.
     member datagramsReadableType
   }
+
   dictionary WebTransportSendStreamOptions {
     member waitUntilAvailable // No implementation as of 2025-12.
   }


inputfiles/patches/webtransport.kdl
```diff
@@ -6,4 +6,7 @@ removals {
     // WebKit has it in a different name datagramsReadableMode.
     member datagramsReadableType
   }
+  dictionary WebTransportSendStreamOptions {
+    member waitUntilAvailable // No implementation as of 2025-12.
```
saschanaz, line 11: Yeah, it seems the comment was wrong. But now it's in [WebKit](https://searchfox.org/wubkat/rev/021be8e5aefd8742e091c6369be13fd745207359/Source/WebCore/Modules/webtransport/WebTransport.idl#47) and [Gecko](https://searchfox.org/firefox-main/rev/7d8644b9d4470a675bf670c2dc7664cc01f14ece/dom/webidl/WebTransport.webidl#84).

saschanaz: Oh wait, but it's not about the dictionary itself.

saschanaz:
LGTM

PR #2310, opened 2025-12-23
Add hasUAVisualTransition

Bashamega:
Webkit: https://searchfox.org/wubkat/source/Source/WebCore/dom/PopStateEvent.idl
Blink: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/events/pop_state_event_init.idl?q=PopStateEventInit%20file:idl


saschanaz:
LGTM, thanks!

PR #2311, opened 2025-12-23
Add xrCompatible

Bashamega:
Blink: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgl/webgl_context_attributes.idl?q=WebGLContextAttributes%20file:idl
Webkit: https://searchfox.org/wubkat/source/Source/WebCore/html/canvas/WebGLContextAttributes.idl


saschanaz:
LGTM, thanks!

PR #2313, opened 2025-12-23
Migrate VideoFrameBufferInit to KDL

Bashamega:
Blink:https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webcodecs/video_frame_buffer_init.idl?q=VideoFrameBufferInit%20file:idl


saschanaz:
LGTM, thanks!

PR #2314, opened 2025-12-24
Migrate IntersectionObserverEntry to KDL

Bashamega:
Also, I added full support for interface removals.
Everything now should be optionally nested ~except for the signature, because the emitter throws an error otherwise. So I had to add special handling for it in the converter~

### DIFF
--- a/src/build/patches.ts
+++ b/src/build/patches.ts
@@ -366,7 +366,7 @@ function handleMethodAndConstructor(
     };
     if (typeof signatureIndex == "number") {
       signature = { [signatureIndex]: signatureObj };
-    } else if (!isConstructor) {
+    } else {
       signature = [signatureObj];
     }
   }


saschanaz:
> Also, I added full support for interface removals. Everything now should be optionally nested except for the signature, because the emitter throws an error otherwise. So I had to add special handling for it in the converter

Can you eleborate, like what kind of error? Working around an error is only good when we know what's happening there.

Bashamega:
> > Also, I added full support for interface removals. Everything now should be optionally nested except for the signature, because the emitter throws an error otherwise. So I had to add special handling for it in the converter
> 
> Can you eleborate, like what kind of error? Working around an error is only good when we know what's happening there.

Scratch that, I have updated the syntax

src/build/patches.ts
```diff
@@ -360,6 +360,10 @@ function handleMethodAndConstructor(
     } else {
       signature = [signatureObj];
     }
+    if (signatureObj.param?.length === 0 && !signatureObj.type) {
+      // If there are no params and no return type, remove the signature
+      signature = undefined;
```
saschanaz: This would cause problem for non-removal patches, no?

Bashamega: This wouldn't cause an issue; we already have patches for non-removal methods, and it is working prefectly

saschanaz: I mean if we want to add a signature... Hmm, but doing so would require a type, so maybe this is fine.

Bashamega: Yes, so this is good. Anything else?

saschanaz: Not "good" but more like "I can live with this for now" for me... but yeah this should be okay.

saschanaz: but constructors don't have a return type, so `constructor` is also a valid non-removal constructor. Doing this would break adding a parameter-less constructor 🤔

(Meaning we should throw if type exists for constructor btw)

inputfiles/patches/intersection-observer.kdl
```diff
@@ -1,4 +1,7 @@
 removals {
+  interface IntersectionObserverEntry {
+   constructor // WebKit-only as of 2025-12
```
saschanaz: also incorrect indentation

saschanaz:
```suggestion
    constructor // WebKit-only as of 2025-12
```

src/build/patches.ts
Bashamega: I will open another one of the throwing an error. It wouldn't count as a removal because the parameters array will be filled. We check for both

saschanaz: Maybe remove it when signatureIndex exists? Having signatureIndex without having params or types don't make sense.

(In that case we'll only remove the constructor/method's signature but not themselves, which I think is fine as it should ultimately behave same)

Bashamega:
Looks good to me 😊

saschanaz:
Nah currently `signatureIndex` is fully ignored because it doesn't enter `type || params.length > 0` check. I'll investigate more.

PR #2315, opened 2025-12-24
Migrate VideoFrameInit to KDL

Bashamega:
Blink: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webcodecs/video_frame_init.idl?q=VideoFrameInit%20file:idl


saschanaz:
LGTM

PR #2318, opened 2025-12-24
Clean up JSON

Bashamega:


inputfiles/addedTypes.jsonc
```diff
@@ -603,10 +603,6 @@
             }
         }
     },
-    "dictionaries": {
```
saschanaz, line 606: So this is obvious...

inputfiles/removedTypes.jsonc
```diff
@@ -309,10 +309,5 @@
                 }
             }
         }
-    },
-    "typedefs": {
-        "typedef": [
-            "ArrayBufferView"
```
saschanaz, line 315: But why is this okay?

ArrayBufferView is referenced by BufferSource that we use, so technically this is needed. But somehow we have been using it after removing it...

That's done by #405 by me 7 years ago, I need to recall what I was thinking about.

saschanaz: Ahhhh because `arrayBufferViewTypes` has ABV so it's okay even if it's not in IDL provided types. But then why is it okay to not remove it? Hmmmmm.

saschanaz (commented):
I need some time to figure this out...

inputfiles/removedTypes.jsonc
saschanaz: Huh because `followTypeReferences` explicitly ignores anything covered by `baseTypeConversionMap`, so even if we have ABV we ignore it because it's part of base types.

saschanaz:
Interesting, LGTM

PR #2319, opened 2025-12-24
Migrate payerErrors to KDL

Bashamega:
Webkit: https://searchfox.org/wubkat/source/Source/WebCore/Modules/paymentrequest/PaymentDetailsUpdate.idl
Gecko: https://searchfox.org/firefox-main/source/dom/webidl/PaymentRequest.webidl#84

### DIFF
--- a/inputfiles/patches/payment-request.kdl
+++ b/inputfiles/patches/payment-request.kdl
@@ -12,6 +12,6 @@ interface PaymentResponse {
 
 removals {
   dictionary PaymentDetailsUpdate {
-    member payerErrors // Webkit only 2025-12
+    member payerErrors // Webkit only as of 2025-12
   }
 }


saschanaz:
Gecko should not be considered as a shipped implementation in this case as it's disabled by default:

* https://searchfox.org/firefox-main/rev/f26084f2dfc00c1e10377d4433cfea594f7ea8c2/dom/payments/PaymentRequest.cpp#81-83
* https://searchfox.org/firefox-main/rev/f26084f2dfc00c1e10377d4433cfea594f7ea8c2/modules/libpref/init/StaticPrefList.yaml#3815-3819

So this is still WebKit only.

Bashamega:
I have updated.
Is there a way for me to quickly check if it is really there? I only check IDL files, and thought they were enough.
I don't know C++

saschanaz:
That's specific to each implementation. In Gecko you may see:

* `Func="mozilla::dom::PaymentRequest::PrefEnabled"`, or
* `Pref="dom.payments.request.enabled"`

`Func` is a bit more complex than `Pref`. In either cases you can search again with the string, and searchfox should show the most relevant thing on the top of the page. And then you would find something starting with `StaticPrefs::` like [this](https://searchfox.org/firefox-main/rev/f26084f2dfc00c1e10377d4433cfea594f7ea8c2/dom/payments/PaymentRequest.cpp#81-83). If you click the line then it shows "Go to definition", which should send you to the StaticPrefList.yaml where you can see the status.

`Pref` is much easier as searching the string should send you right to the StaticPrefList file.

Bashamega:
Thank you very much

saschanaz:
LGTM

PR #2322, opened 2025-12-25
Run TypeScript scripts directly

### LINKED ISSUE #2321: Stop building JS and run TS directly
Native TS support is now stable in Node.js 22 since https://github.com/nodejs/node/pull/57298, we should be able to run it directly without having JS.

saschanaz:
Also removes enums. There's `--experimental-transform-types` for that but that's still experimental and TS team regretted adding enums after all.

Fixes #2321
Closes #2125

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,8 +10,10 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "resolveJsonModule": true,
-    "allowImportingTsExtensions": true,
     "noEmit": true,
+    "rewriteRelativeImportExtensions": true,
+    "erasableSyntaxOnly": true,
+    "verbatimModuleSyntax": true
   },
   "include": ["./src"],
 }


jakebailey:
> TS team regretted adding enums after all.

This isn't really true 😄

deploy/deployChangedPackages.js
jakebailey, line 1: We should really just move these to src tbh

jakebailey: I can't properly review this at the moment but should be able to do so in a few days

jakebailey (commented):
I can't properly review this at the moment but should be able to do so in a few days

saschanaz:
> > TS team regretted adding enums after all.
> 
> This isn't really true 😄

That's what I'm told in some decades old issue, are things changed?

src/build/emitter.ts
```diff
@@ -10,15 +10,11 @@ import {
...
+} from "./helpers.ts";
+import { collectLegacyNamespaceTypes } from "./legacy-namespace.ts";
 
 /// Decide which members of a function to emit
-enum EmitScope {
-  StaticOnly,
-  InstanceOnly,
-  All,
-}
+type EmitScope = "StaticOnly" | "InstanceOnly" | "All";
```
jakebailey, line 17: I'd just use an as const object and the keyof typeof trick

saschanaz: Hmmmmmmmmmmmmmmmm..................... I kinda prefer this way as it's simpler and it's what we do in web APIs 🤔

deploy/deployChangedPackages.js
saschanaz: That's not strictly needed here and can happen separately IMO

jakebailey: Fair enough

tsconfig.json
```diff
@@ -1,17 +1,17 @@
...
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
+    "outDir": "./lib",
+    "strict": true,
+    "esModuleInterop": true,
+    "sourceMap": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "resolveJsonModule": true,
+    "allowImportingTsExtensions": true,
```
jakebailey: We should set these to match https://nodejs.org/api/typescript.html#type-stripping

deploy/jsconfig.json
```diff
@@ -4,6 +4,7 @@
     "module": "esnext", // es2022
     "moduleResolution": "node",
     "strict": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "allowImportingTsExtensions": true
```
jakebailey, line 8: We'll probably want to set the options listed in https://nodejs.org/api/typescript.html#type-stripping

src/build/emitter.ts
jakebailey: Sure, but this was a TS enum before, so IMO this is easier:

```ts
const EmitScope = {
    StaticOnly: "StaticOnly",
    InstanceOnly: "InstanceOnly",
    All: "All",
} as const;
type EmitScope = typeof EmitScope[keyof typeof EmitScope];
```

Then no other code has to change.

But, the strings aren't the worst either

deploy/jsconfig.json
jakebailey: this is jsconfig, sorry, my bad.

saschanaz:
LGTM

jakebailey:
Ah, duh

PR #2328, opened 2025-12-27
Migrate EventListenerObject to KDL

Bashamega:


Bashamega:
Cool 
Please merge

saschanaz:
(we somehow still accept void, I think that can be unshipped at this point. Or not?)

LGTM

PR #2336, opened 2025-12-28
Remove ownerDocument in ProcessingInstruction

Bashamega:
It is the same as in CharacterData

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -26921,12 +26921,6 @@ interface PopoverTargetAttributes {
  */
 interface ProcessingInstruction extends CharacterData, LinkStyle {
     /**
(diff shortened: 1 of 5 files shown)


saschanaz:
This is from Node, one of ancestors. I recommend to check git blame history to see why this is added.

saschanaz (changes requested):
as above

Bashamega:
Sorry for the late reply I was working out
It seems it originates from #819

But It is still not clear to me, no browser has it

saschanaz:
ProcessingInstruction inherits CharacterData which then inherits Node. Node has ownerDocument.

saschanaz:
Hmm. But CharacterData has the same overriding type, so ProcessingInstruction shouldn't need another one. This should be good but the PR description should be different.

Bashamega:
I have updated the pr

saschanaz:
LGTM

PR #2338, opened 2025-12-28
Migrate HTMLInputElement 

Bashamega:


inputfiles/addedTypes.jsonc
```diff
@@ -606,6 +597,10 @@
                 // Full spec at https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill.
                 "name": "AutoFill",
                 "overrideType": "AutoFillBase | `${OptionalPrefixToken<AutoFillSection>}${OptionalPrefixToken<AutoFillAddressKind>}${AutoFillField}${OptionalPostfixToken<AutoFillCredentialField>}`"
+            },
+            {
+                "name": "Directions",
```
HolgerJeromin: Do we have a plan to remove that again?

Bashamega: For now, let's keep it in the JSON until we migrate because we still don't have support for type defs in KDL

saschanaz: 1. Maybe something like SelectionDirection to prevent future conflict.
2. Maybe do the KDL work first to prevent back and forth?

saschanaz:
Can you say maybe HTMLInputElement or something to be more clear for the title?

Bashamega:
Done

src/build/patches.ts
```diff
@@ -326,6 +326,7 @@ function handleParam(node: Node) {
     name,
     ...optionalMember("type", "string", node.properties?.type),
     ...optionalMember("overrideType", "string", node.properties?.overrideType),
+    ...optionalMember("nullable", "boolean", node.properties?.nullable),
```
saschanaz: This is part of Typed and that should be handled by type node.

HolgerJeromin (approved):

src/build/patches.ts

inputfiles/patches/html.kdl
```diff
@@ -47,6 +47,30 @@ interface HTMLImageElement {
...
 }
 
+interface HTMLInputElement {
+  property labels {
+    type NodeListOf {
+      type HTMLLabelElement
+    }
+  }
+  property autocomplete type=AutoFill
+  property valueAsDate type=Date
```
saschanaz, line 57: I'm not sure this should be allowed, Date is not an IDL type. But maybe not worth disallowing it... Let's see.

saschanaz:
LGTM

PR #2342, opened 2025-12-31
Create a Globalized Type

Bashamega:
This PR introduces a reusable generic utility type, Globalized<T>, to model environments where a scoped global (such as WorkerGlobalScope) is augmented with globalThis.
This will allow us to use type instead of override type when using it in KDL or JSONc

### DIFF
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -10867,7 +10867,7 @@ interface WorkerGlobalScope extends EventTarget, FontFaceSource, WindowOrWorkerG
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/WorkerGlobalScope/self)
      */
-    readonly self: Globalized<WorkerGlobalScope>;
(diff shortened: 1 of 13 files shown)


jakebailey:
I don't think we should just be adding random global types like this that aren't from a spec; this seems like it's just the same code but not inlined?

Bashamega:
We already created a similar type in #2338 for SelectionDirection

jakebailey:
That's totally different,no?

Bashamega:
> That's totally different,no?

I think no, I think this is just using the DRY method like that pr

saschanaz:
Adding string enum is OK-ish as users may want to use the same definition without repeating, but in this case it's unlikely that users would ever do this manually.

Bashamega:
Understood @saschanaz

saschanaz:
(also the spec would use string enum if it was added in 2025; it does not because it's added back in the old days when it wasn't using Web IDL and doing so now would cause potential webcompat issue because of order of exception)

PR #2350, opened 2026-01-02
Add linguist support for KDL and jsonc files

Bashamega:
This adds KDL to the list of languages. Check https://github.com/bashamega/TypeScript-DOM-lib-generator

Got this idea from https://github.com/mdn/content/pull/14123

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -62,13 +62,8 @@
 #*.rtf   diff=astextplain
 #*.RTF   diff=astextplain
 
-
 ###############################################################################
-# Show KDL in the language stats 
-###############################################################################
-*.kdl linguist-detectable
-
-###############################################################################
-# Show JSONC in the language stats 
+# Show manual patches in the language stats 
 ###############################################################################
 *.jsonc linguist-detectable
+*.kdl linguist-detectable


saschanaz:
Not sure we want to do that. mdn/content is almost 100% about markdown but this repo is about generating types and that's very typescript. We use KDL, sure, but to augment.

HolgerJeromin:
> We use KDL, sure, but to augment.

The KDL files are no generated part, but written by hand. So IMO it would be nice to have that detected. 
This change does only put the repo from
> TypeScript 86.4%, JavaScript 13.6%

to the far more correct

> TypeScript 71.0%, KDL 17.9%, JavaScript 11.1%

Bashamega:
I agree with @HolgerJeromin

.gitattributes
```diff
@@ -61,3 +61,10 @@
 #*.PDF   diff=astextplain
 #*.rtf   diff=astextplain
 #*.RTF   diff=astextplain
+
+
+###############################################################################
+# linguist support for kdl
+###############################################################################
+*.kdl linguist-detectable=true
```
saschanaz: https://github.com/github-linguist/linguist/blob/a7e40d31e271f6747fa1234df37b5fff3e6e2406/docs/overrides.md#detectable doesn't use `=true` 🤔, so maybe skip it?

.gitattributes
```diff
@@ -61,3 +61,10 @@
...
+*.kdl linguist-language=KDL
```
saschanaz: Do we need to manually give language at all? This exists to reclassify the language, but the default classification should just work.

HolgerJeromin:
> This should include .jsonc too then. But doing so would make it like 90% jsonc/kdl.

Which would be the truth.

> Do we want that? I don't know...

Why not?

Bashamega:
I have updated it.


> This should include `.jsonc` too then. But doing so would make it like 90% jsonc/kdl. Do we want that? I don't know...

Maybe not since we are currently in the progress of removing them

saschanaz:
> Why not?

Because it's not code but augmentations? (I mean, data instead of code)

> Maybe not since we are currently in the progress of removing them

Which still has far way to go, and those are also handwritten so IMO either include both or neither

Bashamega:
JSONC and Typescript are competing LOL

JSON 42.1%
TypeScript 41.1%

Bashamega:
Good change

saschanaz:
(This needs maintainer permission to merge)

jakebailey (approved):
It's weird that GitHub needs this hint but, alrighty

PR #2351, opened 2026-01-02
Migrate OptionalPostfixToken and OptionalPrefixToken

Bashamega:


saschanaz:
This is added part of autofill, so maybe put it together until there's some other usage?

Bashamega:
I have renamed it

saschanaz:
We already have html-autocomplete

Bashamega:
Done

src/build/patches.ts
```diff
@@ -448,7 +449,23 @@ function handleMember(c: Node): DeepPartial<Member> {
...
 function handleTypedef(node: Node): DeepPartial<TypeDef> {
-  const typeNodes = node.children.filter((c) => c.name === "type");
+  const typeNodes: Node[] = [];
+  let typeParameters = {};
+  for (const child of node.children) {
+    switch (child.name) {
+      case "type":
+        typeNodes.push(child);
+        break;
+      case "typeParameters": {
```
saschanaz, line 459: BTW this shouldn't be plural, it should be `typeParameter`. Next PR?

(And `typeParameters` should ultimately be an array)

saschanaz:
LGTM

PR #2353, opened 2026-01-03
Fix TypedParams

### LINKED ISSUE #2352: Rename typeParameters to typeParameter
BTW this shouldn't be plural, it should be `typeParameter`. Next PR?

(And `typeParameters` should ultimately be an array)

_Originally posted by @saschanaz in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2351#discussion_r2658256323_

Opening the issue so I don't forget

Bashamega:
fixes #2352


saschanaz:
This is not a refactoring but a fix 👀

Bashamega:
I have updated the title

Bashamega:
I have fixed the conflicts, can you check it, thank you @saschanaz

src/build/patches.ts
```diff
@@ -242,7 +238,7 @@ function handleMixinAndInterfaces(
   }
 
   const interfaceObject = type === "interface" && {
-    ...typeParameters,
+    ...optionalNestedMember("typeParameters", typeParameters, typeParameters),
```
saschanaz: It's not a nested member, so there's a shorter way:

```suggestion
    ...typeParameters.length : { typeParameters } : {},
```

src/build/patches.ts
```diff
@@ -270,7 +266,7 @@ function handleMixinAndInterfaces(
       "string",
       node.properties?.replaceReference,
     ),
-    ...handleTypeParameters(node.properties?.typeParameters),
+    ...handleTypeParameters(node.properties?.typeParameter),
```
saschanaz: Now the function name doesn't match the parameter. IMO the function signature should be similar to handleTyped.

Bashamega:
I have updated it to make it similar to the handle typed

Bashamega:
Is there any blocker in this pr @saschanaz

PR #2358, opened 2026-01-07
Add X25519 generateKey overload to SubtleCrypto

kairosci:
## Summary

This PR adds a specific overload for the X25519 key exchange algorithm to the `SubtleCrypto.generateKey()` method, following the same pattern as the existing Ed25519 overload.

## Changes

Added the following overload to `SubtleCrypto` interface in `inputfiles/overridingTypes.jsonc`:

```typescript
generateKey(algorithm: "X25519" | { name: "X25519" }, extractable: boolean, keyUsages: ReadonlyArray<"deriveBits" | "deriveKey">): Promise<CryptoKeyPair>;
```

## Rationale

X25519 is used for key derivation (ECDH), so it uses `deriveBits` and `deriveKey` as valid keyUsages, unlike Ed25519 which uses `sign` and `verify`.

This allows TypeScript to correctly infer `CryptoKeyPair` as the return type when using X25519, eliminating the need for type assertions or instanceof checks:

```typescript
// Before: keyPair is CryptoKey | CryptoKeyPair, requires type assertion
const keyPair = await crypto.subtle.generateKey({ name: "X25519" }, true, ["deriveKey"]);

// After: keyPair is CryptoKeyPair, publicKey is directly accessible
const keyPair = await crypto.subtle.generateKey({ name: "X25519" }, true, ["deriveKey"]);
const publicKey = keyPair.publicKey; // No error!
```

## Browser Support

X25519 is supported by:
- **Chrome/Edge** (Blink): Yes, since version 133
- **Firefox** (Gecko): Yes, since version 135
- **Safari** (WebKit): Yes, since version 17.0

This meets the requirement of support by 2+ major browser engines.

## References

- Fixes microsoft/TypeScript#62926
- [MDN: SubtleCrypto.generateKey()](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey)

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -34600,6 +34600,7 @@ interface SubtleCrypto {
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/SubtleCrypto/generateKey)
      */
     generateKey(algorithm: "Ed25519" | { name: "Ed25519" }, extractable: boolean, keyUsages: ReadonlyArray<"sign" | "verify">): Promise<CryptoKeyPair>;
+    generateKey(algorithm: "X25519" | { name: "X25519" }, extractable: boolean, keyUsages: ReadonlyArray<"deriveBits" | "deriveKey">): Promise<CryptoKeyPair>;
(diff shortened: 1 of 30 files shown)


kairosci:

saschanaz:
LGTM, thanks!

(#2050 is a bigger version of it but it's too big)

PR #2361, opened 2026-01-10
Support records in KDL

Bashamega:


src/build/patches.ts
```diff
@@ -59,11 +59,15 @@ function handleSingleTypeNode(type: Node): DeepPartial<Typed> {
...
   }
+  const typeValue = type.values[0];
   const subType =
     type.children.length > 0 ? handleTyped(type.children) : undefined;
   return {
-    ...optionalMember("type", "string", type.values[0]),
-    subtype: subType,
+    ...optionalMember("type", "string", typeValue),
+    subtype:
+      typeValue == "record" && typeof subType?.type !== "string"
```
saschanaz: I think this shouldn't check the type name. We could use this for any parameterized type, right?

Bashamega: I think we should, when I removed that check, it broke the output to be:
```
 readonly links: HTMLCollectionOf<HTMLAnchorElement, HTMLAreaElement>;
```
Instead of the original:
```
readonly links: HTMLCollectionOf<HTMLAnchorElement | HTMLAreaElement>;
```

saschanaz: Oh no, ambiguate way to pass second type vs union. We need to disambiguate it rather than using the same way here then.

Bashamega:
What do you think of the idea of adding the union property, like how I did it in the last commit?

saschanaz:
How do we handle `record<DOMString, (some union type)>` in that case?

Bashamega:
> How do we handle `record<DOMString, (some union type)>` in that case?

I have searched the JSONC files and there is nothing like this. If there is an extremely rare case when this is needed, we can use overrideType

inputfiles/patches/wasm-js-api.kdl
```diff
@@ -0,0 +1,6 @@
+typedef Exports legacyNamespace=WebAssembly {
+  type record union=#true {
```
saschanaz, line 2: But this definitely does not intend to be union, as the intended result is `record<DOMString, ExportValue`?

saschanaz (commented):
Hmm. What about:

```kdl
type record {
  type DOMString
  // nameless type becomes a union
  type {
    type Boolean
    type Number
  }
}
```

because a union type shouldn't have a name, thoughts?

(Not a super fan of that either, as listing types suddenly behaves differently based on the existence of a name...)

But also this PR doesn't have a union type anymore, so maybe we can cover that when we get a union type.

PR #2371, opened 2026-01-11
Migrate MutationRecord to KDL

Bashamega:


Bashamega:
Hello @saschanaz 
What is the blocker here?

PR #2377, opened 2026-01-11
Element error event handler

### LINKED ISSUE #62075: HTMLElement inherits onerror and addListener("error" definitions that are unique to Window
### 🔎 Search Terms

onerror GlobalEventHandlers

### 🕗 Version & Regression Information

As far as I can tell, this has always been wrong. There are a number of issues (and merged fixes) around `onerror` and/or `addEventListener("error"`, but all of them have concerned themselves with getting the definition correct for `Window`.

### ⏯ Playground Link

https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEYD2A7AzgF3gCwwWwgFEIQ8QUMAueACQBUBZAGWNPIwG4AoL3A1shQB0qEDBhIYACikgAbu2qEFFAJTwAvAD54AbwC+qjvAD0J+GmxIArhGDwUSLGIkwuQA

### 💻 Code

```ts
declare const htmlElement: HTMLElement;

htmlElement.onerror((event: Event) => {}); // should not error
```

### 🙁 Actual behavior

The type definitions in `lib.dom.d.ts` for `addEventListener("error"` and `onerror` used by `HTMLElement` match the spec for `Window`, which has unique handling of these events for historical reasons.

### 🙂 Expected behavior

The type definitions in `lib.dom.d.ts` for `addEventListener("error"` and `onerror` event handling on `HTMLElement` would match the spec for `HTMLElement`.

### Additional information about the issue

In `lib.dom.d.ts`, we have `GlobalEventHandlersEventMap`, which is used to define `addEventListener`, and `GlobalEventHandlers`, used to define the various `on${string}` event handler properties. As the name suggests, these are meant to handle “global” events, and `Document`, `Window`, `HTMLElement`, `SVGElement`, and `MathMLElement` all inherit these.

Unfortunately, the definitions here for `addEventListener("error"` and `onerror` are used uniquely by `Window`. None of the other types inheriting these definitions are correct for error events.

These are the definitions:

```ts
interface GlobalEventHandlersEventMap {
    // …
    "error": ErrorEvent;
    // …
}

interface GlobalEventHandlers {
    // …
    /**
     * Fires when an error occurs during object loading.
     * @param ev The event.
     *
     * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement/error_event)
     */
    onerror: OnErrorEventHandler;
    // …
}

// …

type OnErrorEventHandler = OnErrorEventHandlerNonNull | null;

// …

interface OnErrorEventHandlerNonNull {
    (event: Event | string, source?: string, lineno?: number, colno?: number, error?: Error): any;
}
```

Note that MDN link for `onerror` that specifically directs to `HTMLElement/error_event`: that page says,

> #### Syntax
> Use the event name in methods like [`addEventListener()`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener), or set an event handler property.
>
> ```js
> addEventListener("error", (event) => { })
>
> onerror = (event) => { }
> ```
>
> #### Event type
> The event object is a [`UIEvent`](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent) instance if it was generated from a user interface element, or an [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event) instance otherwise.

Just a single argument that is either `Event` or `UIEvent`. Both the five-argument `onerror` definition and the single `ErrorEvent` argument `"error"` definition are wrong for `HTMLElement`.

Instead, those definitions are from [`Window`](https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event), and are defined correctly for `Window`, including the discrepancy in the arguments between `addEventListener` and `onerror`. Well, the `GlobalEventHandlersEventMap` definition is possibly less than precise:

> The event object is a [`ErrorEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) instance if it was generated from a user interface element, or an [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event) instance otherwise.

However, the page notes that all of this is unique to `Window`:

> **Note:** For historical reasons, `onerror` on `Window` and [`WorkerGlobalScope`](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope) objects is the only event handler property that receives more than one argument.

> The event handler's signature is asymmetric between `addEventListener()` and `onerror`. The event handler passed to `Window.addEventListener()` receives a single [`ErrorEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) object, while the `onerror` handler receives five arguments, matching the [`ErrorEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) object's properties:

> This special behavior only happens for the `onerror` event handler on `window`. The [`Element.onerror`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/error_event) handler still receives a single [`ErrorEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) object.

(Side note: the “`Element.onerror`” link in that last quote actually goes to `HTMLElement`, and according to that page, it’s a `UIEvent` or `Event`, not an `ErrorEvent`, so MDN has some issues here, too.)

Thus, these definitions should only be applied to `Window` in `lib.dom.d.ts`, and `onerror`/`addEventListener("error"` for `HTMLElement`, `SVGElement`, and `MathMLElement` should be defined differently. (I can’t actually _find_ definitions of `onerror` for `MathMLElement` on MDN, nor for `Document` which I also see using `GlobalEventHandlers` in `lib.dom.d.ts`, but I haven’t dug into those cases much.)

cosmic-pixel-painter:
This PR fixes https://github.com/microsoft/TypeScript/issues/62075. and is a result of the comment left on the PR over [here](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2272/files#r2653546804), which I closed. Thanks for the advice on that 🙏 . I re-read your comment and figured out a path forward. I also dropped `UIEvent` from this PR.

## Recap
As a recap, this handles this [bug report here](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2272). The issue is that `HTMLElement.onerror` (and other element` onerror` handlers) uses the incorrect interface `OnErrorEventHandler`, which has the signature: 
```ts
type OnErrorEventHandler = (
  event: Event | string,
  source?: string,
  lineno?: number,
  colno?: number,
  error?: Error
) => any;
```

This causes a type error when assigning a Event-based handler:
```ts
const div = document.createElement("div");
div.onerror = (event: Event) => {  // TS Error!
  // Type '(event: Event) => void' is not assignable to type 'OnErrorEventHandler'.
  // Types of parameters 'event' and 'event' are incompatible.
  // Type 'string | Event' is not assignable to type 'Event'.
};
```

It should be that `window.onerror` uses `OnErrorEventHandler` and elements uses the `Event` object.

## Changes
- I ended up changing `GlobalEventHandlers.onerror` to use an `event-based` handler instead of `OnErrorEventHandler`. I did this by updating `events.kdl` to use `Event` instead of `ErrorEvent`.
- I then overrode `window.onerror` to keep using `OnErrorEventHandler` since `Window` is the only place where this handler signature is expected.
- This caused duplicate declarations since `Window` extends `GlobalEventHandler`s and both define `onerror` differently. This was happening in the global scope "polluter". I fixed this by tracking already emitted members in `emitAllMembers`.

## Before/After
Before:

```ts
interface GlobalEventHandlers {
  onerror: OnErrorEventHandler;
}
// Window inherits OnErrorEventHandler from GlobalEventHandlers
```
After:

```ts
interface GlobalEventHandlers {
  onerror: ((this: GlobalEventHandlers, ev: Event) => any) | null;
}

interface Window extends GlobalEventHandlers {
  onerror: OnErrorEventHandler;  // Explicit override
}
```

## Testing
I added the `onerror.ts` test file and updated the `eventlistener.ts` test file.

## Other Thoughts
This has been a crazy and fun bug to hunt down. I think I nailed it this time 😂


inputfiles/overridingTypes.jsonc
```diff
@@ -427,9 +427,22 @@
                             // making it nullable breaks compatibility as it has historically been non-nullable in TS lib.
                             // See the example in https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-parent.
                             "nullable": false
+                        },
+                        "onerror": {
+                            "name": "onerror",
+                            "overrideType": "OnErrorEventHandler",
```
saschanaz:
```suggestion
                            "type": "OnErrorEventHandler",
```

overrideType is for things that are not expressed in Web IDL but strictly in TypeScript

saschanaz: Also this should stay in KDL?

saschanaz: Hmm, probably didn't do that because of `eventHandler`.

cosmic-pixel-painter: I just tried this and then we run into this issue
Screenshot 2026-01-11 at 3 30 34 PM
without the override, the types conflict

cosmic-pixel-painter:
@saschanaz I was able to fix it using the KDL file and didn't need to touch the `overrideType` 🎉 . I think this nails it 🔨

cosmic-pixel-painter:
Hey @saschanaz it's ready for another look. Let me know what you think 😄

cosmic-pixel-painter:
I hope you're doing well @saschanaz. This PR is ready for a another review when you have a moment.

Issue #1557, opened 2023-05-08
web Video types


jozefchutka:
separating issues in https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1556

existing VideoEncoderConfig to be modified (adding `undefined`):

```ts
interface VideoEncoderConfig {
    alpha?: AlphaOption | undefined;
    avc?: AvcEncoderConfig | undefined;
    bitrate?: number | undefined;
    bitrateMode?: VideoEncoderBitrateMode | undefined;
    codec: string;
    displayHeight?: number | undefined;
    displayWidth?: number | undefined;
    framerate?: number | undefined;
    hardwareAcceleration?: HardwareAcceleration | undefined;
    height: number;
    latencyMode?: LatencyMode | undefined;
    scalabilityMode?: string | undefined;
    width: number;
}
```

same for VideoFrameBufferInit:

```ts
interface VideoFrameBufferInit {
    codedHeight: number;
    codedWidth: number;
    colorSpace?: VideoColorSpaceInit | undefined;
    displayHeight?: number | undefined;
    displayWidth?: number | undefined;
    duration?: number | undefined;
    format: VideoPixelFormat;
    layout?: PlaneLayout[] | undefined;
    timestamp: number;
    visibleRect?: DOMRectInit | undefined;
}
```

and VideoDecoderConfig:

```ts
interface VideoDecoderConfig {
    codec: string;
    codedHeight?: number | undefined;
    codedWidth?: number | undefined;
    colorSpace?: VideoColorSpaceInit | undefined;
    description?: BufferSource | undefined;
    displayAspectHeight?: number | undefined;
    displayAspectWidth?: number | undefined;
    hardwareAcceleration?: HardwareAcceleration | undefined;
    optimizeForLatency?: boolean | undefined;
}
```

while its unexpected a `VideoFrame` to have `format=null` https://developer.mozilla.org/en-US/docs/Web/API/VideoFrame/format

```ts
interface VideoFrame {
...
    readonly format: VideoPixelFormat;
...
}
```

saschanaz:
I don't think explicit `undefined` is needed for optional properties?

jozefchutka:
Here is how I need to create config following current types:

```ts
const config:VideoEncoderConfig = {codec, width, height};
if(alpha !== undefined)
    config.alpha = alpha;
if(avc !== undefined)
    config.avc = avc;
if(bitrate !== undefined)
[...]

PR #2378, opened 2026-01-14
[wontfix] feat(emitter): generalize optional dictionary members to allow explicit undefined

kairosci:
### Description

This PR updates the emitter logic to consistently handle optional dictionary members by explicitly allowing `undefined` in their type definitions. Following the maintainer's feedback, I have shifted from a specific fix for `WebCodecs` to a **generalized emitter change**, avoiding manual KDL patches or redundant entries in `overridingTypes.jsonc`.

### Key Changes

1.  **Generalized Emitter Logic**: Modified `src/build/emitter.ts` to automatically append `| undefined` to all optional dictionary members during the emission process.
2.  **Smart Deduplication**: Implemented a check within the emitter to ensure that `| undefined` is only added if not already present in the type union (e.g., preventing cases like `type?: undefined | undefined`).
3.  **Removal of Redundant Overrides**: Removed all manual `| undefined` overrides for dictionary members from `overridingTypes.jsonc`, as they are now handled natively by the generator.
4.  **Baseline Synchronization**: Regenerated all baseline files against the latest `main` branch to resolve conflicts and ensure consistency across all supported TypeScript versions.

### Why this is important?

*   **Consistency**: Interface properties in this project already follow the `prop?: T | undefined` pattern. This change brings Dictionaries (configuration objects) into alignment with that standard.
*   **Support for `exactOptionalPropertyTypes`**: In strict TypeScript configurations where `exactOptionalPropertyTypes` is enabled, a property defined as `prop?: T` cannot be explicitly assigned `undefined`. By changing this to `prop?: T | undefined`, we allow developers to explicitly pass `undefined` to configuration dictionaries, which is a common and often necessary pattern in Web APIs.
*   **Alignment with WebIDL Semantics**: WebIDL optionality often implies that `undefined` is a valid value. This change accurately reflects those semantics in the generated TypeScript definitions.
*   **Maintainability**: By centralizing this logic in the emitter rather than using KDL patches or JSON overrides, we keep the codebase cleaner and reduce the risk of future regressions.

### Implementation Details

The PR is organized into two functional commits:
- `feat(emitter): generalize optional dictionary members to allow explicit undefined`: Contains the core logic changes in the generator.
- `refactor(overrides): update baselines with generalized optional dictionary members`: Contains the baseline updates, fully synced with the current state of the upstream repository.

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -1580,9 +1580,11 @@ export function emitWebIdl(
         .sort(compareName)
         .forEach((m) => {
           emitComments(m, printer.printLine);
-          printer.printLine(
-            `${m.name}${m.required ? "" : "?"}: ${convertDomTypeToTsType(m)};`,
-          );
+          let type = convertDomTypeToTsType(m);
+          if (!m.required && !type.split(" | ").includes("undefined")) {
+            type += " | undefined";
+          }
+          printer.printLine(`${m.name}${m.required ? "" : "?"}: ${type};`);
         });
     }
     if (dict.overrideIndexSignatures) {
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -371,17 +371,39 @@ export function emitWebIdl(
     obj: Browser.Typed,
     forReturn: boolean,
   ): string {
+    let type: string;
     if (obj.overrideType) {
-      return obj.nullable ? makeNullable(obj.overrideType) : obj.overrideType;
+      type = obj.overrideType;
+    } else {
+      if (!obj.type) {
+        throw new Error("Missing 'type' field in " + JSON.stringify(obj));
+      }
+      type = convertDomTypeToTsTypeWorker(obj, forReturn);
+      if (type === "Promise<undefined>") {
+        type = "Promise<void>";
+      }
     }
-    if (!obj.type) {
-      throw new Error("Missing 'type' field in " + JSON.stringify(obj));
+
+    if (obj.additionalTypes) {
+      const additional = obj.additionalTypes
+        .filter((t) => t !== "undefined")
+        .map((t) => convertDomTypeToTsTypeSimple(t));
+      if (additional.length > 0) {
+        type = distinct([type, ...additional]).join(" | ");
+      }
     }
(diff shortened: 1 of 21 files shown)


Bashamega (commented):
Please use KDL, and DO NOT us OverrideType

saschanaz:
This should not be webcodecs specific but be a general emitter change. KDL should not be used either.

kairosci:
> This should not be webcodecs specific but be a general emitter change. KDL should not be used either.

I have updated this PR to address your feedback. 

I have shifted the implementation from a WebCodecs-specific fix to a generalized emitter change in src/build/emitter.ts.

The generator now automatically appends | undefined to all optional dictionary members, ensuring consistency with how interface properties are handled. 

This approach eliminates the need for KDL patches or manual overrides in overridingTypes.jsonc, which I have cleaned up.

I have also synchronized all baselines with the latest main branch and organized the work into two clean, functional commits. This change should correctly support exactOptionalPropertyTypes for all configuration dictionaries across the board.

Bashamega (approved):

src/build/emitter.ts
```diff
@@ -1580,9 +1580,11 @@ export function emitWebIdl(
         .sort(compareName)
         .forEach((m) => {
           emitComments(m, printer.printLine);
-          printer.printLine(
-            `${m.name}${m.required ? "" : "?"}: ${convertDomTypeToTsType(m)};`,
-          );
+          let type = convertDomTypeToTsType(m);
+          if (!m.required && !type.split(" | ").includes("undefined")) {
```
saschanaz: This should probably pass a modified type to the conversion function rather than modifying the stringified result.

kairosci: I think I have satisfied this change too.

Now there are no more concatenations, but the management of | undefined, is handled natively, passing it directly to the conversion function.

kairosci:
@saschanaz PTAL, thanks!

saschanaz:
I think we should do this, but we can't do this right now per @jakebailey as TS team is going to freeze code changes to prepare for TS 6/7. See also https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1557#issuecomment-3751110602.

PR #2399, opened 2026-02-03
Fix Iterable generation caused by iterable dts merging

### LINKED ISSUE #2386: TS 6.0 type regressed the iterable parameters
dom.iterable.generated.d.ts has extra `Iterable<T>` signatures but somehow 6.0 lib doesn't get that but instead is getting a second copy of `T[]` signature.

@Bashamega, want to take a look?

jakebailey:
When I modified the code to merge Iterable in for 6.0, I failed to notice that `iterable` was closed over and so the funcs I added extra calls to don't do anything.

This isn't an incredible fix but it does work. Probably a restructuring is in order but I want the bug fixed.

Fixes #2386

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -2046,7 +2046,7 @@ type ReadableStreamType = "bytes";
 
 interface MessageEvent<T = any> {
     /** @deprecated */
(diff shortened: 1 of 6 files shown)


saschanaz:
Yeah it depends on a surprising side effect, but I guess it's fine for now. LGTM

PR #2402, opened 2026-02-07
Add GPUQueue

saschanaz:


saschanaz:
So many dictionaries that are not autochecked. I really really need some way to make those autochecked... If that existed this would have been merged months ago.

saschanaz:
LGTM

PR #2405, opened 2026-02-08
Add GPUDevice except methods

saschanaz:
Because each method brings it's own dictionary and that becomes massive...


saschanaz:
LGTM

PR #2407, opened 2026-02-09
Bring pushManager back to ServiceWorkerRegistration

### LINKED ISSUE #2406: pushManager is missing from ServiceWorkerRegistration
This happened in https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/2190

Leading to:

```
Error in serviceworker-webpack-plugin
Error: 
/home/runner/work/DefinitelyTyped/DefinitelyTyped/types/serviceworker-webpack-plugin/serviceworker-webpack-plugin-tests.ts
  49:22  error  TypeScript@6.0 compile error: 
Property 'pushManager' does not exist on type 'ServiceWorkerRegistration'  @definitelytyped/expect
  53:22  error  TypeScript@6.0 compile error: 
Property 'pushManager' does not exist on type 'ServiceWorkerRegistration'  @definitelytyped/expect

✖ 2 problems (2 errors, 0 warnings)
```

But, it seems like it should be there: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/pushManager

jakebailey:
Fixes #2406, I think...

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -28320,6 +28320,12 @@ declare var PushManager: {
     readonly supportedContentEncodings: ReadonlyArray<string>;
 };
 
+/** Available only in secure contexts. */
+interface PushManagerAttribute {
+    /** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ServiceWorkerRegistration/pushManager) */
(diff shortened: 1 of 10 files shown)


saschanaz:
LGTM, thanks! My bad 🤦🏻‍♀️

PR #2408, opened 2026-02-09
Add GPUDevice methods except creators

saschanaz:
Because the creator methods are the real culprits for the (remaining) massive additions...


saschanaz:
LGTM

PR #2425, opened 2026-02-16
Make chunk param of ReadableStreamDefaultController.enqueue non-optional

### LINKED ISSUE #1682: Make the chunk parameter of ReadableStreamDefaultController.enqueue non optional
The Web IDL definition specifies an `enqueue` method for the [ReadableStreamDefaultController](https://streams.spec.whatwg.org/#readablestreamdefaultcontroller) class with the following signature:

```
undefined enqueue(optional any chunk);
```
The `chunk` parameter is optional which probably is fine since chunk is `any`.

However, this is a little problematic with TypeScript generics.
Currently, the generated TypeScript interface looks like this:
```
interface ReadableStreamDefaultController<R = any> {
    enqueue(chunk?: R): void;
}
```
This allows `undefined` to be enqueued into the stream while the reader of the stream still expects to read `R` out of it.

This is demonstrated in the following example:
```typescript
const stream = new ReadableStream<string>({
  start: (controller) => {
    // currently the signature is
    // ReadableStreamDefaultController<string>.enqueue(chunk?: string | undefined): void
    // undefined is enqueued here
    controller.enqueue();
  },
});

// ReadableStreamDefaultReader<string>
const reader = stream.getReader();

// ReadableStreamReadResult<string>
const result = await reader.read();
if (result.done === false) {
  // result is narrowed down to ReadableStreamReadValueResult<string>

  // TypeScript assumes ReadableStreamReadValueResult<string>.value is string
  // so the compilation will succeed
  // but result.value is actually undefined
  result.value.toUpperCase();
}
```
To address this, could we consider overriding the `chunk` parameter to be required? 
There is a similar issue with [TransformStreamDefaultController](https://streams.spec.whatwg.org/#transformstreamdefaultcontroller).

ikeyan:
Fixes #1682

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -1031,7 +1031,7 @@ interface ReadableStreamDefaultController<R = any> {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ReadableStreamDefaultController/enqueue)
      */
-    enqueue(chunk?: R): void;
+    enqueue(chunk: R): void;
     /**
      * The **`error()`** method of the ReadableStreamDefaultController interface causes any future interactions with the associated stream to error.
      *
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -30349,7 +30349,7 @@ interface ReadableStreamDefaultController<R = any> {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ReadableStreamDefaultController/enqueue)
      */
-    enqueue(chunk?: R): void;
+    enqueue(chunk: R): void;
     /**
      * The **`error()`** method of the ReadableStreamDefaultController interface causes any future interactions with the associated stream to error.
      *
--- a/baselines/serviceworker.generated.d.ts
+++ b/baselines/serviceworker.generated.d.ts
@@ -8153,7 +8153,7 @@ interface ReadableStreamDefaultController<R = any> {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ReadableStreamDefaultController/enqueue)
      */
-    enqueue(chunk?: R): void;
+    enqueue(chunk: R): void;
     /**
      * The **`error()`** method of the ReadableStreamDefaultController interface causes any future interactions with the associated stream to error.
      *
--- a/baselines/sharedworker.generated.d.ts
+++ b/baselines/sharedworker.generated.d.ts
@@ -7736,7 +7736,7 @@ interface ReadableStreamDefaultController<R = any> {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ReadableStreamDefaultController/enqueue)
      */
-    enqueue(chunk?: R): void;
+    enqueue(chunk: R): void;
     /**
      * The **`error()`** method of the ReadableStreamDefaultController interface causes any future interactions with the associated stream to error.
      *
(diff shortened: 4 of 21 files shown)


ikeyan:

saschanaz:
@MattiasBuelens thoughts? Why is `enqueue()` accepting optional parameter at all? Who's calling it without any argument?

MattiasBuelens:
Hmm, this wasn't optional when I made #541. This looks like a regression from f817b6e28d08153fb21018e706b7e8230b93d1b3, [see diff](https://github.com/microsoft/TypeScript-DOM-lib-generator/commit/f817b6e28d08153fb21018e706b7e8230b93d1b3#diff-d57c51d024219b9f37e6b42df127db25f700acbaade23ce1edcecda2853b995cR11317). Not sure how that happened?

Anyway, `chunk` should definitely be a required argument. 😉

saschanaz:
But it's optional in the spec, why?

saschanaz:
For now let's merge this. whatever the spec intended to do probably is for a very edge case.

LGTM

saschanaz:
(And thank you all!)

MattiasBuelens:
> But it's optional in the spec, why?

I think the spec wants you to be able to write `controller.enqueue(undefined)` as just `controller.enqueue()`. Without the `optional`, that second call would throw a "not enough arguments" `TypeError`.

From a type checking perspective, this only make senses if the `R` in `ReadableStreamDefaultController<R>` is allowed to be `undefined` (i.e. it is of the form `T | undefined`). I don't think there's a way [to cleanly express that in TypeScript right now](https://www.typescriptlang.org/play/?ts=5.9.3#code/JYOwLgpgTgZghgYwgAgMIHtxXQGx9AHgCUA+ZAbwChkbkIQBHAVwhYAoEALJkAawC5kRAJSCAbumAATANyUAvpUpSICHHCgoEmAM5hk2rLnxRBGI3kIgmAWwBG0ZAB9kPFTFAQpJOYbDZLKAA6emZWCDZhX0x-Y2gQxhZ2NwgPEC8oyj8AkwSw9gAWACZMoA), unless maybe with some weird conditional types?

I think it's still better to deviate from the Web IDL in this case and ignore the `optional`. The `R` generic type parameter is unlikely to allow `undefined` in real-world code, so it's better to catch more type errors by making it required. 🙂

saschanaz:
But `any` should accept `undefined` in Web IDL, no?

saschanaz:
Oh, maybe it predates `undefined` being a separate type.

MattiasBuelens:
> But `any` should accept `undefined` in Web IDL, no?

Yes, `any` includes `undefined`. (I don't think `any or undefined` or `any?` is allowed in Web IDL.)

There's no difference in what is *accepted* by `any` and `optional any`, it only matters whether you have to explicitly pass that parameter or whether you can leave it out.

I had a look at other APIs that accept a single `any` parameter. I found `structuredClone`, which is [defined in Web IDL](https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope-mixin) as:
```
interface mixin WindowOrWorkerGlobalScope {
  any structuredClone(any value, optional StructuredSerializeOptions options = {});
}
```
Note that `undefined` is a valid value:
```javascript
structuredClone(undefined)
// -> undefined
```
But since `value` is not optional, this doesn't work:
```javascript
structuredClone()
// -> TypeError: Failed to execute 'structuredClone' on 'Window': 1 argument required, but only 0 present.
```

Now I'm not so sure if `controller.enqueue()` should be allowed... 🤔

ikeyan:
@MattiasBuelens 
We can let it "enqueue with no argument" by allowing `void` in `R` (e.g. `R = T | void`). In that case, `enqueue()` is just "enqueue a void chunk":

```ts
new ReadableStream<Uint8Array | void>({
  start(controller) {
    controller.enqueue();
    controller.enqueue(new Uint8Array([1, 2, 3]));
    controller.close();
  },
});
```

This is the same pattern commonly used with `Promise<void>`, where `resolve()` can be called without an argument.

For the typical `R` that doesn’t include void/undefined, requiring chunk is the safer typing.

PR #2430, opened 2026-02-23
Update @mdn/browser-compat-data and keep-alive WebTransportWriter

jakebailey:
WebTransportWriter is apparently missing info.

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -1414,6 +1414,17 @@ declare var URLSearchParams: {
     new(init?: string[][] | Record<string, string> | string | URLSearchParams): URLSearchParams;
 };
 
+/** Available only in secure contexts. */
+interface WebTransportWriter extends WritableStreamDefaultWriter {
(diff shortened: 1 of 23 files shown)


saschanaz:
Where is WebTransportWriter from? I don't see any implementation?

jakebailey:
https://w3c.github.io/webtransport/#webtransportwriter is where this seems to be from

Bashamega:
> [w3c.github.io/webtransport#webtransportwriter](https://w3c.github.io/webtransport/#webtransportwriter) is where this seems to be from

But there is no actual implementation in the browser
check:
https://source.chromium.org/search?q=WebTransportWriter&sq=
https://searchfox.org/firefox-main/search?q=WebTransportWriter&path=&case=false&regexp=false

jakebailey:
I've got no idea, if this PR is wrong that's fine, I was just trying to do dep updates here (unrelated to the actual generated code) and hit the same thing that's blocking the automatic updater.

saschanaz:
I think the WT spec upgraded an existing method or property to use a new sub-interface, we should probably override that.

PR #2431, opened 2026-02-25
Fully merge iterable types into existing definitions

jakebailey:
The code to do it seems nasty (largely just had copilot to it), but this does merge everything together instead of just tacking the types onto the end.

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -260,7 +260,7 @@ declare var AbortSignal: {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/AbortSignal/any_static)
      */
-    any(signals: AbortSignal[]): AbortSignal;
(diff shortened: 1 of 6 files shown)


jakebailey:
Not as nasty as I claimed, I think; it's just sort of the same code shifted outward. I went through and everything seems to be there for all declarations that moved / were replaced.

Bashamega:
One question, if we are going to merge them, why do we still need the *iterable.ts?

jakebailey:
The stubs exist because people could have written the name explicitly in tsconfig or reference directive. Otherwise <6.0 doesn't have this merge.

(This PR just moves the declarations up from the bottom, since my original change for this was just a hack to tack them on at the bottom.)

Bashamega:
Got it, thanks for the reply

jakebailey:
https://github.com/microsoft/TypeScript/pull/63193 implies this will cause more breaks. Nevermind.

PR #2451, opened 2026-03-14
Manually update to idl@3.74.0

saschanaz:
Also ran `npm audit fix`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,4 +15,4 @@ jobs:
       - run: npm ci
       - run: npm run build
       - run: npm test
-      - run: npx tsgo --stableTypeOrdering
+      - run: npx tsgo


saschanaz:
tsgo fails?

src/build/patches.ts
```diff
@@ -21,9 +21,11 @@ import type {
...
 import { merge } from "./helpers.ts";
 
-type DeepPartial<T> = T extends object
-  ? { [K in keyof T]?: DeepPartial<T[K]> }
-  : T;
+type DeepPartial<T> = T extends any[]
+  ? Record<number, DeepPartial<T>> | DeepPartial<T>[]
+  : T extends object
+    ? { [K in keyof T]?: DeepPartial<T[K]> }
+    : T;
```
saschanaz, line 28: tsgo hates this change, should I care? @jakebailey

saschanaz: https://github.com/microsoft/typescript-go/issues/2609, but `--stableTypeOrdering` doesn't seem to help.

HolgerJeromin: The option brings ts6 to use the ts7 ordering.
So in this case it would only "activate" the bug in ts6, too. 
But also you could try to fix the crash with ts6 without ts7

saschanaz: Hmm, interesting. TS6 hates that too regardless of the flag.

saschanaz:
LGTM

src/build/patches.ts
jakebailey: I'm sure that linked issue is separate; it's a very generic issue.

A lot of these types only worked accidentally :\

PR #2452, opened 2026-03-17
Make chunk param of WritableStreamDefaultWriter.write  and TransformStreamDefaultController.enqueue non-optional

ikeyan:
Follow up #2425
The issue was first mentioned in #1682

## Test script

This script demonstrates that calling `enqueue()` without arguments is behaviorally identical to `enqueue(undefined)`. This confirms that allowing the omission of arguments in these methods poses a risk to type safety, unless the stream's generic type is explicitly defined to include `undefined` or `void`.

I successfully ran the script on Node.js v24.13.1, Chrome 145.0.7632.160, Firefox 148.0.2, Safari 26.3.1

```js
function assert(condition, message) {
  if (!condition) {
    throw new Error("Assertion failed: " + message);
  }
}

async function runTests() {
  // === TransformStream Test ===
  const ts = new TransformStream({
    start(controller) {
      controller.enqueue();          // No arguments
      controller.enqueue(undefined); // Explicit undefined
    }
  });

  const reader = ts.readable.getReader();
  const tsResult1 = await reader.read();
  const tsResult2 = await reader.read();

  assert(tsResult1.value === undefined, "enqueue() should result in undefined");
  assert(tsResult2.value === undefined, "enqueue(undefined) should result in undefined");

  // === WritableStream Test ===
  const writtenChunks = [];
  const ws = new WritableStream({
    write(chunk) {
      writtenChunks.push(chunk); // Store written chunks for verification
    }
  });

  const writer = ws.getWriter();
  await writer.write();          // No arguments
  await writer.write(undefined); // Explicit undefined
  await writer.close();

  assert(writtenChunks[0] === undefined, "write() should write undefined");
  assert(writtenChunks[1] === undefined, "write(undefined) should write undefined");
  assert(writtenChunks.length === 2, "exactly 2 elements are written");
}

// Execute tests. If success, no output.
await runTests();
```

### DIFF
--- a/inputfiles/overridingTypes.jsonc
+++ b/inputfiles/overridingTypes.jsonc
@@ -41,9 +41,7 @@
                                         {
                                             "name": "id",
                                             "optional": false,
-                                            "additionalTypes": [
-                                                "undefined"
-                                            ]
+                                            "additionalTypes": ["undefined"]
                                         }
                                     ]
                                 }
@@ -75,9 +73,7 @@
                                         {
                                             "name": "id",
                                             "optional": false,
-                                            "additionalTypes": [
-                                                "undefined"
-                                            ]
+                                            "additionalTypes": ["undefined"]
                                         }
                                     ]
                                 }
@@ -1098,13 +1094,13 @@
                 }
             },
             "FontFace": {
-                "properties": {
-                    "property": {
-                        "display": {
-                            "type": "FontDisplay"
-                        }
-                    }
+              "properties": {
+                "property": {
+                  "display": {
+                    "type": "FontDisplay"
+                  }
                 }
+              }
             },
             "Element": {
                 "methods": {
@@ -1225,11 +1221,7 @@
                                     "param": [
                                         {
                                             "name": "algorithm",
-                                            "additionalTypes": [
-                                                "EcdhKeyDeriveParams",
-                                                "HkdfParams",
-                                                "Pbkdf2Params"
-                                            ]
+                                            "additionalTypes": ["EcdhKeyDeriveParams", "HkdfParams", "Pbkdf2Params"]
                                         }
                                     ]
                                 }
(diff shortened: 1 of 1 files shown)


inputfiles/overridingTypes.jsonc
```diff
@@ -73,7 +75,9 @@
                                         {
                                             "name": "id",
                                             "optional": false,
-                                            "additionalTypes": ["undefined"]
+                                            "additionalTypes": [
```
Bashamega: Why this change?

inputfiles/overridingTypes.jsonc
```diff
@@ -1094,13 +1098,13 @@
                 }
             },
             "FontFace": {
-              "properties": {
-                "property": {
-                  "display": {
-                    "type": "FontDisplay"
-                  }
+                "properties": {
```
Bashamega: Why the change?

inputfiles/overridingTypes.jsonc
```diff
@@ -1221,7 +1225,11 @@
                                     "param": [
                                         {
                                             "name": "algorithm",
-                                            "additionalTypes": ["EcdhKeyDeriveParams", "HkdfParams", "Pbkdf2Params"]
+                                            "additionalTypes": [
```
Bashamega: Why?

inputfiles/overridingTypes.jsonc
ikeyan: 😨
Fixed!

inputfiles/overridingTypes.jsonc
ikeyan: Fixed!

inputfiles/overridingTypes.jsonc
ikeyan: Fixed!

saschanaz:
Thanks all! LGTM

PR #2460, opened 2026-04-01
fix: Allow CSSStyleValue in Keyframe index signature

ngtrnhao:
Fixes microsoft/TypeScript #63325


ngtrnhao:

PR #2467, opened 2026-04-06
Add Element.setHTML() to addedTypes

victorhuangwq:
The setHTML() method is defined in the Sanitizer API spec (https://wicg.github.io/sanitizer-api/) but is currently missing from @webref/idl, as setHTML has not migrated over to the WHATWG spec yet.

However, there's already 2 engine support, Chrome 146+ (Blink) and Firefox 148+ (Gecko), meeting the 2-engine requirement. https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -2694,6 +2694,10 @@ interface SecurityPolicyViolationEventInit extends EventInit {
     violatedDirective?: string;
 }
 
+interface SetHTMLOptions {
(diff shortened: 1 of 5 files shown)


Bashamega (commented):
Please migrate it to KDL format. It is the new standard

saschanaz:
> but is currently missing from @webref/idl,

This is a spec issue. I filed https://github.com/WICG/sanitizer-api/issues/391. We should not add things manually.

PR #2472, opened 2026-04-19
Manually update to idl@3.75.2

saschanaz:
wasm support is not clear from IDL, but:

https://github.com/WebAssembly/spec/pull/2084
https://bugzilla.mozilla.org/show_bug.cgi?id=1965217

... suggest at least two implementations.


saschanaz:
LGTM

PR #2474, opened 2026-04-20
Mention libReplacement in READMEs

### LINKED ISSUE #2473: Call out that `libReplacement: true` is needed for `@typescript/lib-dom` override
https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/

> libReplacement is now false by default: This flag previously incurred a large number of failed module resolutions for every run, which in turn increased the number of locations we needed to watch under --watch and editor scenarios. In a new project, libReplacement never does anything until other explicit configuration takes place, so it makes sense to turn this off by default for the sake of better performance by default.

README for each npm package should call it out.

jakebailey:
Fixes #2473


saschanaz (approved):
Thanks!

PR #2476, opened 2026-04-30
Make HeadersInit use readonly types

DavidANeil:


DavidANeil:
@microsoft-github-policy-service agree company="Lucid Software"

saschanaz:
I don't like how it uses overrideType... but maybe that's the easiest. Can you add unit tests?

PR #2482, opened 2026-05-09
Manually update to idl@3.77.0

saschanaz:
Closes #2481 
Closes #2467

### DIFF
--- a/baselines/ts5.9/dom.generated.d.ts
+++ b/baselines/ts5.9/dom.generated.d.ts
@@ -17994,7 +17994,7 @@ interface HTMLElementEventMap extends ElementEventMap, GlobalEventHandlersEventM
  *
  * [MDN Reference](https://developer.mozilla.org/docs/Web/API/HTMLElement)
  */
-interface HTMLElement extends Element, ElementCSSInlineStyle, ElementContentEditable, GlobalEventHandlers, HTMLOrSVGOrMathMLElement {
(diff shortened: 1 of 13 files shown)


baselines/ts5.9/dom.generated.d.ts
```diff
@@ -20309,7 +20319,7 @@ declare var HTMLOptionsCollection: {
     new(): HTMLOptionsCollection;
 };
 
-interface HTMLOrSVGElement {
+interface HTMLOrSVGOrMathMLElement {
```
jakebailey, line 20325: The old name should be preserved somehow, right?

saschanaz: I don't have proof that this is actively extended out there, but maybe...

HolgerJeromin: I am pretty sure that `HTMLOrSVGElement` is used in many code bases like `protected __element: HTMLOrSVGElement` or `public addElement(element: HTMLOrSVGElement)`. 
So this type should be kept valid

saschanaz: Pushed a fix, does it look good enough? 👀

jakebailey: Seems fine? Not sure about the impl itself but it has the effect I was looking for

saschanaz:
LGTM

PR #2487, opened 2026-05-28
Fix Element.matches strict narrowing

### LINKED ISSUE #63497: DOM: `Element#matches()` incorrectly narrows types
### 🔎 Search Terms

element matches narrow never

### 🕗 Version & Regression Information

- This changed between versions 5.9.3 and 6.0.3

### ⏯ Playground Link

https://www.typescriptlang.org/play/?ts=6.0.3#code/CYUwxgNghgTiAEEQBd4iQWxAO2QLngAkAVAWQBkBRTHZAbgCgBCdELXAOgymTAAsQAZwAU8AOTBkY+AEp4AMnloanbrwEjxwYNJlA

### 💻 Code

```ts
declare let element: HTMLElement;
!element.matches( 'dt' ) && element.matches( 'dd' )
```

### 🙁 Actual behavior

`element.matches( 'dd' )` throws a `Property 'matches' does not exist on type 'never'` compilation error.

### 🙂 Expected behavior

`element.matches( 'dd' )` does not throw any error.

### Additional information about the issue

`element.matches( 'dt' )` uses the following overload:
```ts
matches<K extends keyof HTMLElementTagNameMap>(selectors: K): this is HTMLElementTagNameMap[K];
```

The return type of `element.matches( 'dt' )` is then `this is HTMLElementTagNameMap['dt']` that maps to `this is HTMLElement`.

Since `this` is `HTMLElement`:
- When false, the type of `element` is `HTMLElement` (not narrowed further).
- When false, the type of `element` is `never`.

cjc0013:
Fixes microsoft/TypeScript#63497.

This follows the direction discussed in that issue: `matches()` should only act as a type guard when the selected tag's element type is a strict subtype of the current receiver type. For tags that map back to the same receiver type, it should still be accepted but return `boolean`, so the false branch does not narrow the receiver to `never`.

## Summary
- restrict `Element.matches()` type-predicate overloads to strict subtype matches
- keep tag-name boolean overloads for same-type matches
- regenerate DOM baselines
- add a compile-only regression fixture for the reported false-branch `never` behavior

## Validation
- `npm run generate`
- `npm test`
- focused local TypeScript probe: the current baseline produced `Property 'id' does not exist on type 'never'`; the patched baseline compiles cleanly

### DIFF
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -645,7 +645,7 @@ export function emitWebIdl(
       const paramName = m.signature[0].param![0].name;
       for (const mapName of tagNameMapNames) {
         printer.printLine(
-          `matches<K extends keyof ElementMatchesMap<${mapName}, this>>(${paramName}: K): this is Extract<${mapName}[K], this>;`,
+          `matches<K extends keyof ${mapName} & keyof ElementMatchesNarrowingMap<${mapName}, this>>(${paramName}: K): this is Extract<${mapName}[K], this>;`,
         );
       }
       for (const mapName of tagNameMapNames) {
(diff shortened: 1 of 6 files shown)


cjc0013:

src/build/emitter.ts
```diff
@@ -715,6 +720,13 @@ export function emitWebIdl(
     printer.printLine("");
   }
 
+  function emitElementMatchesMap() {
+    printer.printLine(
+      "type ElementMatchesMap<T, U> = { [K in keyof T as T[K] extends U ? U extends T[K] ? never : K : never]: T[K] };",
```
saschanaz: This is hard to parse, can you use something like ElementMap and ElementType than T and U? And also with parentheses.

It also kinda works weird, `keyof ElementMatchesMap<HTMLElementTagNameMap, HTMLDivElement>` is somehow `"object" | "col" | "colgroup" | "embed" | "hr" | "iframe" | "img" | "input" | "legend" | "table" | "tbody" | "td" | "tfoot" | "th" | "thead" | "tr"` ? It should be `"div"`, right?

saschanaz (changes requested):

src/build/emitter.ts
cjc0013: Thanks, that makes sense. I updated the helper to make the strict-narrowing intent explicit with clearer names and a parenthesized mapped-type condition.

I also changed the narrowing map so same-type matches and structurally unrelated matches fall back to `boolean`. That covers the `HTMLDivElement` case you pointed out, so `HTMLDivElement.matches("div")` stays boolean and `HTMLDivElement.matches("object")` no longer gets a bogus type-guard path.

I added fixture coverage for the original false-branch regression, `HTMLElement` narrowing to `HTMLDivElement`, `HTMLDivElement.matches("div")` staying boolean, and `HTMLDivElement.matches("object")` not narrowing to `HTMLObjectElement`.

Issue #2506, opened 2026-07-08
Make `RequestInit` method typesafe


GustavoOS:
Currently, if you use something like `fetch(url, {method: "foo"})` there is no type error, despite "foo" not being a proper Http verb.

PR #2507, opened 2026-07-08
Add type to RequestInit method

GustavoOS:
As described by #2506 this PR allows type safe use of RequestInit method type.


GustavoOS:
@microsoft-github-policy-service agree company="Bosch"

jakebailey:
A point of consideration here is going to be compatibility with `@types/node`, so keep that in mind

turansky:
`RequestMethod`?

Screenshot 2026-07-11 at 01 55 03

PR #2510, opened 2026-07-18
Manually update to idl@3.79.0

saschanaz:

PR #2512, opened 2026-07-18
Feat: Support comments for mixins

Bashamega:
Previously, MDN comments were only generated for interfaces. However, some internal mixin interfaces (e.g. `HyperlinkElementUtils`) don't exist in the MDN data and instead derive their documentation from a corresponding interface.

To avoid documenting these manually, this PR introduces an `interfaceAliasMap` that maps MDN interface names to internal mixin interface names. During generation, the documentation comments from the MDN interface are copied to each mapped mixin interface.

This removes the need for manual documentation duplication and fixes the inconsistent behavior seen in #2510.


Bashamega:
If this strategy for comments is good i can generate comments for the majority of the missing data

saschanaz:
This is better but not ideal, as we should automatically map it rather than manually, just as how mixins are getting MDN links now.

That automation is not possible purely with MDN data, we would need Web IDL data to get the mapping, like how it's done in bcd-idl-mapper. Maybe that logic can somehow be exposed to be reused here...

saschanaz:
But I think this is still an acceptable patch and it unblocks #2510. Considering an automated way would be the next step.

LGTM

Bashamega:
> This is better but not ideal, as we should automatically map it rather than manually, just as how mixins are getting MDN links now.
> 
> That automation is not possible purely with MDN data, we would need Web IDL data to get the mapping, like how it's done in bcd-idl-mapper. Maybe that logic can somehow be exposed to be reused here...

What do you think of changing our strategy completely for mapping, so instead we generate comments for everything.
We map the webIDL variable and map out all the mdn urls and insert the comment based on it

PR #2513, opened 2026-07-18
Emit object for Web IDL object types

### LINKED ISSUE #2475: Emit `object` for Web IDL `object` types?
This generator has been treating `object` as `any` ever since it's created in 2017: https://github.com/microsoft/TypeScript-DOM-lib-generator/blame/317c27a36a49528e86f6f16ab8f01f3239f80ea0/TS.fsx#L719

But in the same year TypeScript introduced `object` type: https://devblogs.microsoft.com/typescript/announcing-typescript-2-2/#the-object-type

In #2459 I got a reason to make it actually `object`, since `Origin.from()` should ideally receive `object | string`. (That's still not the most strict type, but I don't think we can do anything more than that.)

@jakebailey Thoughts about any reason we should not do this?

Bashamega:
closes #2475

### DIFF
--- a/baselines/audioworklet.generated.d.ts
+++ b/baselines/audioworklet.generated.d.ts
@@ -1887,7 +1887,7 @@ interface Console {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/console/dir_static)
      */
-    dir(item?: any, options?: any): void;
+    dir(item?: any, options?: object | null): void;
     /**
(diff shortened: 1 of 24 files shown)


saschanaz:
This will need some test against DT. I think it needed maintainer permission? @jakebailey

jakebailey:
As in, pull into TS and then try to run it?

Probably not so easy now that we can't test nightlies on DT anymore (until that infra is reworked), but potentially still possible to an extent

jakebailey:
I do expect this to be very breaking, however. The `toJSON` changes alone scare me a bit at first glance (but, I know what I said on the issue, of course)

PR #2515, opened 2026-07-21
fix: type observer callback entries and box sizes as non-empty tuples

homayounmmdy:
This PR addresses [microsoft/TypeScript#61691](https://github.com/microsoft/TypeScript/issues/61691).

**The Problem:**  
The current Web IDL spec types observer callback parameters (like `entries` in `ResizeObserverCallback`) as standard arrays (`T[]`). However, the browser specification guarantees that these callbacks are only ever invoked when there is at least one observation. This forces developers to use unnecessary non-null assertions (`!`) or runtime checks for a state that is theoretically impossible.

**The Solution:**  
This PR adds manual overrides in `overridingTypes.jsonc` to enforce a non-empty tuple type (`[T, ...T[]]`) for:
- `ResizeObserverCallback`
- `IntersectionObserverCallback`
- `MutationCallback`
- `ResizeObserverEntry`'s `contentBoxSize`, `borderBoxSize`, and `devicePixelContentBoxSize` properties.

**Verification:**  
Ran `npm run generate` and verified the output in `baselines/dom.generated.d.ts` reflects the correct non-empty tuple types without breaking the build.

### DIFF
--- a/baselines/dom.generated.d.ts
+++ b/baselines/dom.generated.d.ts
@@ -31257,13 +31257,13 @@ interface ResizeObserverEntry {
      *
      * [MDN Reference](https://developer.mozilla.org/docs/Web/API/ResizeObserverEntry/borderBoxSize)
      */
-    readonly borderBoxSize: ReadonlyArray<ResizeObserverSize>;
+    readonly borderBoxSize: readonly [ResizeObserverSize, ...ResizeObserverSize[]];
     /**
      * The **`contentBoxSize`** read-only property of the ResizeObserverEntry interface returns an array containing the new content box size of the observed element when the callback is run.
      *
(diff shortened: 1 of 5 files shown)


homayounmmdy:

homayounmmdy:
Hi @saschanaz , friendly ping on this one 👋

This PR is ready for review when you have a chance. It updates the DOM lib overrides so observer callback entries and related box-size arrays are typed as non-empty tuples where the spec guarantees at least one entry, addressing microsoft/TypeScript#61691.

I’ve run  npm run generate  and verified the generated baseline changes locally. Happy to make any adjustments if you’d prefer a different approach.

Thanks for your time!

HolgerJeromin:
I think new overrides should be done as KDL files now:

https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/2053

Bashamega (approved):
Looks good, what do you think @saschanaz

PR #2533, opened 2026-08-28
Support async_sequence

### LINKED ISSUE #2480: Support new WebIDL `async_sequence<T>` type
Not immediately urgent given no current usage AFAIA, but will likely start to diffuse into definitions (whatwg/streams#1372 etc.)

In terms of binding, this maps to `AsyncIterable<T>` when converting from IDL to JS. When converting from JS to IDL, it can technically fall back on a sync iterator if an async iterator isn't found, so so would probably technically be something like `AsyncIterable<T> | Iterable<T>`.

This won't be an issue for the "primary" builds (ts6.0+), since the generator adds a `<reference>` to lib.2018.asynciterable, so `Symbol.asyncIterator` is always present in the project global scope.

This will be an issue for older targets, where there are no `<reference>` directives, and interfaces marked `async_iterable` have their iterator methods (`[Symbol.asyncIterator]`/`entries`/`keys`/`values`) separated out into the `*.asynciterable` definition files. If we follow the same logic, that is going to mean that any method or property that references `async_sequence` is going to have to be excluded from the main definition files. A change in approach might be necessary, ie. "backporting" the `<reference>` pattern to these older targets, or just typing `async_sequence`s as `any` for older targets.

Bashamega:
closes #2480

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -252,6 +252,7 @@ async function emitDom() {
         allowUnrelatedSetterType: true,
         useGenericTypedArrays: true,
         includeIterable: true,
+        treatAsyncSequence: true,
       },
     },
     // ts5.7 (and later)
@@ -262,6 +263,7 @@ async function emitDom() {
         useIteratorObject: true,
         allowUnrelatedSetterType: true,
         useGenericTypedArrays: true,
+        treatAsyncSequence: true,
       },
     },
     // ts5.6
@@ -271,12 +273,15 @@ async function emitDom() {
       compilerBehavior: {
         useIteratorObject: true,
         allowUnrelatedSetterType: true,
+        treatAsyncSequence: false,
       },
     },
     // ts5.5 (and earlier)
     {
       outputFolder: new URL("./ts5.5/", outputFolder),
-      compilerBehavior: {}, // ts5.5 does not support `IteratorObject` or unrelated setter types
+      compilerBehavior: {
+        treatAsyncSequence: false,
+      }, // ts5.5 does not support `IteratorObject` or unrelated setter types
     },
   ];
 
--- a/src/build.ts
+++ b/src/build.ts
@@ -273,7 +273,6 @@ async function emitDom() {
       compilerBehavior: {
         useIteratorObject: true,
         allowUnrelatedSetterType: true,
-        treatAsyncSequence: false,
       },
     },
     // ts5.5 (and earlier)
--- a/src/build.ts
+++ b/src/build.ts
@@ -279,9 +279,7 @@ async function emitDom() {
     // ts5.5 (and earlier)
     {
       outputFolder: new URL("./ts5.5/", outputFolder),
-      compilerBehavior: {
-        treatAsyncSequence: false,
-      }, // ts5.5 does not support `IteratorObject` or unrelated setter types
+      compilerBehavior: {}, // ts5.5 does not support `IteratorObject` or unrelated setter types
     },
   ];
 
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -409,24 +409,13 @@ export function emitWebIdl(
     function convertBaseType() {
       // Support async_sequence (see https://github.com/whatwg/streams/pull/1372)
       if (obj.type === "async_sequence") {
-        // If we're in a build that includes async iterables (TS 4.4+ or forced), emit as AsyncIterable<T>.
-        // For TS <4.4, the generator separates out [Symbol.asyncIterator] into *.asynciterable.d.ts, so
-        // these references must not leak into the main file, so we can treat them as `any` as a fallback.
-        if (compilerBehavior.treatAsyncSequence) {
-          const subtype = arrayify(obj.subtype)
-            .map((t) => convertDomTypeToTsTypeBase(t, forReturn))
-            .join(", ");
-          // The assignment to `subtype` is only needed to validate/trigger side-effects,
-          // but its value is not used because we always return "AsyncIterable"
-          // unless subtype is empty, in which case we substitute "any"
-          if (!subtype) {
-            return "AsyncIterable";
-          }
-          return "AsyncIterable";
-        } else {
-          // Legacy TypeScript versions: no async iterable interface available.
-          return "any";
-        }
+        // AsyncIterable<T> requires separate asynciterable.d.ts until TS 6.0.
+        // To make it correct, we need to defer any functions using async_sequence to the asynciterable variant.
+        // For now, skip generation and warn.
+        console.warn(
+          "Skipping generation for a function or property with async_sequence type: " + JSON.stringify(obj)
+        );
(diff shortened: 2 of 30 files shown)


saschanaz:
Please just do the async_sequence part, the other parts are too big to consume.

Bashamega:
> Please just do the async_sequence part, the other parts are too big to consume.

Sure, but how can we test it without upgrading the library?

saschanaz:
That's a good question. But it's kinda more important to make the job pass than making it correct right now, we'll get it checked later when we get the actual IDL updated.

Bashamega:
> That's a good question. But it's kinda more important to make the job pass than making it correct right now, we'll get it checked later when we get the actual IDL updated.

Done. Thanks for looking at this. When will you check my previous PRs? Or should I close them?

src/build/emitter.ts
```diff
@@ -405,6 +407,30 @@ export function emitWebIdl(
...
+            .join(", ");
+          // The assignment to `subtype` is only needed to validate/trigger side-effects,
+          // but its value is not used because we always return "AsyncIterable"
+          // unless subtype is empty, in which case we substitute "any"
+          if (!subtype) {
+            return "AsyncIterable";
+          }
+          return "AsyncIterable";
+        } else {
+          // Legacy main & fallback builds: no async iterable interface available.
```
saschanaz: What is main & fallback builds? Can we say legacy TS versions?

src/build/emitter.ts
```diff
@@ -405,6 +407,30 @@ export function emitWebIdl(
...
     function convertBaseType() {
+      // Support async_sequence (see https://github.com/whatwg/streams/pull/1372)
+      if (obj.type === "async_sequence") {
+        // If we're in a build that includes async iterables (TS 2.3+ or forced), emit as AsyncIterable<T>.
+        // For TS <2.3, the generator separates out [Symbol.asyncIterator] into *.asynciterable.d.ts, so
+        // these references must not leak into the main file, so we can treat them as `any` as a fallback.
+        if (compilerBehavior.treatAsyncSequence) {
+          // Forwards-compatible definition:
+          // NOTE: In IDL→TS (whatwg/streams#1372 etc) async_sequence<T> -> AsyncIterable<T>
+          // If stricter interop wanted, could use AsyncIterable<T> | Iterable<T>
```
saschanaz: Why?

src/build/emitter.ts
```diff
@@ -405,6 +407,30 @@ export function emitWebIdl(
     forReturn: boolean,
   ): string {
     function convertBaseType() {
+      // Support async_sequence (see https://github.com/whatwg/streams/pull/1372)
+      if (obj.type === "async_sequence") {
+        // If we're in a build that includes async iterables (TS 2.3+ or forced), emit as AsyncIterable<T>.
+        // For TS <2.3, the generator separates out [Symbol.asyncIterator] into *.asynciterable.d.ts, so
+        // these references must not leak into the main file, so we can treat them as `any` as a fallback.
+        if (compilerBehavior.treatAsyncSequence) {
+          // Forwards-compatible definition:
```
saschanaz: forwards-compatible?

src/build/emitter.ts
```diff
@@ -405,6 +407,30 @@ export function emitWebIdl(
     forReturn: boolean,
   ): string {
     function convertBaseType() {
+      // Support async_sequence (see https://github.com/whatwg/streams/pull/1372)
+      if (obj.type === "async_sequence") {
+        // If we're in a build that includes async iterables (TS 2.3+ or forced), emit as AsyncIterable<T>.
+        // For TS <2.3, the generator separates out [Symbol.asyncIterator] into *.asynciterable.d.ts, so
```
saschanaz: I don't think we care about TS < 2.3, minimum supported version is already 4.4.

That said, AsyncIterable does require separate asynciterable.d.ts until TS 6.0, so to make it correct we need to defer any functions that uses `async_sequence` to asynciterable variant.

Given we don't really test it here, it's also acceptable to just entirely skip the generation for those functions with some console warning.

src/build.ts
```diff
@@ -271,12 +273,15 @@ async function emitDom() {
...
         allowUnrelatedSetterType: true,
+        treatAsyncSequence: false,
       },
     },
     // ts5.5 (and earlier)
     {
       outputFolder: new URL("./ts5.5/", outputFolder),
-      compilerBehavior: {}, // ts5.5 does not support `IteratorObject` or unrelated setter types
+      compilerBehavior: {
+        treatAsyncSequence: false,
```
saschanaz: Let it just empty as others did here

saschanaz (changes requested):

Bashamega:
done

src/build/emitter.ts
saschanaz: I don't think we are doing this?

Bashamega: I don't quite understand this comment but i have pushed a change tell me if this is what you mean

src/build.ts
```diff
@@ -271,6 +273,7 @@ async function emitDom() {
       compilerBehavior: {
         useIteratorObject: true,
         allowUnrelatedSetterType: true,
+        treatAsyncSequence: false,
```
saschanaz: Again, please skip it as others do

Bashamega: Sorry, done

src/build/emitter.ts
```diff
@@ -442,6 +455,9 @@ export function emitWebIdl(
     }
 
     const type = convertBaseType();
+    if (type === "any") {
+      return type;
+    }
```
saschanaz: Does this still matter?

Bashamega: Updated

saschanaz:
Thanks!

For future reference - I think the fix is in a wrong place - Ideally it should be somewhere in emitMethod or the caller of emitMethod, check the return type from there, and skip the emission for older TS versions. We should not emit `any` because that's unrecoverable with overloads.

But for now this is okay as we'll check it later.

LGTM

saschanaz:
As for future work - I think it would be nice to consume each of webref/idl and BCD updates release-by-release to reduce #2495 and eventually close it.

About KDL work - I have been thinking of having a library - https://codeberg.org/krosylight/kdl2json, but hasn't been made any progress recently. I was thinking maybe I should encourage contribution once I get it more in form, but any help (or any question for future direction to contribute for) would be nice.

Bashamega:
> As for future work - I think it would be nice to consume each of webref/idl and BCD updates release-by-release to reduce #2495 and eventually close it.
> 
> About KDL work - I have been thinking of having a library - https://codeberg.org/krosylight/kdl2json, but hasn't been made any progress recently. I was thinking maybe I should encourage contribution once I get it more in form, but any help (or any question for future direction to contribute for) would be nice.

Hello,
Thank you for sharing this with me.
I would love to help you out.
Please tell me how I can help out with the kdl2json library. Meanwhile I will try to reduce the pr.

PR #2535, opened 2026-08-30
Update @webref/idl to version 3.82.1

Bashamega:
Split #2495


baselines/ts5.5/dom.generated.d.ts
```diff
@@ -29163,7 +29163,7 @@ interface ProcessingInstruction extends CharacterData, LinkStyle {
 
 declare var ProcessingInstruction: {
     prototype: ProcessingInstruction;
-    new(): ProcessingInstruction;
+    new(target: string, data?: string): ProcessingInstruction;
```
saschanaz: This is not supported by anyone I believe, we'll have to also remove this in KDL.

Bashamega: I have fixed it and added support for it in kdl

src/build/types.ts
```diff
@@ -7,11 +7,13 @@ export interface Typed {
   allowShared?: boolean;
 }
 
-export interface Param extends Typed {
-  name: string;
-  optional?: boolean;
-  variadic?: boolean;
-}
+export type Param =
```
saschanaz: Can you do it without changing types.ts? Patching should not change base types unless the type is already wrong.

src/build/patches.ts
```diff
@@ -378,18 +382,27 @@ function handleMethodAndConstructor(
...
+      case "param": {
+        const param = handleParam(c);
+        if (
+          Object.keys(param).length === 2 &&
+          param.name &&
+          param.additionalTypes === undefined
+        ) {
+          params.push(param.name);
+        } else {
+          params.push(param);
```
saschanaz: It's better, but still kinda sad. I wonder it would be better to somehow treat `{ name: "foo" }` with no other fields as `foo` on prune() level. If that's possible then we won't need this workaround for any future array items either. Thoughts?

Bashamega:
I have improved the logic @saschanaz

src/build/patches.ts
```diff
@@ -532,9 +532,27 @@ async function readPatchDocument(fileUrl: URL): Promise<Document> {
...
+      obj.length &&
+      obj.every(
+        (item) =>
+          item &&
+          typeof item === "object" &&
+          !Array.isArray(item) &&
+          Object.keys(item).length === 1 &&
+          Object.prototype.hasOwnProperty.call(item, "name") &&
+          typeof (item as Record<string, unknown>).name === "string",
+      )
```
saschanaz: Can you extract this lambda in every() into a separate function so that it becomes `obj.every(isObjectWithOnlyName)` (name free to change)?

saschanaz (commented):
Thanks! I'll need to double check other changes, but the logic patch looks really good 🙏🏻👍🏻🎉

src/build/patches.ts
Bashamega: Thanks for the learning, I have updated it

inputfiles/patches/pointerevents.kdl
```diff
@@ -0,0 +1,5 @@
+removals {
+    dictionary WheelEventInit {
+        member momentum // Blink only as of 2025-08
```
saschanaz: 2025 or 2026? 👀

Bashamega: LOL, this was a typo, updated

saschanaz:
Now about the compiler test failure.

```
--- FAIL: TestLocal (2.34s)
    --- FAIL: TestLocal/globalThisBlockscopedProperties.ts (0.02s)
        --- FAIL: TestLocal/globalThisBlockscopedProperties.ts/type (0.00s)
            baseline.go:79: the baseline file /home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/TypeScript/tsc/testdata/baselines/reference/conformance/globalThisBlockscopedProperties.types has changed. (Run `hereby baseline-accept` if the new baseline is correct.)
    --- FAIL: TestLocal/mappedTypeRecursiveInference.ts (0.15s)
        --- FAIL: TestLocal/mappedTypeRecursiveInference.ts/type (0.12s)
            baseline.go:79: the baseline file /home/runner/work/TypeScript-DOM-lib-generator/TypeScript-DOM-lib-generator/TypeScript/tsc/testdata/baselines/reference/compiler/mappedTypeRecursiveInference.types has changed. (Run `hereby baseline-accept` if the new baseline is correct.)
```

@jakebailey, this is the first time I've seen this test failing. I don't understand why it fails because no line change happened, and [the test file](https://github.com/microsoft/TypeScript/blob/253c5e2074a43301c93e34e5616c7249d86d71f6/tsc/testdata/tests/cases/conformance/es2019/globalThisBlockscopedProperties.ts#L4) looks very simple. what do we do? Ignore it?

inputfiles/patches/dom.kdl
```diff
@@ -40,3 +40,12 @@ enum InsertPosition {
 // ChildNode and ParentNode are actually defined as mixins, but because of their names they have historically been used as a sub-interface of Node.
 interface-mixin ChildNode extends=Node
 interface-mixin ParentNode extends=Node
+
+removals { 
+  interface ProcessingInstruction {
+    constructor signatureIndex=0 {
+      param target
+      param data
+    }
```
saschanaz, line 49: We should ideally remove the constructor itself than the parameters, but the result is same for TS, so this is fine.

saschanaz:
The only change is SVG geometry stuff, which is fine. LGTM, thanks!

PR #2546, opened 2026-09-04
Update @webref/idl to version 3.83.0

Bashamega:


saschanaz:
Huh, added quite recently to implementations. LGTM

PR #2552, opened 2026-09-07
Update URL & Request & XMLHttpRequest interface to accept Location

### LINKED ISSUE #2536: [Web API type definition issue] `URL`, `fetch`, and `XMLHttpRequest.prototype.open` should accept `Location` as an argument
`Location.prototype.toString` behaves identically to `URL.prototype.toString`; both return the value of their own `href` property.

```js
new URL(location)
new URL("./", location)

URL.canParse(location)
URL.canParse("./", location)

URL.parse(location)
URL.parse("./", location)

fetch(location)

new XMLHttpRequest().open("GET", location)

new Request(location)
```

Bashamega:
closes #2536 

Also, I couldn't figure out how to override fetch

### DIFF
--- a/inputfiles/patches/fetch.kdl
+++ b/inputfiles/patches/fetch.kdl
@@ -7,7 +7,7 @@ removals {
 interface Request {
   constructor signatureIndex=0 {
     param input {
-      additionalTypes input Location
+      additionalTypes Location
     }
   }
 }
--- a/inputfiles/patches/fetch.kdl
+++ b/inputfiles/patches/fetch.kdl
@@ -3,3 +3,11 @@ removals {
     text // No implementation as of 2026-04
   }
 }
+
+interface Request {
+  constructor signatureIndex=0 {
+    param input {
+      additionalTypes input Location
+    }
+  }
+}
--- a/inputfiles/patches/fetch.kdl
+++ b/inputfiles/patches/fetch.kdl
@@ -7,7 +7,7 @@ removals {
 interface Request {
   constructor signatureIndex=0 {
     param input {
-      additionalTypes Location
+      additionalTypes Location WorkerLocation
     }
   }
 }
--- a/inputfiles/patches/url.kdl
+++ b/inputfiles/patches/url.kdl
@@ -9,26 +9,26 @@ interface URLSearchParams {
 interface URL {
     constructor signatureIndex=0 {
         param url {
-            additionalTypes URL Location
+            additionalTypes URL Location WorkerLocation
         }
         param base {
-            additionalTypes URL Location
+            additionalTypes URL Location WorkerLocation
         }
     }
     method canParse signatureIndex=0 {
         param url {
-            additionalTypes URL Location
+            additionalTypes URL Location WorkerLocation
         }
         param base {
-            additionalTypes URL Location
+            additionalTypes URL Location WorkerLocation
         }
     }
     method parse signatureIndex=0 {
         param url {
-            additionalTypes URL Location
+            additionalTypes URL Location WorkerLocation
         }
         param base {
-            additionalTypes URL Location
+            additionalTypes URL Location WorkerLocation
         }
     }
 }
--- a/inputfiles/patches/url.kdl
+++ b/inputfiles/patches/url.kdl
@@ -8,6 +8,7 @@ interface URLSearchParams {
 
 interface URL {
     constructor signatureIndex=0 {
+        param url overrideType="string | URL | Location"
         param base overrideType="string | URL | Location"
     }
     method canParse signatureIndex=0 {
--- a/inputfiles/patches/url.kdl
+++ b/inputfiles/patches/url.kdl
@@ -8,15 +8,27 @@ interface URLSearchParams {
 
 interface URL {
     constructor signatureIndex=0 {
-        param url overrideType="string | URL | Location"
-        param base overrideType="string | URL | Location"
+        param url {
+            additionalTypes URL Location
+        }
+        param base {
+            additionalTypes URL Location
+        }
     }
     method canParse signatureIndex=0 {
-        param url overrideType="string | URL | Location"
-        param base overrideType="string | URL | Location"
+        param url {
+            additionalTypes URL Location
+        }
+        param base {
+            additionalTypes URL Location
+        }
     }
     method parse signatureIndex=0 {
-        param url overrideType="string | URL | Location"
-        param base overrideType="string | URL | Location"
+        param url {
+            additionalTypes URL Location
+        }
+        param base {
+            additionalTypes URL Location
+        }
     }
 }
--- a/inputfiles/patches/url.kdl
+++ b/inputfiles/patches/url.kdl
@@ -12,9 +12,11 @@ interface URL {
         param base overrideType="string | URL | Location"
     }
     method canParse signatureIndex=0 {
+        param url overrideType="string | URL | Location"
         param base overrideType="string | URL | Location"
     }
     method parse signatureIndex=0 {
+        param url overrideType="string | URL | Location"
         param base overrideType="string | URL | Location"
     }
 }
--- a/inputfiles/patches/url.kdl
+++ b/inputfiles/patches/url.kdl
@@ -5,3 +5,15 @@ interface URLSearchParams {
         }
     }
 }
+
+interface URL {
+    constructor signatureIndex=0 {
+        param base overrideType="string | URL | Location"
+    }
+    method canParse signatureIndex=0 {
+        param base overrideType="string | URL | Location"
+    }
+    method parse signatureIndex=0 {
+        param base overrideType="string | URL | Location"
+    }
+}
--- a/src/build/emitter.ts
+++ b/src/build/emitter.ts
@@ -429,7 +429,11 @@ export function emitWebIdl(
           typeof obj.type === "string"
             ? [{ ...obj, additionalTypes: undefined }]
             : obj.type;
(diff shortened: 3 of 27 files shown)


Bashamega:
Should i handle the fallback to WorkerLocation in the emitter or what should i do in this situation?

inputfiles/patches/url.kdl
```diff
@@ -5,3 +5,15 @@ interface URLSearchParams {
...
+    constructor signatureIndex=0 {
+        param base overrideType="string | URL | Location"
+    }
+    method canParse signatureIndex=0 {
+        param base overrideType="string | URL | Location"
+    }
+    method parse signatureIndex=0 {
+        param base overrideType="string | URL | Location"
+    }
+}
```
bddjr, line 34: The first parameter `url` passed to `URL` should also support the `Location` type.
```js
new URL(location)
```

Bashamega: Updated

bddjr:
```js
URL.canParse(location)
URL.parse(location)
```

bddjr:
🤔It seems that we need to separately make `webworker.generated.d.ts` exempt from this change.  
I don't know how to do that, so I didn't submit a PR.

inputfiles/patches/url.kdl
```diff
@@ -5,3 +5,18 @@ interface URLSearchParams {
         }
     }
 }
+
+interface URL {
+    constructor signatureIndex=0 {
+        param url overrideType="string | URL | Location"
```
saschanaz: Why not additionalTypes? 🤔

Bashamega: Because location is not supported in web worker, i don't know how to handle web worker

saschanaz: the exposure checker should autoremove them... but maybe we don't do that for additionalTypes? 🤔

Bashamega:
It wasn't implemented, but I have added it @saschanaz

turansky:
> Also, I couldn't figure out how to override `fetch`

@Bashamega should it also affect `Request` in that case?

bddjr:
> should it also affect `Request` in that case?

yes

```js
new Request(location)
```

Bashamega:
I have updated it. What do you think @saschanaz

bddjr:
ChatGPT suggested adding the type for `fetch(location)` like this:

```diff
 inputfiles/overridingTypes.jsonc

 "WindowOrWorkerGlobalScope": {
     "methods": {
         "method": {
+            "fetch": {
+                "signature": {
+                    "0": {
+                        "param": [
+                            {
+                                "name": "input",
+                                "additionalTypes": ["Location"]
+                            }
+                        ]
+                    }
+                }
+            },
             ...
```

Bashamega:
Thanks for the help @bddjr ... Updated

bddjr:
Review by Gemini 3.8 Flash:

Thanks for working on this! A few findings during the review:

### 1. Typo in `inputfiles/patches/fetch.kdl`
In `inputfiles/patches/fetch.kdl`:
```kdl
interface Request {
  constructor signatureIndex=0 {
    param input {
      additionalTypes input Location
    }
  }
}
```
`input` was accidentally included as a type in `additionalTypes`. It should be:
```kdl
      additionalTypes Location
```
(It was silently stripped only because `isKnownType("input")` is false).

### 2. `filterAdditionalTypes` in `src/build/expose.ts` didn't actually affect parameters
In `filterUnknownTypeFromSignature`, `param.push({ ...p, type: flattenType(filtered) })` spreads `p` rather than using the filtered `additionalTypes` from `filtered[0]`. In addition, `deepClone` returns early at method signatures and does not recurse into `param` directly.
As a result, `additionalTypes` on parameters remained unfiltered until `isKnownType` in `emitter.ts`.
To properly filter `additionalTypes` in `expose.ts`:
```ts
const additionalTypes = filterAdditionalTypes(p.additionalTypes, unexposedTypes);
param.push({
  ...p,
  type: flattenType(filtered),
  additionalTypes: additionalTypes?.length ? additionalTypes : undefined,
});
```

### 3. Supporting `WorkerLocation` in Web Workers
Regarding your question about `WorkerLocation`:
> *Should i handle the fallback to WorkerLocation in the emitter or what should i do in this situation?*

Once `expose.ts` properly filters `additionalTypes`, we can simply specify:
```kdl
additionalTypes Location WorkerLocation
```
- For DOM lib: `WorkerLocation` is unexposed and filtered out $\rightarrow$ results in `Location`.
- For Worker lib: `Location` is unexposed and filtered out $\rightarrow$ results in `WorkerLocation`.
This allows workers to naturally accept `location` without any special fallback in `emitter.ts`!

bddjr:
Review by Gemini 3.8 Flash:

Nice work on adding `WorkerLocation`! The generated baselines look spot on now.

Just one architectural detail regarding `src/build/expose.ts`:

Currently, the changes in `src/build/expose.ts` are actually dead code because:
1. `deepClone` returns early on `o.signature` and never traverses down into `param` objects.
2. In `filterUnknownTypeFromSignature`, `param.push({ ...p, type: flattenType(filtered) })` spreads `p` (the original param) rather than preserving `filtered[0].additionalTypes`.

The reason the baselines generated correctly is actually due to `filter((t) => isKnownType(t))` in `src/build/emitter.ts`.

To clean this up, we have two options:
- **Option 1 (Recommended)**: Make `expose.ts` actually handle `additionalTypes` on parameters:
  ```ts
  // In filterUnknownTypeFromSignature in src/build/expose.ts
  const additionalTypes = filterAdditionalTypes(p.additionalTypes, unexposedTypes);
  param.push({
    ...p,
    type: flattenType(filtered),
    additionalTypes: additionalTypes?.length ? additionalTypes : undefined,
  });
  ```
  (and remove the unreachable `if (o.additionalTypes)` at the top of `deepClone`).
- **Option 2**: If filtering via `isKnownType` in `emitter.ts` is deemed sufficient, revert the changes in `src/build/expose.ts` completely so we don't leave unused code in `expose.ts`.

Bashamega:
I have updated it @bddjr 
Can you please check it when you have time @saschanaz 
Thank you

bddjr (approved):

src/build/emitter.ts
```diff
@@ -795,7 +811,9 @@ export function emitWebIdl(
       p = resolvePromise(p);
       if (acceptsUrl(p)) {
         p = { ...p, additionalTypes: [...(p.additionalTypes ?? [])] };
-        p.additionalTypes!.push("URL");
+        if (!p.additionalTypes!.includes("URL")) {
+          p.additionalTypes!.push("URL");
+        }
```
saschanaz: Why? Just don't add it in patch that case?

src/build/emitter.ts
```diff
@@ -520,6 +524,18 @@ export function emitWebIdl(
...
+    return (
+      baseTypeConversionMap.has(type) ||
+      allInterfacesMap[type] ||
+      allLegacyWindowAliases.includes(type) ||
+      allCallbackFunctionsMap[type] ||
+      allDictionariesMap[type] ||
+      allEnumsMap[type] ||
+      allTypedefsMap[type]
+    );
+  }
```
saschanaz: This is a wrong layer, this filter should belong to expose.ts where things are already happening

src/build/expose.ts
```diff
@@ -204,7 +204,17 @@ function deepFilterUnexposedTypes(
...
+        const additionalTypes = filterAdditionalTypes(
+          p.additionalTypes,
+          unexposedTypes,
+        );
+        param.push({
+          ...p,
+          type: flattenType(filtered),
+          additionalTypes: additionalTypes?.length
+            ? additionalTypes
+            : undefined,
```
saschanaz: Perhaps make filterAdditionalTypes return undefined if empty?

inputfiles/patches/fetch.kdl
```diff
@@ -3,3 +3,11 @@ removals {
     text // No implementation as of 2026-04
   }
 }
+
+interface Request {
+  constructor signatureIndex=0 {
+    param input {
+      additionalTypes Location WorkerLocation
```
saschanaz, line 10: I wonder we could have something like `acceptsUrl=#true` for type and then the emitter could add Location/WorkerLocation based on the emission target 🤔, at the same line where URL is being added right now. Thoughts?

saschanaz (changes requested):

inputfiles/patches/fetch.kdl
Bashamega: We will need to change a lot of other files is it worth it?
Also should this be similar to IDL?

Bashamega:
I have updated it

inputfiles/patches/fetch.kdl
saschanaz: We are doing this because IDL doesn't say whether URL is expected or not at all. I thought doing it would cause less files, any idea how much changes would be needed?

But we may need the additionalTypes feature in the future after all, so I'm going to accept this as-is

Bashamega: Will this mean we translate that property to jsonc and support it in the original parser? Or did I misunderstand?

saschanaz: There's no original parser? Not sure I understand

saschanaz:
LGTM, thanks!

inputfiles/patches/fetch.kdl
Bashamega: NVM

